PR #2, opened 2017-10-19
Add lock for preventing race condition

ScorpioCPH:
Fix `FIXME: there is a race condition if another goroutine calls m.Update concurrently.`

@flx42 PTAL.

### DIFF
--- a/server.go
+++ b/server.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"strings"
 	"time"
+	"sync"
 
 	"github.com/NVIDIA/nvidia-docker/src/nvml"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


flx42:
Hello,

That's not what I had in mind. The race condition is theoretical right now, only one goroutine can call `Update` for one NvidiaDevicePlugin object.
Even if there were multiple goroutines running `HealthCheck()`, this wouldn't solve the problem since `m.devs` is actually updated later. 
Goroutine 1 could read `m.devs` and get `[{dev1, Healthy}, {dev2, Healthy}]`, but then it detects `dev1` is unhealthy thus it calls `m.Update` with `[{dev1, Unhealthy}, {dev2, Healthy}]`. The update is queued in the channel but `m.devs` is unchanged.
After goroutine 1 released the lock, goroutine 2 will also get `[{dev1, Healthy}, {dev2, Healthy}]`. If it detects `dev2` is unhealthy it will call `m.Update` with `[{dev1, Healthy}, {dev2, Unhealthy}]`.

In summary: sending all devices in the `Update` is a bad idea. It should be done per-device, on state change.

ScorpioCPH:
@flx42 Thanks for your explanation.

I think there is only one NvidiaDevicePlugin object running on a node as we deploy it by `DaemonSet`, so maybe this is OK for now.

Close this PR, and continues to follow this question :-)

PR #3, opened 2017-10-19
Update readme for build binary on ubuntu

ScorpioCPH:
Update readme for build binary on ubuntu, @flx42 PTAL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ It requires nvidia-docker 2.0 with our runtime configured as the [default runtim
 
 ## Usage
 
+### Docker Env
+
 #### Build
 ```
 docker build -t nvidia-device-plugin:1.0.0 .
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,6 +23,28 @@ docker run -it -v /var/lib/kubelet/device-plugins:/var/lib/kubelet/device-plugin
 kubectl create -f nvidia-device-plugin.yml
 ```
 
+### Local Env
+
+You `MUST` set environment variable `LIBRARY_PATH` for static linking `cudart_static` library which required [here](https://github.com/NVIDIA/nvidia-docker/blob/master/src/cuda/bindings.go#L5), as show below (Ubuntu 16.04):
```
flx42: This is too complicated. You can just assume that the `/usr/local/cuda` symlink exists.
And then you can do:
```bash
C_INCLUDE_PATH=/usr/local/cuda/include LIBRARY_PATH=/usr/local/cuda/lib64 go build
```

README.md
```diff
@@ -21,6 +23,28 @@ docker run -it -v /var/lib/kubelet/device-plugins:/var/lib/kubelet/device-plugin
...
+```
+
+#### Build Binary
+```shell
+$ go build
+```
+
+#### Run locally
+```shell
+$ ./k8s-device-plugin
```
flx42: You will need `sudo`.
Also remove the `$`, to be consistent with the rest of the README.

README.md
```diff
@@ -21,6 +23,28 @@ docker run -it -v /var/lib/kubelet/device-plugins:/var/lib/kubelet/device-plugin
...
+# 2017/10/19 11:37:22 Loading NVIDIA management library
```
flx42: Remove this line to be consistent.

README.md
ScorpioCPH: @flx42 Thanks, good idea!

ScorpioCPH:
@flx42 Addressed your comments, PTAL.

flx42:
Pushed manually after some minor fixes, thanks.

PR #8, opened 2017-11-22
Change versioning to match the upstream version of K8s

idealhack:


flx42:
Ah yes I noticed that I missed that, I was planning to change that at the same time I push to Docker Hub (should be done today).

flx42:
Fixed by b919c1d3389af8c4a3b71aa782b340c4601b1d95

PR #21, opened 2018-01-12
Add documentation for selectively enable GPUs.

rongou:
Add a bit of documentation on how to work around older GPUs.


rongou:
I guess this is no longer needed?

PR #42, opened 2018-04-12
[bugfix] fix bug in server.go

liuyaoli-xiaomi:
`xids := make(chan *pluginapi.Device)`
should be
`xids = make(chan *pluginapi.Device)`


liuyaoli-xiaomi:
@flx42

flx42:
Pushed manually, thanks @liuyaoli-xiaomi !

PR #63, opened 2018-07-20
Add missing NVML_ERROR_NOT_FOUND in systemGetProcessName

samedguener:
Hey guys,
we were not able to read the name of attached processes at the GPU, due to the fact that this functionality is somehow not supported. This error was not caught, so we propose a small fix overhere.

We have following setup:
Tesla K80
cuda_driver: 387.26
nvidia-docker:  2.0.3
nvidia-container-runtime v1.3.0-1
docker-version: 17.03.02 

Best,
Samed


flx42:
> we were not able to read the name of attached processes at the GPU, due to the fact that this functionality is somehow not supported.

What do you mean it's not supported? If that's the case you should not be seeing this error. Most likely you are facing a TOCTOU issue instead.

This is not the right place for this PR, for Go projects you don't usually modify the vendored projects directly. We have moved the NVML bindings here:
https://github.com/NVIDIA/gpu-monitoring-tools

@guptaNswati we need to fix the vendoring in this repo to remove the reference to `nvidia-docker`.

PR #91, opened 2019-01-19
Set the default 'updateStrategy' to 'RollingUpdate' in branch v1.11

klueska:
This change allows us to patch the device plugin to (for example) update the
image tag to the latest version and have an automated rolling upgrade be
performed to update the plugin on all agent nodes.

Previously, the device plugin on each node would *only* be restarted with the
new image if the pod representing it was explicitly shutdown and restarted.

### DIFF
--- a/nvidia-device-plugin.yml
+++ b/nvidia-device-plugin.yml
@@ -4,6 +4,8 @@ metadata:
   name: nvidia-device-plugin-daemonset
   namespace: kube-system
 spec:
+  updateStrategy:
+    type: RollingUpdate
   template:
     metadata:
       # Mark this pod as a critical add-on; when enabled, the critical add-on scheduler


thomas-riccardi:
Maybe we should update the `apiVersion` from `extensions/v1beta1` to `apps/v1` where the default for `updateStrategy` is `RollingUpdate` (since kubernetes 1.9, see https://v1-9.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#daemonsetupdatestrategy-v1-apps).

The oldest branch for this device plugin is for kubernetes 1.8: we could keep the old `apiVersion` only there (and apply you fix there); or update it only for the current branche...

klueska:
That seems like a better approach to me, assuming there are no other negative implications.

klueska:
@RenaudWasTaken, @flx42 Do you guys see any issue with bumping the API to apps/v1? Why was extensions/v1beta1 chosen in the first place? Just for legacy reasons?

PR #92, opened 2019-01-19
Set the default 'updateStrategy' to 'RollingUpdate' in branch v1.12

klueska:
This change allows us to patch the device plugin to (for example) update the
image tag to the latest version and have an automated rolling upgrade be
performed to update the plugin on all agent nodes.

Previously, the device plugin on each node would *only* be restarted with the
new image if the pod representing it was explicitly shutdown and restarted.

### DIFF
--- a/nvidia-device-plugin.yml
+++ b/nvidia-device-plugin.yml
@@ -4,6 +4,8 @@ metadata:
   name: nvidia-device-plugin-daemonset-1.12
   namespace: kube-system
 spec:
+  updateStrategy:
+    type: RollingUpdate
   template:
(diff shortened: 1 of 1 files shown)


RenaudWasTaken:
Almost looks good to me, can you change the email to your nvidia email though?
Same comment for the other PR

thomas-riccardi:
Mayb  update the `apiVersion` from `extensions/v1beta1` to `apps/v1` instead?
See https://github.com/NVIDIA/k8s-device-plugin/pull/91#issuecomment-456103347

PR #104, opened 2019-03-26
Change how to specify pod as critical

Ethyling:
Use annotation and tolerations to mark pod as critical is deprecated in version 1.13 and removed in version 1.14 of Kubernetes. We need to use the priorityClassName pod attribute set as “system-cluster-critical” or “system-node-critical”.

Signed-off-by: Jordan Jacobelli <jjacobelli@nvidia.com>

### DIFF
--- a/nvidia-device-plugin.yml
+++ b/nvidia-device-plugin.yml
@@ -8,22 +8,26 @@ spec:
     type: RollingUpdate
   template:
     metadata:
-      # Mark this pod as a critical add-on; when enabled, the critical add-on scheduler
-      # reserves resources for critical add-on pods so that they can be rescheduled after
-      # a failure.  This annotation works in tandem with the toleration below.
+      # This annotation is deprecated. Kept here for backward compatibility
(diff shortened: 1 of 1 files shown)


nvidia-device-plugin.yml
```diff
@@ -8,22 +8,17 @@ spec:
     type: RollingUpdate
   template:
     metadata:
-      # Mark this pod as a critical add-on; when enabled, the critical add-on scheduler
-      # reserves resources for critical add-on pods so that they can be rescheduled after
-      # a failure.  This annotation works in tandem with the toleration below.
-      annotations:
```
RenaudWasTaken: Do you mind linking to the change / docs that mention this depreciation?

What happens when we have this Annotation in the newer versions?

Ethyling: Done.

The annotation will just be ignored and the pod's priority will be 0 (meaning low priority)

RenaudWasTaken: Can you keep this annotation

Ethyling: Done

PR #115, opened 2019-05-22
POC Multistage UBI8 build

zvonkok:
Adding Dockerfile for building the device-plugin based on UBI8.


RenaudWasTaken:
Hello @zvonkok, can you explicitly confirm here that you would be fine if after contributing this PR we relicensed the project to an Apache 2.0 project.

zvonkok:
@RenaudWasTaken yes you can relicense the project to an Apache 2.0 License. 

> UBI does NOT change the licensing requirements of any of the software built on top of it. Stated another way, the layers you build on top of UBI retain their own licensing – be it open source or proprietary, it’s up to you – and, you are still responsible for meeting those license requirements.

See e.g. http://crunchtools.com/ubi-licensing/. 

In a layered container scenario, each layer may have a different license and EULAs. For the base layer (UBI) the licenses associated with UBI and its components (GPL, Apache, ... licenses) still apply even if the resulting product is licensed under a proprietary license. 

Each layer in a container has its own license like every package (RPM/DEB) has its a own license in a Linux distribution.

A container can be seen as a distribution/combination of licensed layers.

elezar:
@zvonkok I am closing this PR. Please rebase and reopen if required.

Issue #148, opened 2019-11-28
m.server.Serve will panic


lujinda:
```
	go func() {
		lastCrashTime := time.Now()
		restartCount := 0
		for m.server != nil {
			log.Println("Starting GRPC server")
			err := m.server.Serve(sock)
			if err != nil {
				log.Printf("GRPC server crashed with error: %v", err)
			}
...
			timeSinceLastCrash := time.Since(lastCrashTime).Seconds()
			lastCrashTime = time.Now()
			if timeSinceLastCrash > 3600 {
				// it has been one hour since the last crash.. reset the count
				// to reflect on the frequency
				restartCount = 1
			} else {
				restartCount += 1
			}
		}
	}()

// Stop stops the gRPC server
func (m *NvidiaDevicePlugin) Stop() error {
	if m.server == nil {
		return nil
	}

	m.server.Stop()
	m.server = nil
	close(m.stop)

	return m.cleanup()
}
```

if Stop is called, it may panic

RenaudWasTaken:
Being addressed here: https://gitlab.com/nvidia/kubernetes/device-plugin/merge_requests/14

PR #149, opened 2019-11-28
fix: m.server.Server will panic

lujinda:
#148

### DIFF
--- a/server.go
+++ b/server.go
@@ -97,7 +97,7 @@ func (m *NvidiaDevicePlugin) Start() error {
 	go func() {
 		lastCrashTime := time.Now()
 		restartCount := 0
-		for {
+		for m.server != nil {
 			log.Println("Starting GRPC server")
 			err := m.server.Serve(sock)
 			if err != nil {


RenaudWasTaken:
@rajatchopra reviewing the following piece of code made me realize we don't have proper handling of the Stop code path:
https://github.com/NVIDIA/k8s-device-plugin/blob/master/server.go#L97-L122

@lujinda do you mind making your pull request against the gitlab repo (https://gitlab.com/nvidia/kubernetes/device-plugin)? This repo is a mirror :)

I'm not sure checking if the server is nil is the solution we want to adopt here. I'll try to think through the different code path during the long weekend and come back with a better suggestion.

lujinda:
> @rajatchopra reviewing the following piece of code made me realize we don't have proper handling of the Stop code path:
> https://github.com/NVIDIA/k8s-device-plugin/blob/master/server.go#L97-L122
> 
> @lujinda do you mind making your pull request against the gitlab repo (https://gitlab.com/nvidia/kubernetes/device-plugin)? This repo is a mirror :)
> 
> I'm not sure checking if the server is nil is the solution we want to adopt here. I'll try to think through the different code path during the long weekend and come back with a better suggestion.

Ok. I have submitted

elezar:
See https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/14

PR #164, opened 2020-04-23
Fix image tag typo

yoyota:
image

Fix beta6 to beta.6


RenaudWasTaken:
Hello and Thanks for your contribution!

I've actually fixed the tag to be `beta6` to stay consistent with the previous releases :)
If you do make a new PR in the future, you should make it against https://gitlab.com/nvidia/kubernetes/device-plugin as the github repo is a mirror!

Thanks again!

Issue #132, opened 2019-09-04
Docker image for Nvidia Jetson Nano


vdups:
Hi all, I have a feature request for using this plugin on Jetson Nano.
I hope you could help, and I'd be happy to test !

### 1. Issue or feature description
I'd like to use Kubernetes on Nvidia Jetson Nano(s)
With July 2019 release of Jetpack, nvidia-docker (in beta) works fine and is set as my default docker runtime.
But it seems that nvidia/k8s-device-plugin container image is NOT distributed for ARM64/v8 architecture. (neither for the version specified in nvidia-device-plugin.yaml nor in any version on Docker hub : only x86_64 & ppc64le)
[https://hub.docker.com/r/nvidia/k8s-device-plugin](https://hub.docker.com/r/nvidia/k8s-device-plugin)

### 2. Steps to reproduce the issue
On a Jetson Nano, (with docker daeamon started and current user in docker group)
`vdups@jetsonk3sgpu01:~$ grep image: /tmp/nvidia-device-plugin.yaml
      - image: nvidia/k8s-device-plugin:1.0.0-beta
vdups@jetsonk3sgpu01:~$ docker pull nvidia/k8s-device-plugin:1.0.0-beta
1.0.0-beta: **Pulling from nvidia/k8s-device-plugin
no matching manifest for linux/arm64/v8 in the manifest list entries**`

### 3. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)

Jetson Nano base image used : r322 downloaded in July 2019 from [https://developer.nvidia.com/jetson-nano-sd-card-image-r322](https://developer.nvidia.com/jetson-nano-sd-card-image-r322)

I hope you could provide some help. I did not found similar open issues in github.

Common error checking:
 - [ ] The output of `nvidia-smi -a` on your host
 - [ ] Your docker configuration file (e.g: `/etc/docker/daemon.json`)
 - [ ] The k8s-device-plugin container logs
 - [ ] The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)

Additional information that might help better understand your environment and reproduce the bug:
 - [ ] Docker version from `docker version`
 - [ ] Docker command, image and tag used
 - [X ] Kernel version from `uname -a`
`vdups@jetsonk3sgpu01:~$ uname -a
[...]

PR #209, opened 2020-11-25
Add support for Tegra devices

pmundt:
This builds on the approach proposed in issue #132 to provide Tegra
support for the device plugin, and expands upon it to provide a more
flexible registration and control interface, allowing both NVML and
non-NVML controlled devices to be supported through the existing
interfaces.

Tested on a Jetson Nano and Xavier NX.

Cc: Pablo Rodriguez <pablo.rodriguez-quesada@windriver.com>
Signed-off-by: Paul Mundt <paul.mundt@adaptant.io>

### DIFF
--- a/device.go
+++ b/device.go
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2019, 2020 NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2020 Adaptant Solutions AG
+ *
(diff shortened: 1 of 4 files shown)


pmundt:
Re-opened on GitLab, as per suggestion: https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/61

PR #284, opened 2021-12-14
update the Dockerfile: NVIDIA_DRIVER_CAPABILITIES=utility,compute

alex337:
There are no nvidia-uvm and nvidia-uvm-tools in the nvidia-device-plugin container , so there will be no nvidia-uvm and nvidia-uvm-tools in the containers which are created by the nvidia-device-plugin container.Add "NVIDIA_DRIVER_CAPABILITIES=utility,compute" in the Dockerfile can resolve this problem.


elezar:
Note that the NVIDIA Container Library (currently used to inject devices into a container) does not filter devices based on the `NVIDIA_DRIVER_CAPABILITIES` environment variable in the container. This also has no influence on the environment variables set in the containers launched.

If you are seeing different behaviour consider creating an issue describing what you are seeing.

PR #298, opened 2022-04-22
fix: Fixed a build error on io.ReadAll

aelgasser:
Building the plugin today fails because the function `ReadAll` is not anymore in `io` but in `io/ioutil`

**Steps to reproduce:**
1. Clone the repo
2. run `docker build -t k8s-device-plugin:devel -f deployments/container/Dockerfile.ubuntu .`
3. At some point during the build you'll see this

```bash
...
# github.com/NVIDIA/k8s-device-plugin/api/config/v1
api/config/v1/config.go:74:20: undefined: io.ReadAll
...
The command '/bin/sh -c export CGO_LDFLAGS_ALLOW='-Wl,--unresolved-symbols=ignore-in-object-files' &&     go build -ldflags="-s -w -X 'main.Version=${VERSION}'" -v ./cmd/nvidia-device-plugin' returned a non-zero code: 2
```

Signed-off-by: Abderraouf El Gasser <abderraouf.elgasser@gmail.com>
Signed-off-by: Abderraouf El Gasser <abderraouf.elgasser@iktos.com>


elezar:
@aelgasser since we have updated our golang version please use the `os` package instead of `ioutil` for this change.

aelgasser:
@elezar actually this doesn't work. running the same docker build command fails to find os.ReadFile. So I might need to update the golang version used to build?

aelgasser:
@elezar Updated golang version to latest 1.17 and it builds and runs successfully. 

I just recreated the MR in gitlab to run the full CI pipeline: https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/129

api/config/v1/config.go
```diff
@@ -53,28 +52,21 @@ type Flags struct {
...
+	config, err := parseConfigFrom(configYaml)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing config file: %v", err)
 	}
 
 	return config, nil
 }
 
-func parseConfigFrom(reader io.Reader) (*Config, error) {
+func parseConfigFrom(configYaml []byte) (*Config, error) {
```
elezar, line 68: Could we keep this using a reader? This will minimize the diff as the change should then be limited to this function. Also, looking at the `ioutil.ReadAll` documentation this mentions that this has been moved to `io.ReadAll` as of Go 1.16. Does this mean that the changes in this commit are redundant (or covered by default Golang version update)?

cmd/nvidia-device-plugin/server.go
```diff
@@ -70,6 +70,7 @@ type NvidiaDevicePlugin struct {
 func NewNvidiaDevicePlugin(config *config.Config, resourceName string, resourceManager ResourceManager, deviceListEnvvar string, allocatePolicy gpuallocator.Policy, socket string) *NvidiaDevicePlugin {
 	return &NvidiaDevicePlugin{
 		ResourceManager:  resourceManager,
+		config:           *config,
```
elezar, line 73: Good catch.

deployments/container/Dockerfile.ubi8
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG GOLANG_VERSION=1.15.8
+ARG GOLANG_VERSION=1.17.9
```
elezar, line 15: A useful addition may be a make target that checks whether this version is the same as in `version.mk` to avoid issues like this in future. (also for the CUDA version).

(Note we need this here because we provide instructions in the README to build the docker images directly).

elezar (changes requested):
Thanks for the proposal @aelgasser. I have some questions / suggestions.

api/config/v1/config.go
aelgasser: `os.ReadFile` was returning a `[]byte` directly from the filename passed to it so I couldn't keep the reader. but as you said `io.ReadAll` is available in 1.17 so I completely reverted the changes. 

I was building the repo since the beginning with the assumption that the Dockerfile.ubuntu (I found the build steps in the README file, so it might be good to note that it's meant 1.17+ only for the next ones) was correct in using golang 1.15, which is not the case anymore as you said earlier. I simply didn't connect the dots afterward to realize that I should pick function availability from 1.17/1.16 in the docs. 

For my defense, my primary objective with your repo was to implement GPU sharing, and as I found an MR (https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/86) doing just that I was simply focused in porting it to the latest version, making it work, and testing it on our k8s clusters (which was a success and it's going live in production by Monday :-) ).

cmd/nvidia-device-plugin/server.go
aelgasser: couldn't miss it: Segfault :-)

deployments/container/Dockerfile.ubi8
aelgasser: That's beyond my knowledge of makefiles (I always hated them since college :) ).
I could try to figure out something but it'll probably be an ugly and suboptimal bash script. I'd rather leave that to experts.

elezar:
Closing this since https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/129 has been merged.

PR #300, opened 2022-04-27
[add]: support for hostNetwork parameter in daemonset deployment

vasudev-singhc-by:
If hostNetwork parameter is set to true then host ip is assigned to the daemonset instead of assigning a new ip.

This can be set to true, when:

Daemonset is not externallly accessed.
It saves an Ip address.
If hostNetwork is available in values yaml. and value is 'true' then add the hostNetwork parameter to template spec.

spec:
{{- if .Values.hostNetwork }}
hostNetwork: {{ .Values.hostNetwork }}
{{- end }}

Signed-off-by: Vasudev Singh C vasudev.singhc@jda.com


ArangoGutierrez:
We can re-open this PR now that we have migrated the Repo to GItHub
@vasudev-singhc-by Please reopen at will

PR #311, opened 2022-06-17
Fix containerd config in README.md

gmrukwa:
I've tried to configure my `containerd` with the snippet you provided and spotted a typo. It's fixed here.


elezar:
This was addressed in https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/200

Issue #364, opened 2022-12-23
Time-slicing config update: "Error: error creating symlink: file exists"


Telemaco019:
### 1. Issue or feature description
When changing between different time-slicing configurations by updating the label `nvidia.com/device-plugin.config` of a node (as explained in the [documentation](https://github.com/NVIDIA/k8s-device-plugin#shared-access-to-gpus-with-cuda-time-slicing)) the nvidia-device-plugin crashes with the following error:

```
│ config-manager time="2022-12-23T13:54:55Z" level=info msg="Waiting for change to 'nvidia.com/device-plugin.config' label"                                                                                                                │
│ config-manager time="2022-12-23T13:54:55Z" level=info msg="Label change detected: nvidia.com/device-plugin.config=config-3"                                                                                                              │
│ config-manager time="2022-12-23T13:54:55Z" level=info msg="Updating to config: config-3"                                                                                                                                                 │
│ config-manager time="2022-12-23T13:54:55Z" level=info msg="Error: error creating symlink: symlink /available-configs/config-3 /config/config.yaml: file exists"
```

This happens when the ConfigMap containing the available configurations is edited by removing the entry corresponding to the current configuration, and the node label `nvidia.com/device-plugin.config` is updated with another configuration name.

The issue is probably related to the Config Manager component trying to create the symlink to the new config file without properly deleting the existing symlink pointing to the previous (deleted) configuration file, thus resulting in the "file exists" error mentioned above. 

### 2. Steps to reproduce the issue

1. Create the ConfigMap "time-slicing-config" containing 3 different time-slicing configurations
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: time-slicing-config
  namespace: gpu-operator
data:
  config-1: |
    version: v1
    sharing:
      timeSlicing:
        resources:
        - name: nvidia.com/gpu
          replicas: 2
  config-2: |
    version: v1
    sharing:
      timeSlicing:
        resources:
        - name: nvidia.com/gpu
          replicas: 4
  config-3: |
    version: v1
    sharing:
      timeSlicing:
        resources:
        - name: nvidia.com/gpu
          replicas: 8
```

2. Apply the configuration "config-1" to the Node
```bash
kubectl label node <node-name> nvidia.com/device-plugin.config=config-1
```

3. Edit the ConfigMap "time-slicing-config" deleting the entry "config-1". This is the resulting ConfigMap:
```yaml
apiVersion: v1
kind: ConfigMap
[...]

PR #365, opened 2022-12-23
#364 Check config symlink instead of file existence in config-manager

Telemaco019:
As described in issue #364, when changing value of the configuration label, the config-manager tries to create the symlink to the new config file without properly deleting the symlink pointing to the previous (deleted) configuration file, thus resulting in the "file exists" error mentioned above when trying to create the symlink to the new configuration. 

The PR makes the config-manager check for symlink existence instead of checking if the linked file exists.

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -394,18 +394,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
 		src = filepath.Join(f.ConfigFileSrcdir, config)
 	}
 
-	exists, err := fileExists(f.ConfigFileDst)
+	exists, err := symlinkExists(f.ConfigFileDst)
 	if err != nil {
-		return false, fmt.Errorf("error checking if file '%s' exists: %v", f.ConfigFileDst, err)
+		return false, fmt.Errorf("error checking if symlink '%s' exists: %v", f.ConfigFileDst, err)
 	}
 	if exists {
 		srcRealpath, err := filepath.EvalSymlinks(src)
-		if err != nil {
+		if err != nil && !os.IsNotExist(err) {
 			return false, fmt.Errorf("error evaluating realpath of '%v': %v", src, err)
 		}
 
 		dstRealpath, err := filepath.EvalSymlinks(f.ConfigFileDst)
-		if err != nil {
+		if err != nil && !os.IsNotExist(err) {
 			return false, fmt.Errorf("error evaluating realpath of '%v': %v", f.ConfigFileDst, err)
 		}
 
@@ -456,8 +456,8 @@ func findPidToSignal(f *Flags) (int, error) {
 	return -1, fmt.Errorf("no process found")
 }
 
-func fileExists(filename string) (bool, error) {
-	info, err := os.Stat(filename)
+func symlinkExists(filename string) (bool, error) {
+	info, err := os.Lstat(filename)
 	if os.IsNotExist(err) {
 		return false, nil
 	}


cmd/config-manager/main.go
```diff
@@ -456,8 +456,8 @@ func findPidToSignal(f *Flags) (int, error) {
 	return -1, fmt.Errorf("no process found")
 }
 
-func fileExists(filename string) (bool, error) {
-	info, err := os.Stat(filename)
+func symlinkExists(filename string) (bool, error) {
+	info, err := os.Lstat(filename)
```
elezar, line 460: Although this change makes sense, is the `return !info.IsDir(), nil` at the end of the function make sense?

cmd/config-manager/main.go
```diff
@@ -394,18 +394,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
...
-	exists, err := fileExists(f.ConfigFileDst)
+	exists, err := symlinkExists(f.ConfigFileDst)
 	if err != nil {
-		return false, fmt.Errorf("error checking if file '%s' exists: %v", f.ConfigFileDst, err)
+		return false, fmt.Errorf("error checking if symlink '%s' exists: %v", f.ConfigFileDst, err)
 	}
 	if exists {
 		srcRealpath, err := filepath.EvalSymlinks(src)
-		if err != nil {
+		if err != nil && !os.IsNotExist(err) {
```
elezar, line 403: Since we're creating a symlink pointing at `src` here, does that not mean that non-existence is an error too?

Telemaco019: It's been quite a long time and I honestly can't remember why I put that condition. 

Reviewing it now, I agree it doesn't make much sense as we want to check whether the file referenced by the link exists. I'm pushing a commit to remove it

cmd/config-manager/main.go
```diff
@@ -394,18 +394,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
...
 			return false, fmt.Errorf("error evaluating realpath of '%v': %v", src, err)
 		}
 
 		dstRealpath, err := filepath.EvalSymlinks(f.ConfigFileDst)
-		if err != nil {
+		if err != nil && !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
Telemaco019: You're right, this is just redundant, I'm removing it

cmd/config-manager/main.go
Telemaco019: I think it does, as in the `updateSymlink` function we want to point to a config file, not to a directory. 

Maybe the name of the function is a bit misleading and could be something like `symlinkFileExists`, what do you think?

elezar: I suppose my quesion should then be, why not check the symlink mode in `info`:
```
	info, err := os.Lstat(filename)
	if err != nil {
		return filename, fmt.Errorf("failed to get file info: %v", info)
	}
	if info.Mode()&os.ModeSymlink == 0 {
		return filename, fmt.Errorf("not a symlink: %v", filename)
	}
```

Telemaco019: If we just check the mode it means that, when`filename` is a path to a symlink pointing to a directory, this function returns `true`, right? 

If so, is this the intended behavior? Updating the config-manager providing a symlink that points to a directory would make it break, so it makes sense to check whether the provided symlink actually points to a file. Maybe I'm missing something?

PR #370, opened 2022-12-29
Update README.md

hholst80:
/etc/apt/trusted.gpg.d/libnvidia-container.gpg Is not really the right place for this but the upstream .list file does not have a path specified and relies on the built-in apt database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,12 @@ Please see: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/in
 ##### Install the `nvidia-container-toolkit`
 ```bash
 distribution=$(. /etc/os-release;echo $ID$VERSION_ID)
-curl -s -L https://nvidia.github.io/libnvidia-container/gpgkey | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,10 +67,12 @@ Please see: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/in
 ##### Install the `nvidia-container-toolkit`
 ```bash
```
elezar, line 68: I would ideally refer to https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html here directly instead of repeating what is there.

README.md
```diff
@@ -92,6 +94,9 @@ And then restart `docker`:
 ```
 $ sudo systemctl restart docker
 ```
+Note that the runtime can be selected with `--runtime=runc` parameter
```
elezar, line 97: Let's also add a note that the use of docker with Kubernetes is deprecated.

elezar (changes requested):

elezar:
I updated the readme in https://github.com/NVIDIA/k8s-device-plugin/pull/561

Issue #373, opened 2023-01-04
container failed to start after the VM node migrated to another host 


borgerli:
### 1. Issue or feature description
When there is some problem with host, the VM node runing on it will be migrated to another health host. After migration, the containers of GPU pods of the node fails to start with following errors:
```
OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: device error: GPU-12345678-1234-1234-1234-1234567890ab: unknown device: unknown
```
### 2. Steps to reproduce the issue
1. Create a GPU pod with nvidia.com/gpu request
2. Migrate the node where the pod in running, to another host
3. After migration, the pod fails to start with above error.

### 3. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)

Common error checking:
 - [ ] The output of `nvidia-smi -a` on your host
 - [ ] Your docker configuration file (e.g: `/etc/docker/daemon.json`)
 - [ ] The k8s-device-plugin container logs
 - [ ] The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)

Additional information that might help better understand your environment and reproduce the bug:
 - [ ] Docker version from `docker version`
 - [ ] Docker command, image and tag used
 - [ ] Kernel version from `uname -a`
 - [ ] Any relevant kernel output lines from `dmesg`
 - [ ] NVIDIA packages version from `dpkg -l '*nvidia*'` _or_ `rpm -qa '*nvidia*'`
 - [ ] NVIDIA container library version from `nvidia-container-cli -V`
 - [ ] NVIDIA container library logs (see [troubleshooting](https://github.com/NVIDIA/nvidia-docker/wiki/Troubleshooting))

klueska:
When you move to a new physical machine, the UUIDs of the underlying GPUs will change (causing this issue). 

To avoid this, you will need to use a different `device-id-strategy` for the plugin (e.g. `index` instead of `uuid`):
https://github.com/NVIDIA/k8s-device-plugin#as-command-line-flags-or-envvars

borgerli:
Great! Thanks.

borgerli:
[...]

PR #386, opened 2023-02-25
report index as RegisteredDevices when device-id-strategy set to index

borgerli:
fix issue https://github.com/NVIDIA/k8s-device-plugin/issues/373


borgerli:
With this PR, when `device-id-strategy` set to `index`, indexes, instead of UUIDs will be reported to kubelet and informations in checkpoint file looks likes this `"RegisteredDevices":{"nvidia.com/gpu":["0","1"]}`.

borgerli:
@klueska, could you plelase take a look? Thanks.

klueska:
Could you please open an MR against https://gitlab.com/nvidia/kubernetes/device-plugin. This is a read-only mirror.

borgerli:
MR on gitlab:https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/259

PR #431, opened 2023-08-08
Update readme file with valid device-plugin KEP link

levikobi:
The former link was invalid, updated the README.md file with the up-to-date one.


levikobi:
Just noticed it's a duplicate of #411

PR #460, opened 2023-11-30
Return all available directly if preferred allocating all

zwpaper:
We found retrieving all GPU statuses will take seconds in nodes with multiple GPUs when kubelet calls `GetPreferredAllocation`.

and if the available ones are all the kubelet requesting, maybe the device plugins can return the available ones directly.

we made some changes and found it was working as expected much faster.

so we are raising up this PR to discuss this solution:

1. is it ok to skip the following nvml detections in `GetPreferredAllocation`?
2. are there any other things we need to consider before adding this change?

### DIFF
--- a/internal/rm/nvml_manager.go
+++ b/internal/rm/nvml_manager.go
@@ -112,6 +112,11 @@ func (r *nvmlResourceManager) getPreferredAllocation(available, required []strin
 // alignedAlloc shells out to the alignedAllocationPolicy that is set in
 // order to calculate the preferred allocation.
 func (r *nvmlResourceManager) alignedAlloc(available, required []string, size int) ([]string, error) {
+	// return all available directly if allocating all available devices
+	if len(available) == size {
+		return available, nil
+	}
+
 	var devices []string
 
 	linkedDevices, err := gpuallocator.NewDevices(


internal/rm/nvml_manager.go
```diff
@@ -112,6 +112,11 @@ func (r *nvmlResourceManager) getPreferredAllocation(available, required []strin
 // alignedAlloc shells out to the alignedAllocationPolicy that is set in
 // order to calculate the preferred allocation.
 func (r *nvmlResourceManager) alignedAlloc(available, required []string, size int) ([]string, error) {
+	// return all available directly if allocating all available devices
```
elezar, line 115: One comment / question I have is whether this is applicable to all allocation types, meaning that we should check this in `getPreferredAllocation` or in `GetPreferredAllocation`?

elezar:
Note that this repository is a read-only mirror of https://gitlab.com/nvidia/kubernetes/device-plugin.

If possible, please create a merge request there instead and close this PR.

zwpaper:
Hi @elezar, thanks for the info, created one at https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/339

elezar:
@zwpaper sorry for the back and forth. We are in the process of migrating to GitHub as our primary repository. I am reopening this PR and closing the GitLab one.

internal/rm/nvml_manager.go
```diff
@@ -112,6 +112,11 @@ func (r *nvmlResourceManager) getPreferredAllocation(available, required []strin
...
+	if len(available) == size {
```
elezar, line 116: Looking at this again, this is not correct, because it doen't take the `required` slice into account.

I would recommend:
1. Moving this check to `getPreferredAllocation` and making the call to `r.alignedAlloc` conditional on `len(available) > size`.
2. Extending `func (r *resourceManager) distributedAlloc(available, required []string, size int) ([]string, error) {` to include a check that matches this *after*  the `candidates` have been determined.

elezar (changes requested):
Please see the comments.

PR #472, opened 2023-12-27
Use WithContextDialer to replace WithDialer

kerthcet:
Cleanup for TODO.

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -457,15 +457,13 @@ func (plugin *NvidiaDevicePlugin) PreStartContainer(context.Context, *pluginapi.
 
 // dial establishes the gRPC communication with the registered device plugin.
 func (plugin *NvidiaDevicePlugin) dial(unixSocketPath string, timeout time.Duration) (*grpc.ClientConn, error) {
-	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 	c, err := grpc.DialContext(ctx, unixSocketPath,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithBlock(),
-		// TODO: We need to switch to grpc.WithContextDialer.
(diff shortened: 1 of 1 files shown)


kerthcet:
close in favor of https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/345

elezar:
@kerthcet please reopen this. We are migrating to GitHub and I will review it here.

internal/plugin/server.go
```diff
@@ -392,15 +392,13 @@ func (plugin *NvidiaDevicePlugin) PreStartContainer(context.Context, *pluginapi.
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 	c, err := grpc.DialContext(ctx, unixSocketPath,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithBlock(),
-		// TODO: We need to switch to grpc.WithContextDialer.
-		//nolint:staticcheck
-		grpc.WithDialer(func(addr string, timeout time.Duration) (net.Conn, error) {
-			return net.DialTimeout("unix", addr, timeout)
+		grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
```
elezar, line 400: One question I have: How does the change in default TCP timeout as described in https://pkg.go.dev/google.golang.org/grpc#WithContextDialer
> Note: All supported releases of Go (as of December 2023) override the OS defaults for TCP keepalive time and interval to 15s. To enable TCP keepalive with OS defaults for keepalive time and interval, use a net.Dialer that sets the KeepAlive field to a negative value, and sets the SO_KEEPALIVE socket option to true from the Control field. For a concrete example of how to do this, see internal.NetDialerWithTCPKeepalive().

affect this change here?

kerthcet: I think they're different, one is for detecting network liveness, another is for waiting for the network connection. `WithDialer` also uses `WithContextDialer` underlying.

elezar (approved):

PR #473, opened 2024-01-02
Fix: counting server restart times error

kerthcet:
restartCount starts with 0, before, we'll crash 7 times then quit. Now, we'll quit at 6th.

/kind bug

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -163,7 +163,15 @@ func (plugin *NvidiaDevicePlugin) Serve() error {
 	go func() {
 		lastCrashTime := time.Now()
 		restartCount := 0
+
 		for {
+			// quite if it has been restarted too often
+			// i.e. if server has crashed more than 5 times and it didn't last more than one hour each time
+			if restartCount > 5 {
+				// quit
+				klog.Fatalf("GRPC server for '%s' has repeatedly crashed recently. Quitting", plugin.rm.Resource())
+			}
+
 			klog.Infof("Starting GRPC server for '%s'", plugin.rm.Resource())
 			err := plugin.server.Serve(sock)
 			if err == nil {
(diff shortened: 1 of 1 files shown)


kerthcet:
close in favor of https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/344

elezar:
@kerthcet sorry for the back and forth on this. We are in the process of migrating our repos to GitHub. Would it be possible to reopen this as a PR and close https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/344 instead?

kerthcet:
Glad to hear that. reopened.

internal/plugin/server.go
```diff
@@ -174,7 +174,7 @@ func (plugin *NvidiaDevicePlugin) Serve() error {
 
 			// restart if it has not been too often
 			// i.e. if server has crashed more than 5 times and it didn't last more than one hour each time
-			if restartCount > 5 {
+			if restartCount >= 5 {
 				// quit
 				klog.Fatalf("GRPC server for '%s' has repeatedly crashed recently. Quitting", plugin.rm.Resource())
 			}
```
elezar, line 180: Let's either move this to the start or the end of the loop to make it more consistent. This should also allow us to keep `restartCount > 5` as the conditional.

internal/plugin/server.go
```diff
@@ -183,14 +183,14 @@ func (plugin *NvidiaDevicePlugin) Serve() error {
 			if timeSinceLastCrash > 3600 {
 				// it has been one hour since the last crash.. reset the count
 				// to reflect on the frequency
-				restartCount = 1
+				restartCount = 0
```
elezar, line 186: 👍

kerthcet:
Done @elezar

internal/plugin/server.go
kerthcet: Done

elezar (approved):

PR #489, opened 2024-01-28
Device plugin cnt 3924 docs

elezar:
Migrated from https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/304


ArangoGutierrez:
Should we revisit this?

elezar:
> Should we revisit this?

Sure. Could we have a single commit that *ADDS* the GFD documentation. We could have follow-up commits / PRs that better organize the documentation.

PR #495, opened 2024-02-01
Move CHANGELOG from README to separate file

elezar:
This should simplify future version bumps.


ArangoGutierrez (approved):
Welcomed idea

PR #496, opened 2024-02-01
Move fs and signal handling to internal pkg

elezar:
The same or similar logic is also required by #490.


ArangoGutierrez:
Is there a remote possibility someone is already depending on this PKG? I don't have objections when going private to pkg, but the other direction could affect projects importing the module. (Not a blocker, just a consideration)

elezar:
> Is there a remote possibility someone is already depending on this PKG? I don't have objections when going private to pkg, but the other direction could affect projects importing the module. (Not a blocker, just a consideration)

The functions were `newOSWatcher` and `newFSWatcher`. These are private. If anyone was depending on the package they would not be consuming these function directly. Also since these functions were defined in the `cmd/*` packages, I would already consider them implicitly internal.

ArangoGutierrez (approved):

PR #497, opened 2024-02-01
Remove unused restarting argument

elezar:
This change removes an unused restarting argument from the startPlugins function and renames a local variable in the main loop for clarity.

This unused parameter was noticed in #490.


cmd/nvidia-device-plugin/main.go
```diff
@@ -163,31 +163,30 @@ func start(c *cli.Context, flags []cli.Flag) error {
...
 		}
 	}
 
 	klog.Info("Starting Plugins.")
-	plugins, restartPlugins, err := startPlugins(c, flags, restarting)
+	plugins, restartPlugins, err := startPlugins(c, flags)
 	if err != nil {
 		return fmt.Errorf("error starting plugins: %v", err)
 	}
+	started = true
```
ArangoGutierrez, line 183: Is there a use for this variable? I can't see it being used after this line

ArangoGutierrez (approved):

PR #503, opened 2024-02-06
Add resource.sharing-strategy labels

elezar:
This change adds sharing-strategy labels per resource.

This label can have the value: none, mps, time-slicing depending on the sharing configuration. For invalid configurations, this label is empty.

Running against the in-tree kind cluster:

1. Deploy GFD only and get the labels:
```
helm upgrade nvidia -i deployments/helm/nvidia-device-plugin \
    --namespace nvidia-device-plugin \
    --create-namespace \
    --set runtimeClassName=nvidia \
    --set nvidiaDriverRoot=/ \
    --set devicePlugin.enabled=false \
    --set gfd.enabled=true
```

```
kubectl get node k8s-device-plugin-cluster-worker  --output=json | jq '.metadata.labels' | grep "nvidia.com" | sort
  "nvidia.com/cuda.driver.major": "515",
  "nvidia.com/cuda.driver.minor": "105",
  "nvidia.com/cuda.driver.rev": "01",
  "nvidia.com/cuda.runtime.major": "11",
  "nvidia.com/cuda.runtime.minor": "7",
  "nvidia.com/gfd.timestamp": "1707221624",
  "nvidia.com/gpu.compute.major": "7",
  "nvidia.com/gpu.compute.minor": "0",
  "nvidia.com/gpu.count": "8",
  "nvidia.com/gpu.family": "volta",
  "nvidia.com/gpu.machine": "kind",
  "nvidia.com/gpu.memory": "16384",
  "nvidia.com/gpu.product": "Tesla-V100-SXM2-16GB-N",
  "nvidia.com/gpu.replicas": "1",
  "nvidia.com/gpu.sharing-strategy": "none",
  "nvidia.com/mig.capable": "false",
  "nvidia.com/sharing.mps.enabled": "false"
```

2. Configure MPS sharing:
```
cat << EOF > dp-mps-config.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  nvidiaDriverRoot: "/"
  plugin:
    passDeviceSpecs: false
    deviceListStrategy:
    - cdi-annotations
    deviceIDStrategy: uuid
sharing:
  mps:
    renameByDefault: false
    resources:
    - name: nvidia.com/gpu
      replicas: 4
EOF
kubectl create cm -n nvidia-device-plugin nvidia-plugin-mps \
    --from-file=config=dp-mps-config.yaml
```

Deploy using the config:
```
helm upgrade nvidia -i deployments/helm/nvidia-device-plugin \
    --namespace nvidia-device-plugin \
    --create-namespace \
    --set runtimeClassName=nvidia \
    --set nvidiaDriverRoot=/ \
    --set devicePlugin.enabled=false \
    --set gfd.enabled=true \
    --set config.name=nvidia-plugin-mps
```

Get the labels:
```
kubectl get node k8s-device-plugin-cluster-worker  --output=json | jq '.metadata.labels' |
 grep "nvidia.com" | sort
  "nvidia.com/cuda.driver.major": "515",
  "nvidia.com/cuda.driver.minor": "105",
  "nvidia.com/cuda.driver.rev": "01",
  "nvidia.com/cuda.runtime.major": "11",
  "nvidia.com/cuda.runtime.minor": "7",
  "nvidia.com/gfd.timestamp": "1707222576",
  "nvidia.com/gpu.compute.major": "7",
  "nvidia.com/gpu.compute.minor": "0",
  "nvidia.com/gpu.count": "8",
  "nvidia.com/gpu.family": "volta",
  "nvidia.com/gpu.machine": "kind",
  "nvidia.com/gpu.memory": "16384",
  "nvidia.com/gpu.product": "Tesla-V100-SXM2-16GB-N-SHARED",
  "nvidia.com/gpu.replicas": "4",
  "nvidia.com/gpu.sharing-strategy": "mps",
  "nvidia.com/mig.capable": "false",
  "nvidia.com/sharing.mps.enabled": "true"
```

3. Configure MPS sharing with renaming:
```
cat << EOF > dp-mps-rename-config.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  nvidiaDriverRoot: "/"
  plugin:
    passDeviceSpecs: false
    deviceListStrategy:
    - cdi-annotations
    deviceIDStrategy: uuid
sharing:
  mps:
    renameByDefault: true
    resources:
    - name: nvidia.com/gpu
      replicas: 4
EOF
kubectl create cm -n nvidia-device-plugin nvidia-plugin-mps-rename \
    --from-file=config=dp-mps-rename-config.yaml
```

Deploy:
```
helm upgrade nvidia -i deployments/helm/nvidia-device-plugin \
    --namespace nvidia-device-plugin \
    --create-namespace \
    --set runtimeClassName=nvidia \
    --set nvidiaDriverRoot=/ \
    --set devicePlugin.enabled=false \
    --set gfd.enabled=true \
    --set config.name=nvidia-plugin-mps-rename
```

Get the labels:
```
kubectl get node k8s-device-plugin-cluster-worker  --output=json | jq '.metadata.labels' | grep "nvidia.com" | sort
  "nvidia.com/cuda.driver.major": "515",
  "nvidia.com/cuda.driver.minor": "105",
  "nvidia.com/cuda.driver.rev": "01",
  "nvidia.com/cuda.runtime.major": "11",
  "nvidia.com/cuda.runtime.minor": "7",
  "nvidia.com/gfd.timestamp": "1707222920",
  "nvidia.com/gpu.compute.major": "7",
  "nvidia.com/gpu.compute.minor": "0",
  "nvidia.com/gpu.count": "8",
  "nvidia.com/gpu.family": "volta",
  "nvidia.com/gpu.machine": "kind",
  "nvidia.com/gpu.memory": "16384",
  "nvidia.com/gpu.product": "Tesla-V100-SXM2-16GB-N",
  "nvidia.com/gpu.replicas": "4",
  "nvidia.com/gpu.sharing-strategy": "mps",
  "nvidia.com/mig.capable": "false",
  "nvidia.com/sharing.mps.enabled": "true"
```


deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -213,6 +212,7 @@ spec:
           hostPath:
           # TODO: This should be /var/run/nvidia/mps
             path: /var/lib/kubelet/device-plugins/mps
+            type: DirectoryOrCreate
```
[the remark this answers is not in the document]
elezar: Created #505

internal/lm/resource_test.go
```diff
@@ -41,87 +41,168 @@ func TestGPUResourceLabeler(t *testing.T) {
...
+							Replicas: 2,
+						},
+					},
+				},
+			},
+			expectedLabels: Labels{
+				"nvidia.com/gpu.count":            "1",
+				"nvidia.com/gpu.replicas":         "2",
+				"nvidia.com/gpu.sharing-strategy": "mps",
+				"nvidia.com/gpu.memory":           "300",
```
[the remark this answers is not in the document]
cdesiniotis: Or maybe have two separate labels, one representing total memory and another representing memory per replica?

tests/e2e/gpu-feature-discovery_test.go
```diff
@@ -43,22 +43,23 @@ var _ = NVDescribe("GPU Feature Discovery", func() {
...
+		"nvidia.com/gpu.count":            "[0-9]+",
+		"nvidia.com/gpu.replicas":         "[0-9]+",
+		"nvidia.com/gpu.sharing-strategy": "[none|mps|time-slicing]",
+		"nvidia.com/gpu.product":          "[A-Za-z_-]+",
+		"nvidia.com/gpu.memory":           "[0-9]+",
+		"nvidia.com/gpu.family":           "[a-z]+",
+		"nvidia.com/mig.capable":          "[true|false]",
+		"nvidia.com/gpu.compute.major":    "[0-9]+",
+		"nvidia.com/gpu.compute.minor":    "[0-9]+",
+		"nvidia.com/sharing.mps.enabled":  "[true|false]",
```
cdesiniotis, line 62: Question -- what component is adding this label to the node?

cdesiniotis: Okay I see this was added to GFD in a previous commit: https://github.com/NVIDIA/k8s-device-plugin/blob/main/internal/lm/nvml.go#L149

Question -- do we still need `nvidia.com/sharing.mps.enabled`? AFAIK we use this as the nodeSelector for the MPS control daemon. Is it sufficient to just use `nvidia.com/gpu.sharing-strategy=mps`?

elezar: The issue is that the `sharing-strategy` label is per-resource. This means that if we enable resource renaming or start allowing MPS with mixed mig mode, the label to trigger off is not known.

I would say that `nvidia.com/sharing.mps.enabled` is akn to `nvidia.com/mig.capable` in this respect.

@klueska I know that you expressed some concerns over the exact label. Do you have any other suggestions for the label? Should we just make it `nvidia.com/mps.capable` to mirror mig and to indicate that the final decision depends on whether a number of replicas greater than 1 is actually selected?

internal/lm/resource_test.go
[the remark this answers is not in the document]
elezar: I don't know about a separate label. What would this label signify when `time-slicing` is selected, for example?

tests/e2e/gpu-feature-discovery_test.go
klueska: `mig.capable` is different than `nvidia.com/sharing.mps.enabled` in that it says that the GPUs on a node are _capable_ of being split into MIG devices -- it doesn't imply that they are in MIG mode or that there are any MIG devices currently configured.

However it is similar in that it is primarily used to signal that the MIG manager should be deployed on the node (whereas we want something to signal that an MPS daemon should be started for this resource type).

I think dropping `nvidia.com/sharing.mps.enabled` and adding a `nvidia.com/mps.capable` seems reasonable.

Just to make sure we are on the same page -- this would get set if there is a `sharing.mps` section in the config, and that would trigger the MPS daemonset to be deployed (similar to MIG manager) and then start one MPS control daemon per resources type that has replicas > 1. Is that correct?

elezar: To start: It could also be `nvidia.com/mps.enabled` instead of `nvidia.com/mps.capable`.

We currently trigger the `nvidia.com/sharing.mps.enabled` label if the `sharing.mps` section is present **and** at lease one resource pattern has a replica > 1. This controls the deployment of the MPS daemonset.

In this daemonset the pattern is evaluated against actual resources and if any of these have replicas > 1 then a `nvidia-cuda-mps-control` daemon is started for this resource. A further note is that we currently only support a single resource name and as such only start a single daemon. The plan is to extend this once we have more user feedback / experience.

klueska: I think we should either make it `nvidia.com/mps.capable` to mimic what we have for MIG, or just stick with `nvidia.com/sharing.mps.enabled`. I think that having  `nvidia.com/mig.capable`, but `nvidia.com/mps.enabled` would be confusing, especially since MPS can be layered on top of MIG.

Speaking of, are there any plans for an equivalent label for `timeSlicing`?  I know there is no daemon that needs to be started in this respect, but does it make sense to add for symmetry?

elezar: Fair point on using `nvidia.com/mps.capable`. 

There are no current plans for `timeSlicing`, but we can add it. There are, for example, some devices that don't support time slicing and adding this label may make sense there. (see https://github.com/NVIDIA/k8s-dra-driver/pull/58).

I can add an issue to create this label as a follow-up.

internal/lm/resource.go
```diff
@@ -155,14 +155,35 @@ func (rl resourceLabeler) key(suffix string) string {
...
-	)
+	replicas := rl.getReplicas()
+	strategy := spec.SharingStrategyNone
+	if rl.sharing != nil && replicas > 1 {
+		strategy = rl.sharing.SharingStrategy()
+	}
+	rawLabels := map[string]interface{}{
+		"product":          rl.getProductName(parts...),
+		"count":            count,
+		"replicas":         rl.getReplicas(),
```
klueska, line 166:
```suggestion
		"replicas":         replicas,
```

internal/lm/resource_test.go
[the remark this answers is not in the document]
klueska: Let's do this as a follow-up.

klueska (approved):

internal/lm/resource_test.go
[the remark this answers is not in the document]
elezar: https://github.com/NVIDIA/k8s-device-plugin/issues/535

PR #505, opened 2024-02-06
Fix mps deployment

elezar:
When deployed without MPS enabled, the device plugin fails due to mounts not being creatable.


deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -186,7 +186,6 @@ spec:
           # This is required for CDI detection to work correctly.
           - name: driver-root
             mountPath: /driver-root
-            readOnly: true
```
cdesiniotis, line 189: Question -- why is this no longer read only?

elezar: The issue is that the mount to `/driver-root/mps` fails if this is a readonly path. 

Do you have suggestions on how we can get around this?

PR #506, opened 2024-02-08
Allow mps root to be specified

elezar:
This change allows the MPS root on the host to be specified and uses /run/nvidia/mps by default.

This requires #516

### DIFF
--- a/cmd/mps-control-daemon/mps/manager.go
+++ b/cmd/mps-control-daemon/mps/manager.go
@@ -85,7 +85,7 @@ func (m *manager) Daemons() ([]*Daemon, error) {
 			klog.InfoS("Resource is not shared", "resource", "resource", resourceManager.Resource())
 			continue
 		}
-		daemon := NewDaemon(resourceManager)
+		daemon := NewDaemon(resourceManager, ContainerRoot)
 		daemons = append(daemons, daemon)
 	}
 
--- a/cmd/mps-control-daemon/mps/root.go
+++ b/cmd/mps-control-daemon/mps/root.go
@@ -1,0 +1,59 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package mps
+
+import (
+	"path/filepath"
+
+	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
+)
+
+const (
+	ContainerRoot = Root("/mps")
+)
+
+// Root represents an MPS root.
+// This is where per-resource pipe and log directories are created.
+// For containerised applications the host root is typically mounted to /mps in the container.
+type Root string
+
+// LogDir returns the per-resource pipe dir for the specified root.
+func (r Root) LogDir(resourceName spec.ResourceName) string {
+	return r.Path(string(resourceName), "log")
+}
+
+// PipeDir returns the per-resource pipe dir for the specified root.
+func (r Root) PipeDir(resourceName spec.ResourceName) string {
+	return r.Path(string(resourceName), "pipe")
+}
+
+// ShmDir returns the shm dir associated with the root.
+// Note that the shm dir is the same for all resources.
+func (r Root) ShmDir(resourceName spec.ResourceName) string {
+	return r.Path("shm")
+}
+
+// startedFile returns the per-resource .started file name for the specified root.
+func (r Root) startedFile(resourceName spec.ResourceName) string {
+	return r.Path(string(resourceName), ".started")
+}
+
+// Path returns a path relative to the MPS root.
+func (r Root) Path(parts ...string) string {
+	pathparts := append([]string{string(r)}, parts...)
+	return filepath.Join(pathparts...)
+}
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
@@ -194,12 +194,11 @@ spec:
       volumes:
       - name: mps-root
         hostPath:
-        # TODO: This should be /var/run/nvidia/mps
-          path: /var/lib/kubelet/device-plugins/mps
+          path: {{ .Values.mps.root }}
           type: DirectoryOrCreate
       - name: mps-shm
         hostPath:
-          path: /var/lib/kubelet/device-plugins/mps/shm
+          path: {{ .Values.mps.root }}/shm
       {{- if eq $hasConfigMap "true" }}
       - name: available-configs
         configMap:
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -63,6 +63,9 @@ type NvidiaDevicePlugin struct {
 	server *grpc.Server
 	health chan *rm.Device
 	stop   chan interface{}
+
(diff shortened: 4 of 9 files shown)


elezar:
Closes https://github.com/NVIDIA/cloud-native-team/issues/21

deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
```diff
@@ -193,12 +193,11 @@ spec:
       volumes:
       - name: mps-root
         hostPath:
-        # TODO: This should be /var/run/nvidia/mps
-          path: /var/lib/kubelet/device-plugins/mps
+          path: {{ .Values.mps.root }}
```
cdesiniotis, line 196: Question -- I notice that `mps-root` is mounted at `/driver-root/mps` in the container. Does this mean that the MPS root will be visible on the host at the configured `mpsRoot` (e.g. `/run/nvidia/mps`) and at `${driverRoot}/mps` (e.g. `/run/nvidia/driver/mps`)?

elezar: Running in my kind cluster, it would appear as if there is indeed a folder created at `/driver-root/mps` on the host. (`/mps` in this case). There is *also* a folder created at `/run/nvidia/mps` on the host -- as expected -- and this is the folder that contains the `shm` directory and the per-resource `pipe` and `log` directories.

cdesiniotis: Okay thanks.

internal/plugin/server.go
```diff
@@ -151,7 +151,9 @@ func (plugin *NvidiaDevicePlugin) waitForMPSDaemon() error {
 	}
 	// TODO: Check the started file here.
 	// TODO: Have some retry strategy here.
-	if err := mps.NewDaemon(plugin.rm).AssertHealthy(); err != nil {
+	// TODO: /mps here is represents the path relative to the /driver-root.
```
cdesiniotis, line 154: This comment is not correct now that you have rebased this branch of https://github.com/NVIDIA/k8s-device-plugin/pull/516, right?

Also, we are hardcoding `/mps` here since that is always the containerPath, correct?

cdesiniotis (approved):
LGTM besides one minor nit / question regarding a comment.

internal/plugin/server.go
elezar: Yes. This would need a rebase on #516. I will update the comment then.

internal/plugin/server.go
```diff
@@ -331,6 +333,9 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 	if err != nil {
 		return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
 	}
+	if err := plugin.updateContainerAllocatResponseForMPS(&response, deviceIDs); err != nil {
```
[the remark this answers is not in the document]
elezar: It's just been rebased of main.

internal/plugin/server.go
```diff
@@ -361,26 +362,31 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
...
+	// TODO: We should check that the deviceIDs are shared using MPS.
+
+	// TODO: We use the Daemon here just to construct the pipe and log dirs for the specified resource.
+	containerMpsDaemon := mps.NewDaemon(plugin.rm, "/mps")
+
+	containerPipeDir := containerMpsDaemon.PipeDir()
+	containerLogDir := containerMpsDaemon.LogDir()
+
+	response.Envs["CUDA_MPS_PIPE_DIRECTORY"] = containerPipeDir
+	response.Envs["CUDA_MPS_LOG_DIRECTORY"] = containerLogDir
```
klueska, line 374: Let's clean this up before merging.

cmd/mps-control-daemon/mps/manager.go
```diff
@@ -85,7 +85,7 @@ func (m *manager) Daemons() ([]*Daemon, error) {
 			klog.InfoS("Resource is not shared", "resource", "resource", resourceManager.Resource())
 			continue
 		}
-		daemon := NewDaemon(resourceManager)
+		daemon := NewDaemon(resourceManager, "/mps")
```
klueska, line 88: Let's make this a constant

internal/plugin/server.go
elezar: I've reworked this, reusing the `mpsDaemon` from plugin construction and adding an `mps.Root` type.

klueska: Did you forget to push?

elezar: SHould be updated now.

klueska: I still don't see the update

elezar: 🤦  let's blame the flight.

internal/plugin/server.go
```diff
@@ -361,26 +374,24 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
...
 		&pluginapi.Mount{
-			ContainerPath: logDir,
-			HostPath:      filepath.Join("/var/lib/kubelet/device-plugins", logDir),
+			ContainerPath: plugin.mpsDaemon.PipeDir(),
+			HostPath:      plugin.mpsHostRoot.LogDir(resourceName),
 		},
-	)
-	response.Mounts = append(response.Mounts,
 		&pluginapi.Mount{
 			ContainerPath: "/dev/shm",
```
klueska, line 393: Likewise have this be what is returned from a call to `plugin.mpsDaemon.ShmDir()`.

cmd/mps-control-daemon/mps/root.go
```diff
@@ -0,0 +1,53 @@
...
+// PipeDir returns the per-resource pipe dir for the specified root.
+func (r Root) PipeDir(resourceName spec.ResourceName) string {
+	return r.Path(string(resourceName), "pipe")
+}
+
+// LogDir returns the per-resource pipe dir for the specified root.
+func (r Root) LogDir(resourceName spec.ResourceName) string {
+	return r.Path(string(resourceName), "log")
+}
+
```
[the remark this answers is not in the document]
elezar: The issue is that we can't have a per-resource shm. The shm MUST exist at `/dev/shm` in the container and as such it is not rooted at the mps Root. (I suppose technically, the mount is created at `/mps/shm` in the init container (`/run/nvidia/mps/shm` on the host) and mounted to `/dev/shm` in both the daemonset container and the device plugin container.

Maybe what I'm trying to say, is that although symmetry would be nice, it isn't possible due to the fundamental differences between shm and the other two dirs.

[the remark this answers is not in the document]
klueska: I understand that, but why can't `plugin.mpsDaemon.ShmDir()` return `/dev/shm` and `plugin.mpsRoot.ShmDir(resource)` return r.Path("shm")?

I'm not talking about symmetry on the location that gets returned, but rather symmetry on the functions that are called to get the info that we need.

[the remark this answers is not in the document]
elezar: OK. That's fair. I can update.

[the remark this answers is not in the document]
elezar: I have updated. I'm not 100% happy with requiring a resource name to `Root.ShmDir()` since it may give the caller the impression that these are different accross resources. Not a blocker though.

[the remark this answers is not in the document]
klueska: It doesn't give that impression to me, necessarily. It just says -- for this resource, what is the `shm` directory. The fact that it happens to be the same (or happens to be different) for all resources shouldn't matter to the caller.

internal/plugin/server.go
```diff
@@ -148,11 +161,12 @@ func (plugin *NvidiaDevicePlugin) waitForMPSDaemon() error {
 	if plugin.config.Sharing.SharingStrategy() != spec.SharingStrategyMPS {
 		return nil
 	}
-	// TODO: Check the started file here.
+	// TODO: Check the .ready file here.
```
klueska, line 164: I this now called `.started`?

klueska (approved):

PR #507, opened 2024-02-08
Add MPS sharing section to README

elezar:
Closes https://github.com/NVIDIA/cloud-native-team/issues/19

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@
   * [As command line flags or envvars](#as-command-line-flags-or-envvars)
   * [As a configuration file](#as-a-configuration-file)
   * [Configuration Option Details](#configuration-option-details)
-  * [Shared Access to GPUs with CUDA Time-Slicing](#shared-access-to-gpus-with-cuda-time-slicing)
+  * [Shared Access to GPUs](#shared-access-to-gpus)
+    * [With CUDA Time-Slicing](#with-cuda-time-slicing)
+    * [With CUDA MPS](#with-cuda-mps)
 - [Deployment via `helm`](#deployment-via-helm)
   * [Configuring the device plugin's `helm` chart](#configuring-the-device-plugins-helm-chart)
     + [Passing configuration to the plugin via a `ConfigMap`.](#passing-configuration-to-the-plugin-via-a-configmap)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -348,18 +350,107 @@ options outside of this section are shared.
...
 granted replicas from the same underlying GPU, and each workload has access to
 the GPU memory and runs in the same fault-domain as of all the others (meaning
 if one workload crashes, they all do).
 
+In the case of MPS, a control daemon is used to manage access to the shared GPU. This daemon also allows for memory and compute limits to be enforced per workload.
+
+### With CUDA MPS
+
+**Note:** Using MPS to share access to GPU resources requires a CDI-enabled device
+list strategy: `cdi-annotations`. The `envvar` and `volume-mounts` strategies are not supported.
```
cdesiniotis, line 372: Given your changes in https://github.com/NVIDIA/k8s-device-plugin/pull/517 are we planning to remove this note?

elezar: Yes, this can be removed.

README.md
```diff
@@ -348,18 +350,106 @@ options outside of this section are shared.
...
+
+**Note:** The use of time-slicing and MPS are mutually exclusive.
+
+In the case of time-slicing, CUDA time-slicing is used to allow workloads sharing a GPU to
+interleave with each other. However, nothing special is done to isolate workloads that are
 granted replicas from the same underlying GPU, and each workload has access to
 the GPU memory and runs in the same fault-domain as of all the others (meaning
 if one workload crashes, they all do).
 
+In the case of MPS, a control daemon is used to manage access to the shared GPU. This daemon also allows for memory and compute limits to be enforced per workload.
```
klueska, line 367: I would add something in here along the lines of "in contrast to time-slicing, MPS does space partitioning"

README.md
```diff
@@ -348,18 +350,106 @@ options outside of this section are shared.
...
+    renameByDefault: <bool>
+    resources:
+    - name: <resource-name>
+      replicas: <num-replicas>
+    ...
+```
+
+That is, for each named resource under `sharing.mps.resources`, a number
+of replicas can be specified for that resource type. As is the case with
+time-slicing, these replicas represent the number of shared accesses that will
```
klueska, line 387: You say "as is the case with time-slicing". Does this imply that we should have the Time-Slicing section above the MPS one? This is also the order they are presented in the intro...

README.md
```diff
@@ -348,18 +350,106 @@ options outside of this section are shared.
...
+version: v1
+sharing:
+  mps:
+    resources:
+    - name: nvidia.com/gpu
+      replicas: 10
+```
+
+If this configuration were applied to a node with 8 GPUs on it, the plugin
+would now advertise 80 `nvidia.com/gpu` resources to Kubernetes instead of 8.
```
klueska, line 409:
```suggestion
If this configuration were applied to a node with 8 GPUs on it, the plugin
would now advertise 80 `nvidia.com/gpu` resources to Kubernetes instead of 8. Each `nvidia.com/gpu` would provide 1/10 of the total GPU memory and compute resources, respectively.
```

klueska: ignore this comment -- it is addressed below.

README.md
elezar: Done.

README.md
```diff
@@ -486,6 +498,87 @@ nvidia.com/mig-3g.40gb
...
+...
+```
+
+Furthermore, each of these resources -- either `nvidia.com/gpu` or
+`nvidia.com/gpu.shared` -- would have access to the same fraction (1/10) of the
+total memory and compute resources of the GPU.
+
+As of now, the only supported resource available for MPS are
+`nvidia.com/gpu` resources. This implies that it is only compatible when using
+full GPUs or the single MIG strategy.
```
klueska, line 580:
```suggestion
full GPUs or the `single` MIG strategy.
```

README.md
elezar: Done. The timeslicing section is now before MPS.

README.md
elezar: Updating to ```As of now, the only supported resource available for MPS are `nvidia.com/gpu`
resources and only with full GPUs.
``` due to our MIG limitation called out earlier.

PR #508, opened 2024-02-08
bump CUDA base image to 12.3.1

tariq1890:


elezar (approved):
Lgtm

elezar:
Does this need to be backported?

tariq1890:
> Does this need to be backported?

I am not sure. I found the need to bump the image as some of the driver containers were failing image scans. Then again, those failures were only for only centos which has since been discontinued

PR #515, opened 2024-02-12
Document setup with CDI annotations

elezar:
In our deployment, we found that CDI annotations were more stable, and avoided the issue (other workarounds were not effective) described at https://github.com/NVIDIA/gpu-operator/issues/485.

The setup is a bit tricky, and requires configuring an explicit runtime class for the device plugin itself. Document this here for other users.

Migrated from https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/298


ArangoGutierrez:
@mikemckiernan, could you have a peak of these documentation changes? thanks!

manhtukhang:
no stale

ArangoGutierrez (commented):
with the toolkit on 1.18 I think we should now move forward with this PR

rajatchopra (approved):
/lgtm

PR #517, opened 2024-02-13
Remove requirement to use CDI for MPS injection

elezar:
This change allows MPS to be used with the `envvar` and `volumeMounts` device list strategies.

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -337,14 +337,14 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 			return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
 		}
 	}
-	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
-		plugin.updateResponseForMPS(response)
-	}
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyEnvvar) {
 		plugin.updateResponseForDeviceListEnvvar(response, deviceIDs...)
 	}
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyVolumeMounts) {
 		plugin.updateResponseForDeviceMounts(response, deviceIDs...)
+	}
+	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
+		plugin.updateResponseForMPS(response)
 	}
 	if *plugin.config.Flags.Plugin.PassDeviceSpecs {
 		response.Devices = append(response.Devices, plugin.apiDeviceSpecs(*plugin.config.Flags.NvidiaDriverRoot, requestIds)...)
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -326,21 +326,28 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *pluginapi.
 func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*pluginapi.ContainerAllocateResponse, error) {
 	deviceIDs := plugin.deviceIDsFromAnnotatedDeviceIDs(requestIds)
 
-	responseID := uuid.New().String()
-	response, err := plugin.getAllocateResponseForCDI(responseID, deviceIDs)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
+	// Create an empty response that will be updated as required below.
+	response := &pluginapi.ContainerAllocateResponse{
+		Envs: make(map[string]string),
 	}
 
+	if plugin.cdiEnabled {
+		responseID := uuid.New().String()
+		if err := plugin.updateResponseForCDI(response, responseID, deviceIDs...); err != nil {
+			return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
+		}
+	}
+	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
+		plugin.updateResponseForMPS(response)
+	}
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyEnvvar) {
-		response.Envs = plugin.apiEnvs(plugin.deviceListEnvvar, deviceIDs)
+		plugin.updateResponseForDeviceListEnvvar(response, deviceIDs...)
 	}
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyVolumeMounts) {
-		response.Envs = plugin.apiEnvs(plugin.deviceListEnvvar, []string{deviceListAsVolumeMountsContainerPathRoot})
-		response.Mounts = plugin.apiMounts(deviceIDs)
+		plugin.updateResponseForDeviceMounts(response, deviceIDs...)
 	}
 	if *plugin.config.Flags.Plugin.PassDeviceSpecs {
-		response.Devices = plugin.apiDeviceSpecs(*plugin.config.Flags.NvidiaDriverRoot, requestIds)
+		response.Devices = append(response.Devices, plugin.apiDeviceSpecs(*plugin.config.Flags.NvidiaDriverRoot, requestIds)...)
 	}
 	if *plugin.config.Flags.GDSEnabled {
 		response.Envs["NVIDIA_GDS"] = "enabled"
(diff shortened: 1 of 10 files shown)


internal/plugin/server.go
```diff
@@ -331,53 +331,54 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 	if err != nil {
 		return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
 	}
+	plugin.updateContainerAllocatResponseForMPS(&response, deviceIDs)
```
klueska, line 334: What about when there is no MPS? I haven't read further yet, but the name of this function implies to me that MPS is always in use.

internal/plugin/server.go
```diff
@@ -331,53 +331,54 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
...
-				HostPath:      "/var/lib/kubelet/device-plugins/mps/shm",
-			},
-		)
+	return &response, nil
+}
+
+func (plugin NvidiaDevicePlugin) updateContainerAllocatResponseForMPS(response *pluginapi.ContainerAllocateResponse, deviceIDs []string) {
+	if plugin.config.Sharing.SharingStrategy() != spec.SharingStrategyMPS {
+		return
 	}
```
klueska, line 357: I would expect this at the callsite, not embedded in this function.

internal/plugin/server.go
```diff
@@ -523,3 +520,10 @@ func (plugin *NvidiaDevicePlugin) apiDeviceSpecs(driverRoot string, ids []string
 
 	return specs
 }
+
+func ensureEnvvar(response *pluginapi.ContainerAllocateResponse, key string, value string) {
+	if response.Envs == nil {
+		response.Envs = make(map[string]string)
+	}
+	response.Envs[key] = value
+}
```
klueska, line 529: I don't love that this is called all over the place, forcing us to always do the check for `nil`.

When is it _ever_ `nil` by the time it's in this function (meaning we never did this check before any of the assignment calls before).

Is it already initialized from:
```
response, err := plugin.getAllocateResponseForCDI(responseID, deviceIDs)
```

internal/plugin/server.go
elezar: There is a check to see whether it is enabled and return as a no-op if not. I elected to pull this into the function, since we may want to extend the logic to cover more compex cases.

Any suggestions on a function name?

internal/plugin/server.go
elezar: For example if CDI is enabled and GDS as well, there is currently a bug since we don't init the envvars then. We could set Envs to an empty map at the start of the allocate function to remove the need for this.

elezar: To clarify, we are checking whether the `Envs` map is nil. Since we just instantiate a `ContainerAllocateResponse` object in `getAllocateResponseForCDI` this would still be nil since we never set any envvars there.

elezar:
Thanks for the review @klueska.

I have reworked the allocate handler to be more consistent. I now do all the conditional checks in the top-level function and have moved to `update*` functions for all.

internal/plugin/server.go
```diff
@@ -326,69 +326,71 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *pluginapi.
...
 	deviceIDs := plugin.deviceIDsFromAnnotatedDeviceIDs(requestIds)
 
-	responseID := uuid.New().String()
-	response, err := plugin.getAllocateResponseForCDI(responseID, deviceIDs)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
+	// Create an empty response that will be updated as required below.
+	response := &pluginapi.ContainerAllocateResponse{
+		// Envs is set to the empty map to support insertions.
+		Envs: make(map[string]string),
```
klueska, line 332: I don't think the comment on this is necessary

internal/plugin/server.go
```diff
@@ -326,69 +326,71 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *pluginapi.
...
+		}
+		return response, nil
 	}
 
+	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
+		plugin.updateAllocatResponseForMPS(response, deviceIDs...)
+	}
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyEnvvar) {
-		response.Envs = plugin.apiEnvs(plugin.deviceListEnvvar, deviceIDs)
+		plugin.updateDeviceListEnvvar(response, deviceIDs...)
```
klueska, line 347: Should this function be called `plugin.updateAllocateResponseDeviceListEnvar()` for symmetry with the names of the other calls. Or alternatively remove the `AllocateResponse` part of the other calls?

internal/plugin/server.go
```diff
@@ -326,69 +326,71 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *pluginapi.
...
+// updateAllocateResponseForCDI updates the specified response for the given device IDs.
 // This response contains the annotations required to trigger CDI injection in the container engine or nvidia-container-runtime.
-func (plugin *NvidiaDevicePlugin) getAllocateResponseForCDI(responseID string, deviceIDs []string) (pluginapi.ContainerAllocateResponse, error) {
-	response := pluginapi.ContainerAllocateResponse{}
-
-	if !plugin.cdiEnabled {
-		return response, nil
-	}
-
+// TODO: We should generate a CDI specification for MPS
```
klueska, line 392: Does this imply (at the moment) that MPS is **only** supported without CDI enabled?

internal/plugin/server.go
elezar: OK. I will take another look at them.

internal/plugin/server.go
elezar: Removed.

internal/plugin/server.go
elezar: renamed all to `updateResponseFor*` as a balance between length and readability.

internal/plugin/server.go
elezar: I have removed the TODOs related to using CDI for MPS injection.

internal/plugin/server.go
```diff
@@ -326,69 +326,70 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *pluginapi.
...
+	if plugin.cdiEnabled {
+		responseID := uuid.New().String()
+		if err := plugin.updateResponseForCDI(response, responseID, deviceIDs...); err != nil {
+			return nil, fmt.Errorf("failed to get allocate response for CDI: %v", err)
+		}
+		return response, nil
+	}
+	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
+		plugin.updateResponseForMPS(response)
+	}
```
klueska, line 343: Does the order of these need to be reversed? Otherwise we will never inject the MPS stuff when CDI is enabled, right?

elezar: Ah ... sorry, the return shoudl not be in the cdiEnabled block.

klueska: Hmm, now I'm more confused :)

When CDI is enabled we add the CDI devices AND still add the envars, mounts, etc. manually? That doesn't seem right...

I can buy it for the MPS envvars/mounts (since we don't have those in CDI), but I would assume we shouldn't rely on CDI as well as set them all ourselves.

elezar: The issue is that we need to support workloads regardless of whether these use CDI or the NVIDIA container runtime -- put another way. In the case of the GPU operator where we have an `nvidia` and `nvidia-cdi` runtime class, the device plugin must inject the devices regardless of the mode. This is done by allowing *multiple* device-list-strategies to be enabled at the same time.

This means that if a device plugin has `cdi-annotations` set, it will set the contianer annotations. If -- in addition -- `envvars` is specified (or volume mounts for that matter), the envvars will be set.

If the workload is started with the CDI-enabled runtime, the annotations will be read and the `NVIDIA_VISIBLE_DEVICES` envvars ignored. If the workload is started by a non-CDI enabled runtime, the envvars are read as per normal and the annotations have no effect.

klueska: OK, I think I can buy that. If the flags are set, we honor them. This makes sense for the various device list strategies as well as for `--pass-device-specs`. It's a little weird when applied to the `GDSEnabled` and `MOFEDEnabled` since we just say "enable me" but then go about setting the envvar both via CDI and manually (this would apply to the MPS flag in the future as well once we have CDI support for it).

In any case, I think it's fine, even if it seems a bit strange at first glance.

One thing I would change just to make the flow a bit more consistent is to move the following to the end amongst all of the other `config` checks:
```
	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
		plugin.updateResponseForMPS(response)
	}
``` 

and then change the check for `if plugin.cdiEnabled {` to `if plugin.deviceListStrategies.IsCDIEnabled() {` since this really is about checkin of the deviceListStrategy is asking for CDI or not (and it then matches the other calls made below).

elezar: Just to clarify, we don't set those envvars for CDI, we actually inject a `nvidia.com/gds=all` device request (same for mofed).

The `plugin.cdiEnabled` change will require some additional refactoring in this case.

elezar: I have made the change that you suggested and made some refactoring.

I have left it a separate commit so that we can move it to a separate PR if required.

cdesiniotis (approved):

elezar:
@klueska any other comments here?

klueska (approved):
I didn't go through line by line as before, but in general it looks good. It seems to address my comments from before.

PR #518, opened 2024-02-14
Bump nvidia-container-toolkit to v1.15.0-rc.3

cdesiniotis:


elezar:
This has already been handled by Dependabot in #501

PR #520, opened 2024-02-14
Improve e2e tests

elezar:


ArangoGutierrez:
Still WIP?

tests/e2e/device-plugin_test.go
```diff
@@ -63,6 +63,7 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 				fmt.Sprintf("image.repository=%s", *ImageRepo),
 				fmt.Sprintf("image.tag=%s", *ImageTag),
 				fmt.Sprintf("image.pullPolicy=%s", *ImagePullPolicy),
+				"runtimeClassName=nvidia",
```
ArangoGutierrez, line 66: Is there a reason why to use this? (without the test was working)

ArangoGutierrez: I am not against, but if we add this, we need to deploy the `RuntimeClass` CR object, or the test won't run

klueska: It means that the underlying container runtime doesn't have to be configured with nvidia as the default runtime.

tests/e2e/device-plugin_test.go
```diff
@@ -144,6 +145,8 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 			It("it should run GPU jobs", func(ctx context.Context) {
 				By("Creating GPU job")
 				job := common.GPUJob.DeepCopy()
+				runtimeClassName := "nvidia"
+				job.Spec.Template.Spec.RuntimeClassName = &runtimeClassName
```
klueska, line 149: Instead of doing this, you can have a function defined as:
```
func ptr[T any](v T) *T {
  return &v
}
```
And then just do this here:
```
job.Spec.Template.Spec.RuntimeClassName = ptr("nvidia")
```
Probably not worth it if this pattern is only used once (I haven't checked), but worthwhile if used alot.

elezar: This is used elsewhere in our tests. Definitely useful. I will look at making it accessible. 

As a matter of interest, is there guidance on where this kind of functionality should be implemented? (`test-utils`?)

ArangoGutierrez: I think for now makes sense to have on the device-plugin file. or we will use this also for the gfd tests? if so, then utils makes sense

elezar:
@ArangoGutierrez this was merged as part of your e2e testing changes, correct?

ArangoGutierrez:
> @ArangoGutierrez this was merged as part of your e2e testing changes, correct?

Not all of it. only the Makefile changes

PR #529, opened 2024-02-15
Remove k8s.io replace rules

elezar:
This PR removes the k8s.io replace rules and uses explicit dependencies instead.

@klueska @ArangoGutierrez @cdesiniotis do any of you know why this is not best practice?


ArangoGutierrez:
Only if there's a compelling reason for us to ping some dependencies into a specific version, there's no need for using replace

ArangoGutierrez (approved):

klueska:
At one point it was *required* to use the replace rules, otherwise transitive dependencies between different packages would cause issues.

elezar:
> At one point it was _required_ to use the replace rules, otherwise transitive dependencies between different packages would cause issues.

Yes, is there something indicating whether this is still the case?

klueska:
Not sure. A quick google search suggests that some updates were made to how transitive dependencies were handled in go v.1.7 (https://go.dev/ref/mod#graph-pruning). Maybe this makes the explicit replace rules unnecessary now?

elezar:
Could be. For what it's worth, helm doesn't include the replace directives: https://github.com/helm/helm/blob/main/go.mod

elezar:
Also see #530 (which could even be added to this PR)

klueska:
It would be good to understand why it was needed in the past, but if things actually compile without doing the explicit replace anymore, then that is clearly better.

elezar:
I agree. If I find some time, I'll do some more digging. I don't think this is an important change to get in though.

elezar:
Pinged the k8s slack here: https://kubernetes.slack.com/archives/C09R23FHP/p1708089275684879

ArangoGutierrez:
After all the PR's merged today, I think we can circle back to this PR and mark it as ready for review

elezar:
CLosing in favour of #560.

PR #537, opened 2024-02-22
MPS cleanup

elezar:
This PR cleans up a number of aspects around MPS. Namely:

* Removes the MPS log dir from the workload
* Disable MPS for MIG
* Fixes an incorrect conditional check for MPS

### DIFF
--- a/cmd/mps-control-daemon/mps/manager.go
+++ b/cmd/mps-control-daemon/mps/manager.go
@@ -85,6 +85,13 @@ func (m *manager) Daemons() ([]*Daemon, error) {
 			klog.InfoS("Resource is not shared", "resource", "resource", resourceManager.Resource())
 			continue
 		}
+		// Check if MIG devices are included.
+		for _, device := range resourceManager.Devices() {
+			if device.IsMigDevice() {
+				klog.Warning("MPS sharing is not supported for MIG devices; skipping daemon creation")
+				continue
+			}
+		}
 		daemon := NewDaemon(resourceManager, ContainerRoot)
 		daemons = append(daemons, daemon)
 	}
--- a/internal/lm/nvml_test.go
+++ b/internal/lm/nvml_test.go
@@ -80,26 +80,26 @@ func TestMigCapabilityLabeler(t *testing.T) {
 
 func TestSharingLabeler(t *testing.T) {
 	testCases := []struct {
-		descrition     string
+		description    string
 		manager        resource.Manager
 		config         *spec.Config
 		expectedLabels map[string]string
 	}{
 		{
-			descrition: "nil config",
+			description: "nil config",
 			expectedLabels: map[string]string{
 				"nvidia.com/mps.capable": "false",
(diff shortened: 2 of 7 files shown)


cmd/mps-control-daemon/mps/manager.go
```diff
@@ -85,6 +85,13 @@ func (m *manager) Daemons() ([]*Daemon, error) {
 			klog.InfoS("Resource is not shared", "resource", "resource", resourceManager.Resource())
 			continue
 		}
+		// Check if MIG devices are included.
+		for _, device := range resourceManager.Devices() {
+			if device.IsMigDevice() {
+				klog.Warning("MPS sharing is not supported for MIG devices; skipping daemon creation")
+				continue
+			}
+		}
```
klueska, line 94: In the happy path we should never reach this point because of the check on applying the `mps.capable=true` label only if MIG is disabled, but good to have this logic here anyway for the transient case of the label still being there from previous runs (or GFD lagging in its application of an updated label).

elezar: Yes, and a user could in theory force the label too.

klueska (approved):

internal/lm/nvml_test.go
```diff
@@ -132,7 +133,19 @@ func TestSharingLabeler(t *testing.T) {
 			},
 		},
 		{
-			descrition: "config with mps replicas",
+			descrition: "config with mps replicas no-mig-devices",
```
klueska, line 136: This field name should be changed as a followup.
```suggestion
			description: "config with mps replicas no-mig-devices",
```

elezar: Done as a commit on-top here.

internal/plugin/server.go
```diff
@@ -375,20 +382,14 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
...
 	resourceName := plugin.rm.Resource()
 	response.Mounts = append(response.Mounts,
 		&pluginapi.Mount{
 			ContainerPath: plugin.mpsDaemon.PipeDir(),
 			HostPath:      plugin.mpsHostRoot.PipeDir(resourceName),
 		},
-		&pluginapi.Mount{
-			ContainerPath: plugin.mpsDaemon.PipeDir(),
-			HostPath:      plugin.mpsHostRoot.LogDir(resourceName),
-		},
```
klueska, line 391: Looks like this was actually buggy, which is why we probably saw that `log` file I questioned when debugging today.

klueska (approved):

internal/plugin/server.go
elezar: Could be.

elezar: Were we not looking in the pipe directory though?

klueska: I don't remember anymore. In any case, the code looks correct now.

PR #538, opened 2024-02-22
Add actions/stale gh-action

ArangoGutierrez:


ArangoGutierrez:
Based on https://github.com/actions/stale

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,29 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v8
+      with:
+        stale-issue-message: 'This issue has become stale and will be closed automatically within a period of time. Sorry about that.'
+        stale-pr-message: 'This pull request has become stale and will be closed automatically within a period of time. Sorry about that.'
```
klueska, line 19: Can we say the amount of time?

klueska: I would also leave out the "Sorry" part.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,29 @@
...
+        stale-issue-label: 'life-cycle/stale'
+        stale-pr-label: 'life-cycle/stale'
```
klueska, line 21: Do these issues need to preexist?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,29 @@
...
+        days-before-stale: 90
+        close-issue-message: 'This issue was automatically closed due to inactivity.'
+        close-pr-message: 'This pull request was automatically closed due to inactivity.' 
+        close-issue-label: 'life-cycle/rotten'
+        close-pr-label: 'life-cycle/rotten'
```
klueska, line 26: Do these issues need to preexist?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,29 @@
+name: Stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "21 4 * * *"
+
+jobs:
+  stale:
+
```
klueska, line 9: nit: remove newline

.github/workflows/stale.yaml
ArangoGutierrez: done

.github/workflows/stale.yaml
ArangoGutierrez: I created them manually to test the PR

.github/workflows/stale.yaml
ArangoGutierrez: I created them manually to test the PR

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,28 @@
...
+  stale:
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v8
+      with:
+        stale-issue-message: 'This issue has become stale and will be closed automatically within 90 days if no activity is recorded.`
```
elezar, line 17:
```suggestion
        stale-issue-message: 'This issue has become stale and will be closed automatically within 90 days if no activity is recorded.'
```

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,28 @@
...
+        stale-pr-message: 'This pull request has become stale and will be closed automatically within 90 days if no activity is recorded.'
+        stale-issue-label: 'life-cycle/stale'
+        stale-pr-label: 'life-cycle/stale'
+        days-before-stale: 90
+        close-issue-message: 'This issue was automatically closed due to inactivity.'
+        close-pr-message: 'This pull request was automatically closed due to inactivity.' 
+        close-issue-label: 'life-cycle/rotten'
```
elezar, line 24: Is `rotten` commonly used in other projects? It has some negative connotations.

ArangoGutierrez: rotten is used in k8s and k8s-sigs, but we could use another label. abandoned  maybe?

.github/workflows/stale.yaml
ArangoGutierrez: ups thanks

.github/workflows/stale.yaml
klueska: So the answer is 'yes'?

.github/workflows/stale.yaml
elezar: I'm happy to align with k8s sigs.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,28 @@
+name: Stale issues and pull requests
```
cdesiniotis, line 1: Question: What will happen with all open issues / PRs on the first invocation of this job? Will all issues / PRs with >90 days of inactivity get labelled `life-cycle/stale`? Nothing will get closed, right?

ArangoGutierrez: Yup, it will be a busy first day for this action

klueska: Yes to which part of the question? Yes to both? Meaning the stale label will be added to all issues older than 90 days, but NO rotten labels will be added on the initial run, right?

As far, as I understand it, an issue first gets the stale label after 90 days of inactivity. It then needs to sit with the stale label for 90 days before it is relabeled as rotten (pinging all watchers of the issue in the process). It then needs to sit with the rotten label for 90 days before it is automatically closed.

If you remove either label (regardless of which lifecycle state it’s in), the whole process starts over and you have another 270 days before the issue is closed (90 until stale label, another 90 until rotten, and another 90 until closed).

So as long as the initial run only applies the stale label, we have 180 more days until an auto close.

klueska: Actually, here’s the rules applied by k8s, can we just configure ours the same:
https://github.com/kubernetes/website/issues/34183#issuecomment-1242765483

ArangoGutierrez: after seeing the action, looks like issues older than 180 days will get closed

ArangoGutierrez: >So 90,30,30 instead of 90,90,90 as I said above.

This GitHub action only has 2 stages, so it would be 90-60, is that ok? , this PR proposes 90-90

ArangoGutierrez: I have updated the PR
removed the rotten label
and adjusted the closing time

ArangoGutierrez: @klueska @elezar @cdesiniotis

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,26 @@
...
+  stale:
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v8
+      with:
+        stale-issue-message: 'This issue has become stale and will be closed automatically within 90 days if no activity is recorded.'
```
klueska, line 17:
```suggestion
        stale-issue-message: 'This issue has become stale and will be closed automatically within 30 days if no activity is recorded.'
```

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,26 @@
...
+        stale-pr-message: 'This pull request has become stale and will be closed automatically within 90 days if no activity is recorded.'
```
klueska, line 18:
```suggestion
        stale-pr-message: 'This pull request has become stale and will be closed automatically within 30 days if no activity is recorded.'
```

ArangoGutierrez:
I'll merge after #541, not to break any release process

PR #552, opened 2024-02-26
bump helm client to v0.12.8

tariq1890:
This fixes the HIGH CVE : https://github.com/advisories/GHSA-r53h-jv2g-vpx6

NOTE: The latest helm client imports k8s v1.29, so this MR also bumps NFD to v0.15.1 to ensure a consistent dependency tree


ArangoGutierrez (changes requested):
Thanks for this, please create a separate PR for "so this MR also bumps NFD to v0.15.1"

elezar:
@tariq1890 does it make sense to bump NFD first and then let delendabot rebase #543?

Note that since we use replace directives in our go modules, we will have to bump the k8s dependencies separately.

elezar:
@tariq1890 I believe this was updated as part of #556

Note that the NFD bump is done in #485, but probably requires #558 to bump the go version.

ArangoGutierrez:
instead of #558 we ended up doing https://github.com/NVIDIA/k8s-device-plugin/pull/559

tariq1890:
Closing this as this is no longer needed

PR #558, opened 2024-02-27
Use go.mod to get go version

elezar:
This change gets the go version from go.mod. This prevents version skew when building binaries.


ArangoGutierrez:
why the close?

elezar:
Because 1.21 is not a valid version for downloading go. The patch release needs to be specified.

PR #559, opened 2024-02-27
Extract GOLANG_VERSION from versions.mk

elezar:
This PR updates the golang actions to use the GOLANG_VERSION defined in versions.mk

This ensures consistency between what is built for the images and what is tested in CI. 

It also adds a `vendor` and `check-vendor` make target to ensure that the local go module files are up to date.

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -45,6 +45,8 @@ jobs:
         version: latest
         args: -v --timeout 5m
         skip-cache: true
+    - name: Check golang modules
+      run: make check-vendor
   test:
       name: Unit test
       runs-on: ubuntu-latest
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -30,10 +30,15 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       name: Checkout code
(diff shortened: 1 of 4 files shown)


ArangoGutierrez (approved):
First this and then the NFD one right?

elezar:
> First this and then the NFD one right?

NFD includes these changes, so that's probably better.

.github/workflows/golang.yaml
```diff
@@ -30,26 +30,38 @@ jobs:
...
+    - name: Get Golang version
+      id: vars
+      run: |
+        GOLANG_VERSION=$( grep "GOLANG_VERSION ?=" versions.mk )
+        echo "GOLANG_VERSION=${GOLANG_VERSION##GOLANG_VERSION ?= }" >> $GITHUB_ENV
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version-file: 'go.mod'
+        go-version: ${{ env.GOLANG_VERSION }}
```
elezar, line 41: @ArangoGutierrez we should probably look at how we can reuse something like this.

ArangoGutierrez: for all other projects, yup

PR #560, opened 2024-02-27
clean up replace directives in go.mod

tariq1890:
The replace directives aren't needed. This change does not alter the `go.sum` checksums, so the dependency tree is maintained as-is

This improves readability of the go.mod and makes it easier for others to discern the dependency tree of k8s-device-plugin

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,34 +4,6 @@ go 1.21.3
 
 toolchain go1.21.5
 
-replace (
-	k8s.io/api => k8s.io/api v0.29.2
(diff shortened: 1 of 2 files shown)


ArangoGutierrez:
https://github.com/NVIDIA/k8s-device-plugin/pull/529

ArangoGutierrez (changes requested):
Duplicate of https://github.com/NVIDIA/k8s-device-plugin/pull/529

elezar (approved):
I'm happy to merge this and revisit one we have some more experience with these.

It seems as if there is precedent to not include them -- e.g. helm -- and this would definitely improve our dependency experience.

PR #567, opened 2024-02-29
Use clientsets and downward API to construct NFD client

elezar:
Closes https://github.com/NVIDIA/cloud-native-team/issues/46


internal/lm/output.go
```diff
@@ -0,0 +1,192 @@
...
+
+const nodeFeatureVendorPrefix = "nvidia-features-for"
+
+type nodeFeatureObject struct {
+	nodeConfig   flags.NodeConfig
+	nfdClientset nfdclientset.Interface
+}
+
+// UpdateNodeFeatureObject creates/updates the node-specific NodeFeature custom resource.
+func (n *nodeFeatureObject) Output(labels Labels) error {
```
ArangoGutierrez, line 154: I'll leave the func name `UpdateNodeFeatureObject` as it was before.

internal/flags/kubeclient.go
```diff
@@ -0,0 +1,114 @@
+/*
+ * Copyright 2023 The Kubernetes Authors.
+ * Copyright 2023 NVIDIA CORPORATION.
```
ArangoGutierrez, line 3: 24?

internal/flags/node.go
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2023 The Kubernetes Authors.
+ * Copyright 2023 NVIDIA CORPORATION.
```
ArangoGutierrez, line 3: 24?

ArangoGutierrez (changes requested):
looks good, just a few questions

internal/lm/output.go
elezar: This needs to be an `Outputer`. The point is that the caller doesn't care whether labels are output to a file or the node feature API.

internal/flags/kubeclient.go
elezar: Was copied from dra. Will update.

PR #568, opened 2024-02-29
Refactor label output

elezar:


elezar:
Closing since I've added this commit to #567

PR #572, opened 2024-03-04
Set mps device memory limit by index

elezar:
This change switches to setting MPS memory limits by index instead of UUID.

The `nvidia-cuda-mps-control` daemon does not seem to handle the case where the NVML UUID is used to set these limits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Bump CUDA base image version to 12.3.2
 - Add `cdi-cri` device list strategy. This uses the CDIDevices CRI field to request CDI devices instead of annotations.
+- Set MPS memory limit by device index and not device UUID. This is a workaround for an issue where
+  these limits are not applied for devices if set by UUID.
 
 ### Version v0.15.0-rc.1
 - Import GPU Feature Discovery into the GPU Device Plugin repo. This means that
--- a/cmd/mps-control-daemon/mps/daemon.go
+++ b/cmd/mps-control-daemon/mps/daemon.go
@@ -105,10 +105,10 @@ func (d *Daemon) Start() error {
 		return err
 	}
 
-	for uuid, limit := range d.perDevicePinnedDeviceMemoryLimits() {
-		_, err := d.EchoPipeToControl(fmt.Sprintf("set_default_device_pinned_mem_limit %s %s", uuid, limit))
+	for index, limit := range d.perDevicePinnedDeviceMemoryLimits() {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## Changelog
 
 - Bump CUDA base image version to 12.3.2
+- Set MPS memory limit by device index and not device UUID. This is a workaround for an issue where
+  these limits are not applied for devices if set by UUID.
```
klueska, line 5: I know we have MIG support for MPS disabled at the moment, but what implications is this going to have once we want to enable it later?

elezar: Since we have to start an MPS control daemon per MIG device and we would explicitly set `CUDA_VISIBLE_DEVICES` in this case, we would always be referring to `index == 0` here. (or `index == 0:0` assuming that the daemon supports "nested" indices).

klueska: It assume it does not support nested indices. That is something I came up with for NVIDIA_VISIBLE_DEVICES, and I don't think anything else adopted it.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -216,18 +216,18 @@ func (m *Daemon) perDevicePinnedDeviceMemoryLimits() map[string]string {
 	totalMemoryInBytesPerDevice := make(map[string]uint64)
 	replicasPerDevice := make(map[string]uint64)
 	for _, device := range m.Devices() {
-		uuid := device.GetUUID()
-		totalMemoryInBytesPerDevice[uuid] = device.TotalMemory
-		replicasPerDevice[uuid] += 1
+		index := device.Index
```
cdesiniotis, line 219: I believe this is related to the discussion above, but what will the indices look like when we are iterating over MIG devices?

CHANGELOG.md
cdesiniotis: If the nested indices is not a valid convention, does this mean we need to deploy one container per MIG device -- where each container launches an MPS control daemon with `CUDA_VISIBLE_DEVICES=0`?

cdesiniotis (approved):
Changes look fine to me. I am still a bit uncertain how this will all work with MIG, but I do not want to block this PR.

CHANGELOG.md
klueska: Unfortunately, yes. See https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#cuda-mps

elezar:
I have created #587 which switches to using per-container envvars to set MPS limits instead of setting them in the daemon a-priory. This could be considered an alternative to this PR.

klueska (approved):

PR #573, opened 2024-03-04
Add cdi-cri device list strategy

elezar:
Add `cdi-cri` device list strategy that makes use of the `CDIDevices` field in the device plugin API.

This field is internally mapped to annotation AND CRI field modifications in the kubelet.

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -422,6 +422,14 @@ func (plugin *NvidiaDevicePlugin) updateResponseForCDI(response *pluginapi.Conta
 		}
 		response.Annotations = annotations
 	}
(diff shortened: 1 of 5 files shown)


internal/plugin/server.go
```diff
@@ -422,6 +422,14 @@ func (plugin *NvidiaDevicePlugin) updateResponseForCDI(response *pluginapi.Conta
 		}
 		response.Annotations = annotations
 	}
+	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyCDICRI) {
```
elezar, line 425: One note: It is probably not recommended to set *both* `cdi-annotations` and `cdi-cri` since the latter also sets annotations if enabled in the kubelet.

cdesiniotis (approved):
Nice!

PR #574, opened 2024-03-06
Update sharing checks for MPS

elezar:
These changes refactor the sharing checks when processing a Allocate request.

The behaviour that allows `failRequestGreaterThanOne` to control whether multiple devices can be allocated is maintained for TimeSlicing, but this setting is ignored for MPS sharing.

This aligns with the behaviour for the GKE plugin with the exception that we do not "merge" partitions if multiples are requested in a single device. This can be added as a future extension (see https://github.com/NVIDIA/k8s-device-plugin/pull/586).

Closes https://github.com/NVIDIA/cloud-native-team/issues/48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Add `cdi-cri` device list strategy. This uses the CDIDevices CRI field to request CDI devices instead of annotations.
 - Set MPS memory limit by device index and not device UUID. This is a workaround for an issue where
   these limits are not applied for devices if set by UUID.
+- Update MPS sharing to disallow requests for multiple devices if MPS sharing is configured.
 
 ### Version v0.15.0-rc.1
 - Import GPU Feature Discovery into the GPU Device Plugin repo. This means that
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -53,6 +53,14 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
 
 	config.Flags.UpdateFromCLIFlags(c, flags)
 
+	// We explicitly set sharing.mps.failRequestsGreaterThanOne = true
+	// This can be relaxed in certain cases -- such as a single GPU -- but
+	// requires additional logic around when it's OK to combine requests and
+	// makes the semantics of a request unclear.
+	if config.Sharing.MPS != nil {
+		config.Sharing.MPS.FailRequestsGreaterThanOne = true
+	}
+
 	return config, nil
 }
 
--- a/internal/rm/rm.go
+++ b/internal/rm/rm.go
@@ -115,19 +115,36 @@ func (r *resourceManager) Devices() Devices {
 var errInvalidRequest = errors.New("invalid request")
 
 // ValidateRequests checks the requested IDs against the resource manager configuration.
-// This asserts that all requested IDs are known to the resource manager and that the request is
+// It asserts that all requested IDs are known to the resource manager and that the request is
 // valid for a specified sharing configuration.
-// specified sharing configuration is respected.
 func (r *resourceManager) ValidateRequest(ids AnnotatedIDs) error {
+	// Assert that all requested IDs are known to the resource manager
 	for _, id := range ids {
 		if !r.devices.Contains(id) {
 			return fmt.Errorf("%w: unknown device: %s", errInvalidRequest, id)
 		}
 	}
+
 	// If the devices being allocated are replicas, then (conditionally)
 	// error out if more than one resource is being allocated.
-	if len(ids) > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne && ids.AnyHasAnnotations() {
-		return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, len(ids))
+	includesReplicas := ids.AnyHasAnnotations()
+	numRequestedDevices := len(ids)
+	switch r.config.Sharing.SharingStrategy() {
+	case spec.SharingStrategyTimeSlicing:
+		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
+			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
+		}
+	case spec.SharingStrategyMPS:
+		// For MPS sharing, we explicitly ignore the FailRequestsGreaterThanOne
+		// value in the sharing settings.
+		// This setting was added to timeslicing after the initial release and
+		// is set to `false` to maintain backward compatibility with existing
+		// deployments. If we do extend MPS to allow multiple devices to be
+		// requested, the MPS API will be extended separately from the
+		// time-slicing API.
+		if includesReplicas && numRequestedDevices > 1 {
+			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
+		}
 	}
 	return nil
 }
--- a/internal/rm/rm.go
+++ b/internal/rm/rm.go
@@ -17,6 +17,7 @@
 package rm
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -42,6 +43,7 @@ type ResourceManager interface {
 	GetDevicePaths([]string) []string
 	GetPreferredAllocation(available, required []string, size int) ([]string, error)
 	CheckHealth(stop <-chan interface{}, unhealthy chan<- *Device) error
+	ValidateRequest(AnnotatedIDs) error
 }
 
 // NewResourceManagers returns a []ResourceManager, one for each resource in 'config'.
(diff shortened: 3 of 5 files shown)


api/config/v1/sharing_test.go
```diff
@@ -0,0 +1,197 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			input: `{
+				"mps": {
+					"failRequestsGreaterThanOne": false,
```
cdesiniotis, line 90: Question -- what does it effectively mean if `failRequestsGreaterThanOne=false` and in my pod spec I request multiple `nvidia.com/gpu` resources for my container (both for MPS and time-slicing)? For MPS specifically, does requesting multiple `nvidia.com/gpu` resources increase the amount of compute / memory allocated to the container? Or does it just limit the amount of countable resources available for other containers?

cdesiniotis (approved):
LGTM. Left a question for my general understanding.

api/config/v1/sharing_test.go
klueska: I don't even think we should have this option for MPS -- or if we do, it should give you multiple slices, instead of 1 (which we can't do currently because we limit the size of each slice at the control daemon level).

elezar: The issue is that we don't have control over which devices the kubelet is allocating and cannot guarantee that the devices / slices are on the same device. We can relax this restriction IF there is only a single device this is what the GKE plugin does here: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/cf2cc6e7a7d0ed2c2d2c2373fcc7fad7233b8be6/pkg/gpu/nvidia/gpusharing/gpusharing.go#L33-L50

Note that they do combine multiple MPS slices since they set the MPS envvars for the container based on the number of requested devices on a single device:  https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/cf2cc6e7a7d0ed2c2d2c2373fcc7fad7233b8be6/pkg/gpu/nvidia/manager.go#L312-L325

The intent of this change was only to ensure that this option is set to `true` by default for MPS, but I'm also happy to not allow this to be changed.

elezar:
@cdesiniotis @klueska I have changed the implementation here to not rely on `failRequestsGreaterThanOne` for MPS sharing. This option was added after the initial timeslicing work and is only set to default to maintain backward compatibility. I would rather -- once it is required -- add a new option (e.g. `allowRequestsGreaterThanOne`) that would allow users to opt-in to allowing this kind of sharing.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Bump CUDA base image version to 12.3.2
 - Add `cdi-cri` device list strategy. This uses the CDIDevices CRI field to request CDI devices instead of annotations.
+- Update MPS sharing to disalow requests for multiple devices if MPS sharing is configured.
```
tariq1890, line 5:
```suggestion
- Update MPS sharing to disallow requests for multiple devices if MPS sharing is configured.
```

internal/rm/rm.go
```diff
@@ -110,6 +112,35 @@ func (r *resourceManager) Devices() Devices {
...
+		}
+	}
+	// If the devices being allocated are replicas, then (conditionally)
+	// error out if more than one resource is being allocated.
+	includesReplicas := ids.AnyHasAnnotations()
+	numRequestedDevices := len(ids)
+	switch r.config.Sharing.SharingStrategy() {
+	case spec.SharingStrategyTimeSlicing:
+		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
+			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
```
tariq1890, line 134: I don't see the point of using a wrapped error here, as it is just a string constant. We could consider using a custom struct  that implements the `Error` interface.

elezar: The main reason for wrapping the error here is to allow testing for the singpost errors in the unit tests. I suppose I could update the tests to check the error contents.

If you have suggestions on how to make this robust, I'd be happy to update this (possibly as a follow up).

cdesiniotis (approved):

api/config/v1/config.go
```diff
@@ -53,6 +53,13 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
...
 	config.Flags.UpdateFromCLIFlags(c, flags)
 
+	// We explicitly set sharing.mps.failRequestsGreaterThanOne = true
+	// This can be relaxed in certain cases -- such as a single GPU -- but
+	// requires additional logic around when it's OK to combine requests and
+	// makes the semantics of a request unclear.
+	if config.Sharing.MPS != nil {
+		config.Sharing.MPS.FailRequestsGreaterThanOne = true
+	}
 	return config, nil
```
klueska, line 63:
```suggestion
	}
	
	return config, nil
```

internal/rm/rm.go
```diff
@@ -110,6 +112,35 @@ func (r *resourceManager) Devices() Devices {
 	return r.devices
 }
 
+var errInvalidRequest = errors.New("invalid request")
+
+// ValidateRequests checks the requested IDs against the resource manager configuration.
+// This asserts that all requested IDs are known to the resource manager and that the request is
+// valid for a specified sharing configuration.
+// specified sharing configuration is respected.
```
klueska, line 120:
```suggestion
// ValidateRequests checks the requested IDs against the resource manager configuration.
// It asserts that all requested IDs are known to the resource manager and that the request is
// valid for a specified sharing configuration.
```

internal/rm/rm.go
```diff
@@ -110,6 +112,35 @@ func (r *resourceManager) Devices() Devices {
...
+func (r *resourceManager) ValidateRequest(ids AnnotatedIDs) error {
+	for _, id := range ids {
+		if !r.devices.Contains(id) {
+			return fmt.Errorf("%w: unknown device: %s", errInvalidRequest, id)
+		}
+	}
```
klueska, line 126:
```suggestion
	// Assert that all requested IDs are known to the resource manager
	for _, id := range ids {
		if !r.devices.Contains(id) {
			return fmt.Errorf("%w: unknown device: %s", errInvalidRequest, id)
		}
	}
	
```

internal/rm/rm.go
```diff
@@ -110,6 +112,35 @@ func (r *resourceManager) Devices() Devices {
...
+	switch r.config.Sharing.SharingStrategy() {
+	case spec.SharingStrategyTimeSlicing:
+		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
+			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
+		}
+	case spec.SharingStrategyMPS:
+		if includesReplicas && numRequestedDevices > 1 {
+			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
+		}
+	}
```
klueska, line 140: Do we actually need this switch anymore since you now do this early on?:
```
	if config.Sharing.MPS != nil {
		config.Sharing.MPS.FailRequestsGreaterThanOne = true
	}
```

elezar: It's not strictly necessary, but this switch makes it simpler to add extensions where we allow more flexibility for MPS. It also makes it clear that that the behaviour for MPS *may* be different to timeslicing.

It also ensures that modifications to the config (after we set the value as you point out above) doesn't change the behaviour.

I'm willing to change it if we think that we'd rather not have this now.

klueska: OK; since they are identical for now, can we at least change it to:
```
switch r.config.Sharing.SharingStrategy() {
	case spec.SharingStrategyTimeSlicing:
		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
	case spec.SharingStrategyMPS:
		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
}
```
or
```
switch r.config.Sharing.SharingStrategy() {
	case spec.SharingStrategyTimeSlicing:
	case spec.SharingStrategyMPS:
		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
}
```
Meaning, keep the switch, but keep the logic itself the same.

elezar: I'm willing to make the change, but would like to make an additional case against it. I thought we agreed that the `failRequestsGreaterThanOne` and the semantics around it is a workaround for what should have been the correct default behaviour for timeSlicing. I would rather *not* carry this forward into the MPS implementation. Ideally I would have introduced a different type for the MPS `ReplicatedResources`, but deemed that reusing the existing defintion would be simpler in terms of the affect on other labelling code.

Note that we are setting the value to `true` for cosmetic reason -- so that the user doesn't see `failRequestsGreaterThanOne: false` when the config is printed to the logs -- and should not affect the behaviour. Once we extend the MPS API to be richer (we talked about supporting something like mixed mig profiles) we would replace the data structure and no longer need to deal with this flag directly.

klueska: Honestly, I don't feel too strongly either way. Maybe having a comment to the effect of what you wrote above would help. Otherwise, it appears to the casual reader of the code as if we made a mistake, and overlooked the fact that we *should* be doing the exact same operation in both places.

internal/rm/rm.go
tariq1890: We could create a struct that implements the Error interface. The struct could have the relevant fields populated during the time of instantiating the error. The implementing `Error` method could print out the error message based on the struct fields

PR #575, opened 2024-03-06
Add log tailer for MPS control logs

elezar:
This change adds a background `tail -n +1 -f` command for the `control.log` associated with an MPS daemon.

This is outputed to stdout to ensure that MPS control daemon logs are also available in the `kubectl logs` output.

```
I0313 19:23:55.081285      69 daemon.go:131] "Starting log tailer" resource="nvidia.com/gpu"
[2024-03-13 19:23:55.063 Control    98] Starting control daemon using socket /mps/nvidia.com/gpu/pipe/control
[2024-03-13 19:23:55.063 Control    98] To connect CUDA applications to this daemon, set env CUDA_MPS_PIPE_DIRECTORY=/mps/nvidia.com/gpu/pipe
...
[2024-03-13 19:24:23.903 Control    98] NEW UI
[2024-03-13 19:24:23.903 Control    98] Cmd:get_default_active_thread_percentage
[2024-03-13 19:24:23.903 Control    98] 10.0
[2024-03-13 19:24:23.903 Control    98] UI closed
I0313 19:24:40.850399      69 main.go:144] Received signal "terminated", shutting down.
I0313 19:24:40.850462      69 main.go:214] Stopping MPS daemons.
[2024-03-13 19:24:40.870 Control    98] Accepting connection...
[2024-03-13 19:24:40.870 Control    98] NEW UI
[2024-03-13 19:24:40.871 Control    98] Cmd:quit
[2024-03-13 19:24:40.871 Control    98] Exit with status 0
I0313 19:24:40.871316      69 daemon.go:145] "Stopped MPS control daemon" resource="nvidia.com/gpu"
I0313 19:24:40.886798      69 daemon.go:148] "Stopped log tailer" resource="nvidia.com/gpu" error="signal: killed"
```

### DIFF
--- a/cmd/mps-control-daemon/mps/daemon.go
+++ b/cmd/mps-control-daemon/mps/daemon.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"k8s.io/klog/v2"
 
@@ -46,6 +47,8 @@ type Daemon struct {
 	// root represents the root at which the files and folders controlled by the
 	// daemon are created. These include the log and pipe directories.
 	root Root
+	// logTailer tails the MPS control daemon logs.
+	logTailer *tailer
 }
 
 // NewDaemon creates an MPS daemon instance.
@@ -124,23 +127,31 @@ func (d *Daemon) Start() error {
 	}
 	defer statusFile.Close()
 
+	d.logTailer = newTailer(filepath.Join(logDir, "control.log"))
+	klog.InfoS("Starting log tailer", "resource", d.rm.Resource())
+	if err := d.logTailer.Start(); err != nil {
+		klog.ErrorS(err, "Could not start tail command on control.log; ignoring logs")
+	}
+
 	return nil
 }
 
 // Stop ensures that the MPS daemon is quit.
 func (d *Daemon) Stop() error {
-	output, err := d.EchoPipeToControl("quit")
+	_, err := d.EchoPipeToControl("quit")
 	if err != nil {
 		return fmt.Errorf("error sending quit message: %w", err)
 	}
-	klog.InfoS("Shut down MPS", "output", output)
+	klog.InfoS("Stopped MPS control daemon", "resource", d.rm.Resource())
+
+	err = d.logTailer.Stop()
+	klog.InfoS("Stopped log tailer", "resource", d.rm.Resource(), "error", err)
 
 	if err := d.setComputeMode(computeModeDefault); err != nil {
 		return fmt.Errorf("error setting compute mode %v: %w", computeModeDefault, err)
 	}
 
-	err = os.Remove(d.startedFile())
-	if err != nil && err != os.ErrNotExist {
+	if err := os.Remove(d.startedFile()); err != nil && err != os.ErrNotExist {
 		return fmt.Errorf("failed to remove started file: %w", err)
 	}
 
--- a/cmd/mps-control-daemon/mps/log-tailer.go
+++ b/cmd/mps-control-daemon/mps/log-tailer.go
@@ -1,0 +1,69 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package mps
+
+import (
+	"context"
+	"os"
+	"os/exec"
+)
+
+// tailer tails the contents of a file.
+type tailer struct {
+	filename string
+	cmd      *exec.Cmd
+	cancel   context.CancelFunc
+}
+
+// newTailer creates a tailer.
+func newTailer(filename string) *tailer {
+	return &tailer{
+		filename: filename,
+	}
+}
+
+// Start starts tailing the specified filename.
+func (t *tailer) Start() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.cancel = cancel
+
+	//nolint:gosec // G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)
+	cmd := exec.CommandContext(ctx, "tail", "-n", "+1", "-f", t.filename)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	t.cmd = cmd
+	return nil
+}
+
+// Stop stops the tailer.
+// The associated cancel function is called after which the command wait is
+// called -- if applicable.
+func (t *tailer) Stop() error {
+	if t.cancel != nil {
+		t.cancel()
+	}
+
+	if t.cmd == nil {
+		return nil
+	}
+
+	return t.cmd.Wait()
+}


cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
...
-	klog.InfoS("Shut down MPS", "output", output)
+	klog.InfoS("Stopped MPS control daemon", "resource", d.rm.Resource())
+
+	if d.tail != nil {
+		klog.InfoS("Stopping log tailer", "resource", d.rm.Resource())
+		if err := d.tail.Process.Signal(os.Kill); err != nil {
+			klog.ErrorS(err, "Failed to stop log tailer")
+		}
+
+		err := d.tail.Wait()
```
cdesiniotis, line 157: Question -- what happens if an error is returned when sending the kill signal to the `tail` process? Should we skip this `Wait()` call in that case? I would expect the tail process to eventually get reaped anyways once this container exits.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
 	}
 	defer statusFile.Close()
 
+	klog.InfoS("Starting log tailer", "resource", d.rm.Resource())
+	tail := exec.Command("tail", "-n", "+1", "-f", filepath.Join(logDir, "control.log")) //nolint:gosec
+	tail.Stdout = os.Stdout
```
klueska, line 132: do we also want `stderr`?

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
 	}
 	defer statusFile.Close()
 
+	klog.InfoS("Starting log tailer", "resource", d.rm.Resource())
+	tail := exec.Command("tail", "-n", "+1", "-f", filepath.Join(logDir, "control.log")) //nolint:gosec
```
klueska, line 131: What is the purpose of `//nolint:gosec`? Meaning, what linting error would be reported here.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
...
 
 // Stop ensures that the MPS daemon is quit.
 func (d *Daemon) Stop() error {
-	output, err := d.EchoPipeToControl("quit")
+	_, err := d.EchoPipeToControl("quit")
 	if err != nil {
 		return fmt.Errorf("error sending quit message: %w", err)
 	}
-	klog.InfoS("Shut down MPS", "output", output)
+	klog.InfoS("Stopped MPS control daemon", "resource", d.rm.Resource())
```
klueska, line 149: Do these klog outputs get mixed with the tail log?

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
...
+
+	if d.tail != nil {
+		klog.InfoS("Stopping log tailer", "resource", d.rm.Resource())
+		if err := d.tail.Process.Signal(os.Kill); err != nil {
+			klog.ErrorS(err, "Failed to stop log tailer")
+		}
+
+		err := d.tail.Wait()
+		klog.InfoS("Stopped log tailer", "resource", d.rm.Resource(), "output", err)
```
klueska, line 158:
```suggestion
		klog.InfoS("Stopped log tailer", "resource", d.rm.Resource(), "error", err)
```

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
...
+	_, err := d.EchoPipeToControl("quit")
 	if err != nil {
 		return fmt.Errorf("error sending quit message: %w", err)
 	}
-	klog.InfoS("Shut down MPS", "output", output)
+	klog.InfoS("Stopped MPS control daemon", "resource", d.rm.Resource())
+
+	if d.tail != nil {
+		klog.InfoS("Stopping log tailer", "resource", d.rm.Resource())
+		if err := d.tail.Process.Signal(os.Kill); err != nil {
```
klueska, line 153: Do we want os.Kill or os.Term?

cmd/mps-control-daemon/mps/daemon.go
klueska: Yes, I was going to ask the same thing.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
...
 
+	klog.InfoS("Starting log tailer", "resource", d.rm.Resource())
+	tail := exec.Command("tail", "-n", "+1", "-f", filepath.Join(logDir, "control.log")) //nolint:gosec
+	tail.Stdout = os.Stdout
+	if err := tail.Start(); err != nil {
+		klog.ErrorS(err, "Could not start tail command on control.log; ignoring logs")
+		d.tail = nil
+	} else {
+		d.tail = tail
+	}
```
klueska, line 138:
```suggestion
	d.tail = exec.Command("tail", "-n", "+1", "-f", filepath.Join(logDir, "control.log")) //nolint:gosec
	d.tail.Stdout = os.Stdout
	d.tail.Stderr = os.Stderr
	
	if err := d.tail.Start(); err != nil {
		klog.ErrorS(err, "Could not start tail command on control.log; ignoring logs")
		d.tail = nil
	}
```

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,25 +127,51 @@ func (d *Daemon) Start() error {
...
-	err = os.Remove(d.startedFile())
-	if err != nil && err != os.ErrNotExist {
+	if err := os.Remove(d.startedFile()); err != nil && err != os.ErrNotExist {
 		return fmt.Errorf("failed to remove started file: %w", err)
 	}
+
+	logDir := d.LogDir()
+	if err := os.RemoveAll(logDir); err != nil {
+		klog.ErrorS(err, "Failed to remove pipe directory", "path", logDir)
+	}
```
klueska, line 173: This seems relevant, but unrelated to this PR.

cmd/mps-control-daemon/mps/daemon.go
elezar: When constructing a command from a go-variable the linter generates:
```
cmd/mps-control-daemon/mps/daemon.go:131:10: G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)
        tail := exec.Command("tail", "-n", "+1", "-f", filepath.Join(logDir, "control.log"))
                ^
```
I could add a `nolint` comment in the preceding line instead and include the full error message in the comment.

cmd/mps-control-daemon/mps/daemon.go
elezar: Yes. klog prints to stderr by default and that's captured by the container logger.

cmd/mps-control-daemon/mps/daemon.go
elezar: `os.Term` is probably better as it allows `tail` to clean things up.

Does it make sense to send `os.Term` and then `os.Kill` if that fails?

cmd/mps-control-daemon/mps/daemon.go
elezar: Moved to #591 

Note that it manifests here since we're tailing the file from the beginning.

cmd/mps-control-daemon/mps/daemon.go
elezar: Yes, that probably makes sense. I have updated the implementation to not call `Wait` in the event of an error. (and also switched to `SIGTERM`).

cmd/mps-control-daemon/mps/daemon.go
elezar: I have updated to include the error that's being ignored.

cmd/mps-control-daemon/mps/daemon.go
klueska: Approved other PR

cmd/mps-control-daemon/mps/daemon.go
klueska: See my comment [here](https://github.com/NVIDIA/k8s-device-plugin/pull/575#discussion_r1523406298) which avoids this entirely.

cmd/mps-control-daemon/mps/daemon.go
elezar: That does sound cleaner. Let me update.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -124,23 +128,38 @@ func (d *Daemon) Start() error {
...
-	output, err := d.EchoPipeToControl("quit")
+	_, err := d.EchoPipeToControl("quit")
 	if err != nil {
 		return fmt.Errorf("error sending quit message: %w", err)
 	}
-	klog.InfoS("Shut down MPS", "output", output)
+	klog.InfoS("Stopped MPS control daemon", "resource", d.rm.Resource())
+
+	klog.InfoS("Stopping log tailer", "resource", d.rm.Resource())
+	d.tailCancel()
```
klueska, line 156: Thinking about this a bit more (after looking at it), we should probably still track `d.tail` and call `d.tail.Wait()` after the cancel(). It is guaranteed to return and we can check / log its error (if there is one).

klueska: Maybe call it `d.tailCmd` though, so as to by symmetric with `d.tailCancel`

klueska (approved):

elezar:
@klueska a minor simplification since the last review. Also ran some tests and updated the description with the output.

klueska (approved):

PR #587, opened 2024-03-07
Set MPS limits using per-container envvars

elezar:
This change switches from setting MPS limits a-priory to using per-container envvars to set MPS limits. This also avoids issues with setting these by UUID and allows for "merging" of slices on the same device.

This aligns with what is done for the GKE device plugin: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/cf2cc6e7a7d0ed2c2d2c2373fcc7fad7233b8be6/pkg/gpu/nvidia/manager.go#L312

### DIFF
--- a/cmd/mps-control-daemon/mps/daemon.go
+++ b/cmd/mps-control-daemon/mps/daemon.go
@@ -105,19 +105,6 @@ func (d *Daemon) Start() error {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


elezar:
This is an alternative to #572

klueska:
Let's close this in favor of https://github.com/NVIDIA/k8s-device-plugin/pull/572

PR #606, opened 2024-03-19
Add syncronization between the MPS control daemon and the device plugin

elezar:
This change adds an init-container to the device plugin to ensure that the MPS daemon with the correct config has been started *before* the device plugin starts.

This init container logs:
```
I0319 13:48:40.749279      28 wait.go:76] Starting OS watcher.
E0319 13:48:40.749531      28 wait.go:87] "MPS is not ready; retrying ..." err="mps manager is not ready: failed to load .ready config: failed to open .ready file: open /mps/.ready: no such file or directory"
I0319 13:49:10.871775      28 wait.go:84] MPS is ready
```

### DIFF
--- a/cmd/mps-control-daemon/main.go
+++ b/cmd/mps-control-daemon/main.go
@@ -110,6 +110,9 @@ func start(c *cli.Context, cfg *Config) error {
 	var restartTimeout <-chan time.Time
 	var daemons []*mps.Daemon
 restart:
+	if err := os.Remove("/mps/.ready"); err != nil {
+		klog.Warningf("Failed to remove .ready file: %v", err)
+	}
 	// If we are restarting, stop daemons from previous run.
 	if started {
 		err := stopDaemons(daemons...)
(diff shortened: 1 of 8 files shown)


cmd/mps-control-daemon/main.go
```diff
@@ -110,6 +112,10 @@ func start(c *cli.Context, cfg *Config) error {
 	var restartTimeout <-chan time.Time
 	var daemons []*mps.Daemon
 restart:
+	readyFile := mps.ReadyFile{}
+	if err := readyFile.Remove(); err != nil {
+		klog.Warningf("Failed to remove .ready file: %v", err)
+	}
```
klueska, line 118: What if MPS is turned off in the config? Will a pevious ready file just be sitting around until MPS get reenabled at some point in the future?

cmd/nvidia-device-plugin/main.go
```diff
@@ -225,6 +226,31 @@ restart:
...
+						klog.ErrorS(err, "failed to check .ready file")
+						goto restart
+					}
+					if !matches {
+						klog.Info("mismatched MPS sharing config; restarting.")
+						goto restart
+					}
+					continue
+				}
+			}
```
klueska, line 253: I haven't read through this in detail, but at first glance it seems a bit convoluted. Can't we rely on the logic already in-place that handles config updates and hook into that somehow?

frittentheke:
@elezar I suppose you don't want this change here to stale out?

elezar:
@frittentheke you're right, I need to revisit this.

cmd/mps-control-daemon/main.go
elezar: Yes, with the current implementation. What options do we have to ensure that this is removed if the MPS daemonset is removed?

cmd/nvidia-device-plugin/main.go
elezar: It may be possible to just rely on the config change restart logic. Let's make the following assumptions:

1. The MPS daemonset owns the lifecycle of the `.ready` file and this file contains the current config used to start the MPS control daemon.
2. If the config changes and both the MPS daemon and the device plugin are restared by their respective config managers, we should be able to fail plugin startup until: 1. A .ready file exists with the expected config, 2. The MPS pipe is ready from the perspective of the device plugin.

This should not require the additional error / watcher events here.

Note that since the `config-manager` sidecars are not always deployed, this seems to imply that there are configuration changes that we may miss in this case. For example, how do we capture the restart of the MPS daemon or some other configuration update that doesn't use the config manager? Do we expect the user to manually restart both pods in this case to ensure that we're in sync?

frittentheke:
Please keep this alive

PR #613, opened 2024-03-25
Allow builds and tests on Darwin

elezar:
This change uses build flags consistently and allows builds and tests on Darwin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,22 +44,26 @@ CLI_VERSION = $(VERSION)
 endif
 CLI_VERSION_PACKAGE = github.com/NVIDIA/k8s-device-plugin/internal/info
 
-GOOS ?= linux
-
 binaries: cmds
 ifneq ($(PREFIX),)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -44,22 +44,26 @@ CLI_VERSION = $(VERSION)
 endif
 CLI_VERSION_PACKAGE = github.com/NVIDIA/k8s-device-plugin/internal/info
 
-GOOS ?= linux
-
 binaries: cmds
 ifneq ($(PREFIX),)
 cmd-%: COMMAND_BUILD_OPTIONS = -o $(PREFIX)/$(*)
 endif
+ifneq ($(shell uname),Darwin)
```
tariq1890, line 51: question: Is this conditional only applicable to Darwin and Linux or does it include other OSes like Windows, FreeBSD etc?

tariq1890 (approved):

PR #614, opened 2024-03-26
Create clientsets only if nodeFeatureAPI is enabled

elezar:


cmd/gpu-feature-discovery/main.go
```diff
@@ -158,10 +158,15 @@ func start(c *cli.Context, cfg *Config) error {
 		manager := resource.NewManager(config)
 		vgpul := vgpu.NewVGPULib(vgpu.NewNvidiaPCILib())
 
-		clientSets, err := cfg.kubeClientConfig.NewClientSets()
-		if err != nil {
-			return fmt.Errorf("failed to create clientsets: %w", err)
+		var clientSets flags.ClientSets
```
ArangoGutierrez, line 161: just a note: NFD will soon (next release or so) make NodeFeature API enabled by default.

ArangoGutierrez (approved):

cmd/gpu-feature-discovery/main.go
elezar: What is the behaviour for a component such as GFD that may not use this API by default? Is the label file still honoured?

PR #616, opened 2024-03-29
Add healthy server for liveness probe and isolate file.

AlanFokCo:


AlanFokCo:
This feature is for liveness probe, and isolate devices by a config file.

elezar:
@AlanFokCo could you provide more context on the reason for the fix?

PR #622, opened 2024-04-05
Add MustGather function to catch logs on E2E failure

ArangoGutierrez:


go.mod
```diff
@@ -21,11 +21,14 @@ require (
 	golang.org/x/mod v0.16.0
 	golang.org/x/net v0.22.0
 	google.golang.org/grpc v1.62.1
+	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.29.3
 	k8s.io/apiextensions-apiserver v0.29.3
 	k8s.io/apimachinery v0.29.3
+	k8s.io/cli-runtime v0.29.2
```
elezar: Why not `v0.29.3`?

go.mod
```diff
@@ -21,11 +21,14 @@ require (
 	golang.org/x/mod v0.16.0
 	golang.org/x/net v0.22.0
 	google.golang.org/grpc v1.62.1
+	gopkg.in/yaml.v2 v2.4.0
```
elezar, line 24: Is this the right YAML package? I know we use another package in other places.

tests/e2e/common/cp.go
elezar, line 1: As a general comment, I was wondering whether we wanted to rather convert `e2e` to a go submodule so that we don't propagate the test dependencies into the top-level go.mod file. Out of scope for this PR though.

tests/e2e/common/cp.go
```diff
@@ -0,0 +1,391 @@
...
+		} else {
+			fmt.Printf("Dropping out copy after %d retries\n", t.retries)
+		}
+	} else {
+		t.bytesRead += uint64(n)
+	}
+	return
+}
+
+func (o *CopyOptions) untarAll(ns, pod string, prefix string, src remotePath, dest localPath, reader io.Reader) error {
```
elezar: Question: why isn't this already done by holodeck?

tests/e2e/common/cp.go
```diff
@@ -0,0 +1,391 @@
...
+type pathSpec interface {
+	String() string
+}
+
+// localPath represents a client-native path, which will differ based
+// on the client OS, its methods will use path/filepath package which
+// is OS dependant
+type localPath struct {
+	file string
+}
```
elezar: Why not just:
```suggestion
type localPath string
```
instead?

tests/e2e/common/cp.go
```diff
@@ -0,0 +1,391 @@
...
+		return false
+	}
+	return relative == "." || relative == stripPathShortcuts(relative)
+}
+
+// remotePath represents always UNIX path, its methods will use path
+// package which is always using `/`
+type remotePath struct {
+	file string
+}
```
elezar:
```suggestion
type remotePath string
```

tests/e2e/common/cp.go
```diff
@@ -0,0 +1,391 @@
...
+
+// strips trailing slash (if any) both unix and windows style
+func stripTrailingSlash(file string) string {
+	if len(file) == 0 {
+		return file
+	}
+	if file != "/" && strings.HasSuffix(string(file[len(file)-1]), "/") {
+		return file[:len(file)-1]
+	}
```
elezar: I don't quite follow this conditional.

Why not just:
```suggestion
	if file != "/" && strings.HasSuffix(file, "/") {
		return strings.TrimSuffix(file, "/")
	}
```

I'm sure this covers the `file == ""` case too.

go.mod
```diff
@@ -21,11 +21,13 @@ require (
...
 	k8s.io/api v0.29.3
 	k8s.io/apiextensions-apiserver v0.29.3
 	k8s.io/apimachinery v0.29.3
 	k8s.io/client-go v0.29.3
 	k8s.io/klog/v2 v2.120.1
+	k8s.io/kubectl v0.29.2
```
elezar, line 30:
```suggestion
	k8s.io/kubectl v0.29.3
```

tests/e2e/common/mustgather.go
```diff
@@ -0,0 +1,289 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/kubectl/pkg/scheme"
+)
+
+func MustGather(artifactDir, component, namespace string) error {
+	// Get the kubeconfig
+	kubeconfig := os.Getenv("KUBECONFIG")
```
elezar, line 38: Is there a way we can use the clientset in the `internal/flags` package instead?

go.mod
elezar: Also a question: Does it make sense to make `tests/e2e`  a go submodule to ensure that we don't have unneeded dependencies in the root go.mod file?

tests/e2e/device-plugin_test.go
```diff
@@ -107,6 +108,11 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 
 		// Cleanup before next test run
 		AfterEach(func(ctx context.Context) {
+			// Gather logs
+			if CurrentSpecReport().Failed() {
+				err := common.MustGather(filepath.Join(*LogArtifactDir, f.UniqueName), "k8s-device-plugin", f.Namespace.Name)
+				Expect(err).NotTo(HaveOccurred())
```
elezar, line 114: Question: Is it common to also have an Expect in the failuer handler?

tests/e2e/gpu-feature-discovery_test.go
```diff
@@ -135,6 +136,11 @@ var _ = NVDescribe("GPU Feature Discovery", func() {
 
 		// Cleanup before next test run
 		AfterEach(func(ctx context.Context) {
+			// Gather logs
+			if CurrentSpecReport().Failed() {
+				err := common.MustGather(filepath.Join(*LogArtifactDir, f.UniqueName), "gpu-feature-discovery", f.Namespace.Name)
```
elezar, line 141: Nit: Can we add this at a global level?

tests/e2e/common/mustgather.go
```diff
@@ -0,0 +1,289 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/kubectl/pkg/scheme"
+)
+
+func MustGather(artifactDir, component, namespace string) error {
```
elezar, line 36: Should the destination (`artifactDir`) be the last argument?

elezar (commented):
Some general comments / questions.

go.mod
ArangoGutierrez: it was `go mod tidy` pick not mine jeje

go.mod
ArangoGutierrez: yeah probably, I am also thinking on moving most of `common` folder to `NVIDIA/k8s-test-infra` so we can share this logic across all repos and avoid code duplication

tests/e2e/common/mustgather.go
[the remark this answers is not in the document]
ArangoGutierrez: is a totally rewritten function with the same name. I want to do it from `NVIDIA/k8s-test-infra`  so it becomes a shared pkg for all repos

go.mod
ArangoGutierrez: you are right, we tend to use `sigs.k8s.io/yaml v1.3.0 // indirect`

tests/e2e/common/mustgather.go
ArangoGutierrez: Since I want to move this func to be imported from `NVIDIA/k8s-test-infra` I ask if we can leave it like this for now

tests/e2e/gpu-feature-discovery_test.go
ArangoGutierrez: Each test is independent so if done at the global level we would only catch the last run. `AfterEach` is the right place so we run the log collection if a specific test fails.

Each test container creates a namespace and deploys helm on it.

tests/e2e/common/mustgather.go
```diff
@@ -0,0 +1,289 @@
...
+		return nil, err
+	}
+
+	return clientset, nil
+}
+
+// setKubernetesDefaults sets default values on the provided client config for accessing the
+// Kubernetes API or returns an error if any of the defaults are impossible or invalid.
+func setKubernetesDefaults(config *rest.Config) error {
+	// TODO remove this hack.  This is allowing the GetOptions to be serialized.
```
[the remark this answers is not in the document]
ArangoGutierrez: Umm this is not needed here, this is just a function needed for `kubectl nv` but not needed here

tests/e2e/common/mustgather.go
ArangoGutierrez: done

tests/e2e/gpu-feature-discovery_test.go
```diff
@@ -135,6 +136,10 @@ var _ = NVDescribe("GPU Feature Discovery", func() {
 
 		// Cleanup before next test run
 		AfterEach(func(ctx context.Context) {
+			// Gather logs
+			if CurrentSpecReport().Failed() {
+				_ = common.MustGather("gpu-feature-discovery", f.Namespace.Name, filepath.Join(*LogArtifactDir, f.UniqueName))
```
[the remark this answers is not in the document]
ArangoGutierrez: done

elezar (approved):
Let's get these in and iterate as required.

PR #629, opened 2024-04-09
[golangci-lint] fix implicit memory aliasing in for loop 

tariq1890:
To get more details on these why fixes were implemented: https://husni.dev/beware-of-implicit-memory-aliasing-in-go-foor-loop/

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  timeout: 10m
 
 linters:
   enable:
@@ -28,11 +28,6 @@ issues:
     linters:
     - gosec
     text: "G404"
-  # We disable the memory aliasing checks in tests
-  - path: ".*_test.go"
-    linters:
-    - gosec
-    text: "G601: Implicit memory aliasing in for loop"
   # We create world-readable files in tests.
   - path: ".*_test.go"
     linters:
--- a/internal/plugin/server_test.go
+++ b/internal/plugin/server_test.go
@@ -131,7 +131,8 @@ func TestCDIAllocateResponse(t *testing.T) {
 		},
 	}
 
-	for _, tc := range testCases {
+	for i := range testCases {
+		tc := testCases[i]
 		t.Run(tc.description, func(t *testing.T) {
 			deviceListStrategies, _ := v1.NewDeviceListStrategies(tc.deviceListStrategies)
 			plugin := NvidiaDevicePlugin{
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -18,10 +18,8 @@ package e2e
 
 import (
 	"flag"
-	"math/rand"
 	"os"
 	"testing"
-	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
@@ -52,7 +50,6 @@ func TestMain(m *testing.M) {
 		e2elog.Failf("Required flags not set. Please set -image.repo, -image.tag and -helm-chart")
 	}
 
-	rand.New(rand.NewSource(time.Now().UnixNano()))
 	os.Exit(m.Run())
 }
 


internal/plugin/server_test.go
```diff
@@ -132,6 +132,7 @@ func TestCDIAllocateResponse(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
```
ArangoGutierrez, line 135:
```suggestion
	for i := range testCases {
	tc := testCases[i]
```

ArangoGutierrez (changes requested):

internal/plugin/server_test.go
tariq1890: Sure, I can make this change. However, the change I've made is a well known and widely followed practice. We have done the same in the other CNT repos

elezar: As a note. This is a false positive if go1.22 is used.

tariq1890: @ArangoGutierrez I've made the change

ArangoGutierrez (approved):

elezar:
Just as a matter of interest, what was the motivation for not allowing this in test files? Was our filter too broad?

tests/e2e/framework/framework.go
```diff
@@ -138,7 +139,11 @@ func (f *Framework) BeforeEach(ctx context.Context) {
 		f.UniqueName = f.Namespace.GetName()
 	} else {
 		// not guaranteed to be unique, but very likely
-		f.UniqueName = fmt.Sprintf("%s-%08x", f.BaseName, rand.Int31())
+		nBig, err := rand.Int(rand.Reader, big.NewInt(31))
+		if err != nil {
+			panic(err)
+		}
+		f.UniqueName = fmt.Sprintf("%s-%08x", f.BaseName, nBig.Int64())
```
elezar, line 146: Question: If we're changing this anyway, why not just use a uuid?

tests/e2e/framework/util.go
```diff
@@ -49,7 +49,11 @@ var RunID = uuid.NewUUID()
 
 // RandomSuffix provides a random sequence to append to pods,services,rcs.
 func RandomSuffix() string {
-	return strconv.Itoa(rand.Intn(10000))
+	nBig, err := rand.Int(rand.Reader, big.NewInt(10000))
```
elezar, line 52: https://github.com/kubernetes/kubernetes/blob/2183a84feb564582489d56c6570917959f003726/staging/src/k8s.io/apimachinery/pkg/util/rand/rand.go#L98

elezar (commented):
Does using existing implementations (like in kubelet) simplify this?

tests/e2e/framework/util.go
tariq1890: That is using `math/rand`, not `crypto/rand`. Although I agree it's a good way of circumventing the golangci-lint check :P

tests/e2e/framework/framework.go
tariq1890: I guess the question si - do we want the UUID format here?

tests/e2e/framework/util.go
elezar: The intent is not to circumvent the checks. The intent is to find a simpler way to generate a unique suffix without having to "roll-our-own" implementation all the time.

I am also not concerned about using weak random number generators in our tests since we do not require them to be cryptographically secure. We just require a lower chance of name collisions, and at the scale we're operating, these are already low.

tests/e2e/framework/framework.go
tariq1890: No longer relevant after addressing your second comment

tests/e2e/framework/util.go
tariq1890: I understand. I have addressed this comment and updated the PR. The diff should look a lot more simpler now

tariq1890:
@elezar  Requesting another round of review. I have addressed your comments and the diff looks a lot simpler now. Thanks!

.golangci.yml
```diff
@@ -20,21 +20,6 @@ linters:
...
-    linters:
-    - gosec
-    text: "G601: Implicit memory aliasing in for loop"
-  # We create world-readable files in tests.
-  - path: ".*_test.go"
-    linters:
-    - gosec
-    text: "G306: Expect WriteFile permissions to be 0600 or less"
\ No newline at end of file
+  gosec:
```
tariq1890, line 23: We simplify the config even further when we use `linters-settings` instead of `exclude-rules` to disable certain lint rules

internal/plugin/server_test.go
```diff
@@ -131,7 +131,8 @@ func TestCDIAllocateResponse(t *testing.T) {
 		},
 	}
 
-	for _, tc := range testCases {
+	for i := range testCases {
```
tariq1890, line 134: @elezar I can limit the changeset to this fix given we don't really need the `math/rand` lint fix

.golangci.yml
```diff
@@ -20,21 +20,7 @@ linters:
...
+    excludes:
+      - G306 # We create world-readable files in tests.
```
elezar, line 25: Does this not mean that we're now ignoring these globally? Why not limit them to tests explicitly?

tariq1890: ah that's a fair point

tariq1890: I guess I'll have to revert this change

tariq1890: Plase check now

elezar (approved):

tests/e2e/e2e_test.go
```diff
@@ -52,7 +50,6 @@ func TestMain(m *testing.M) {
 		e2elog.Failf("Required flags not set. Please set -image.repo, -image.tag and -helm-chart")
 	}
 
-	rand.New(rand.NewSource(time.Now().UnixNano()))
```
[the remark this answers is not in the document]
tariq1890: Yeah, I checked this. Pretty sure we don't need it. The function doesn't seem to set/alter a global state

PR #639, opened 2024-04-11
Update go version to 1.22.2

tariq1890:


elezar (approved):
Note that you can also run:
```
make docker-vendor
```
to ensure that the right go version is used.

PR #644, opened 2024-04-12
use the helm quote function to wrap boolean values in quotes

tariq1890:
It's generally recommended to use the `quote` function instead of explicitly wrapping helm values with quotes in-line

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -78,9 +78,9 @@ spec:
         - name: CONFIG_FILE_DST
           value: "/config/config.yaml"
         - name: DEFAULT_CONFIG
-          value: "{{ .Values.config.default }}"
+          value: {{ .Values.config.default }}
         - name: FALLBACK_STRATEGIES
-          value: "{{ join "," .Values.config.fallbackStrategies }}"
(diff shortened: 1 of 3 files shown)


shivamerla (approved):

klueska:
I'm not necessarily opposed to this, just curious is there there is any magic going on with doing things this way. Do all double quotes within the string being passed to `quote` get escaped, for example?

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -78,9 +78,9 @@ spec:
         - name: CONFIG_FILE_DST
           value: "/config/config.yaml"
         - name: DEFAULT_CONFIG
-          value: "{{ .Values.config.default }}"
+          value: {{ .Values.config.default | quote }}
```
elezar, line 81: Note that since we're dealing with a yaml file, we don't even need the quotes in the first place. Would:
```suggestion
          value: {{ .Values.config.default }}
```
not be a better cleanup?

elezar (commented):
@tariq1890 would removing the quotes from string literals entirely not be a better option than adding the `quote` operator?

tariq1890:
@elezar I think this would be a safer change as it results in no changes to the yaml parser. Strings like `true` and `false` unquoted are interpreted as booleans. In this helm chart, we are choosing to quote them explicitly.

elezar (approved):

PR #649, opened 2024-04-15
E2E test: Ommit error ifNotFound during NFD cleanup

### LINKED ISSUE #642: Fix e2e tests
The e2e tests fail (see https://github.com/NVIDIA/k8s-device-plugin/actions/runs/8660409546/job/23748292556) with the following shown:
```
FAILED] Unexpected error:
      <*errors.StatusError | 0xc0000cc8c0>: 
      nodefeatures.nfd.k8s-sigs.io "holodeck" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "nodefeatures.nfd.k8s-sigs.io \"holodeck\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "holodeck",
                  Group: "nfd.k8s-sigs.io",
                  Kind: "nodefeatures",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
  occurred
  In [AfterEach] at: /home/runner/work/k8s-device-plugin/k8s-device-plugin/tests/e2e/common/kubernetes.go:187 @ 04/12/24 10:15:04.46
```

This happens when trying to delete the nfd namespace, but it is aready found to not exist. I believe the check needs to be updated to include logic such as https://github.com/kubernetes-sigs/node-feature-discovery/blob/624c02e1e2b5a4340b4921ebc04dd751c200f452/pkg/nfd-gc/nfd-gc.go#L79-L90

Something like (although I must admit that I don't know what the equivalent Gomega expressions are):
```
diff --git a/tests/e2e/common/kubernetes.go b/tests/e2e/common/kubernetes.go
index a30c763f..71c634ea 100644
--- a/tests/e2e/common/kubernetes.go
+++ b/tests/e2e/common/kubernetes.go
@@ -174,7 +174,7 @@ func CleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset, name
 		By("Deleting NodeFeatureRule objects from the cluster")
 		for _, nfr := range nfrs.Items {
 			err = cli.NfdV1alpha1().NodeFeatureRules().Delete(ctx, nfr.Name, metav1.DeleteOptions{})
-			Expect(err).Or(NotTo(HaveOccurred()), MatchError(errors.New("not found"))))
+			Expect(err).Or(NotTo(HaveOccurred()), MatchError(errors.New("not found")))
 		}
 	}
 
@@ -185,7 +185,7 @@ func CleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset, name
 		By("Deleting NodeFeature objects from namespace " + namespace)
 		for _, nf := range nfs.Items {
 			err = cli.NfdV1alpha1().NodeFeatures(namespace).Delete(ctx, nf.Name, metav1.DeleteOptions{})
-			Expect(err).Or(NotTo(HaveOccurred()), MatchError(errors.New("not found"))))
+			Expect(err).Or(NotTo(HaveOccurred()), MatchError(errors.New("not found")))
 		}
 	}
 }
```

ArangoGutierrez:
Fixes #642

### DIFF
--- a/tests/e2e/common/kubernetes.go
+++ b/tests/e2e/common/kubernetes.go
@@ -165,29 +165,8 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
 	}
 }
 
-func CleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
-	// Clean NodeFeatureRule objects
-	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
-	if errors.IsNotFound(err) {
-		// Omitted error, nothing to do.
-	} else {
-		Expect(err).NotTo(HaveOccurred())
-
-		if len(nfrs.Items) != 0 {
-			By("Deleting NodeFeatureRule objects from the cluster")
-			for _, nfr := range nfrs.Items {
-				err = cli.NfdV1alpha1().NodeFeatureRules().Delete(ctx, nfr.Name, metav1.DeleteOptions{})
-				if errors.IsNotFound(err) {
-					// Omitted error
-					continue
-				} else {
-					Expect(err).NotTo(HaveOccurred())
-				}
-			}
-		}
-	}
-
-	// Clean NodeFeature objects
+// cleanupNodeFeatures deletes all NodeFeature objects in the given namespace
+func cleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
 	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
 	if errors.IsNotFound(err) {
 		// Omitted error, nothing to do.
@@ -207,4 +186,32 @@ func CleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespac
 			}
 		}
 	}
+}
+
+// cleanupNodeFeatureRules deletes all NodeFeatureRule objects
+func cleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset) {
+	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
+	if errors.IsNotFound(err) {
+		// Omitted error, nothing to do.
+	} else {
+		Expect(err).NotTo(HaveOccurred())
+
+		if len(nfrs.Items) != 0 {
+			By("Deleting NodeFeatureRule objects from the cluster")
+			for _, nfr := range nfrs.Items {
+				err = cli.NfdV1alpha1().NodeFeatureRules().Delete(ctx, nfr.Name, metav1.DeleteOptions{})
+				if errors.IsNotFound(err) {
+					// Omitted error
+					continue
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+		}
+	}
+}
+
+func CleanupNFDObjects(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
+	cleanupNodeFeatureRules(ctx, cli)
+	cleanupNodeFeatures(ctx, cli, namespace)
 }
--- a/tests/e2e/common/kubernetes.go
+++ b/tests/e2e/common/kubernetes.go
@@ -165,25 +165,29 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
 	}
 }
 
+func CleanupNFDObjects(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
+	cleanupNodeFeatureRules(ctx, cli)
+	cleanupNodeFeatures(ctx, cli, namespace)
+}
+
 // cleanupNodeFeatures deletes all NodeFeature objects in the given namespace
 func cleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
 	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
 	if errors.IsNotFound(err) {
 		// Omitted error, nothing to do.
-	} else {
-		Expect(err).NotTo(HaveOccurred())
+		return
+	}
+	Expect(err).NotTo(HaveOccurred())
 
-		if len(nfs.Items) != 0 {
-			By("Deleting NodeFeature objects from namespace " + namespace)
-			for _, nf := range nfs.Items {
-				err = cli.NfdV1alpha1().NodeFeatures(namespace).Delete(ctx, nf.Name, metav1.DeleteOptions{})
-				if errors.IsNotFound(err) {
-					// Omitted error
-					continue
-				} else {
-					Expect(err).NotTo(HaveOccurred())
-				}
+	if len(nfs.Items) != 0 {
+		By("Deleting NodeFeature objects from namespace " + namespace)
+		for _, nf := range nfs.Items {
+			err = cli.NfdV1alpha1().NodeFeatures(namespace).Delete(ctx, nf.Name, metav1.DeleteOptions{})
+			if errors.IsNotFound(err) {
+				// Omitted error
+				continue
 			}
+			Expect(err).NotTo(HaveOccurred())
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,43 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
-			Expect(err).NotTo(HaveOccurred())
+			if err != nil {
+				if errors.IsNotFound(err) {
+					// Ommitted error
+					continue
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+
```
elezar, line 185:
```suggestion
			if errors.IsNotFound(err) {
				// Omitted error
				continue
			} else {
				Expect(err).NotTo(HaveOccurred())
			}
```

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,43 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
 			err = cli.NfdV1alpha1().NodeFeatures(namespace).Delete(ctx, nf.Name, metav1.DeleteOptions{})
-			Expect(err).NotTo(HaveOccurred())
+			if err != nil {
+				if errors.IsNotFound(err) {
+					// Ommitted error
+					continue
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
```
elezar, line 204:
```suggestion
			if errors.IsNotFound(err) {
				// Omitted error
				continue
			} else {
				Expect(err).NotTo(HaveOccurred())
			}
```

elezar (changes requested):
We don't really need the two levels of nesting.

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,38 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
 	}
 }
 
-func CleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
-	// Drop NodeFeatureRule objects
+func CleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
+	// Clean NodeFeatureRule objects
 	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
```
elezar, line 170: Question: Don't we have to do the same thing here? Does `List` return `ErrNotFound` if there are no nodeFeatureRules?

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,38 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
 	Expect(err).NotTo(HaveOccurred())
 
 	if len(nfrs.Items) != 0 {
 		By("Deleting NodeFeatureRule objects from the cluster")
 		for _, nfr := range nfrs.Items {
```
elezar, line 175: Question: Could we be deleting node feature rules that are NOT controlled by our components?

tests/e2e/common/kubernetes.go
ArangoGutierrez: umm, it is possible, yes

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,46 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
 	}
 }
 
-func CleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
-	// Drop NodeFeatureRule objects
+func CleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
+	// Clean NodeFeatureRule objects
 	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
```
elezar, line 170: Given that each of these have multiple levels of nesting, let's move this to a separate `cleanupNodeFeatureRules` function with a quick return if `errors.IsNotFound`?

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,46 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
+					continue
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
 		}
 	}
 
+	// Clean NodeFeature objects
 	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
```
elezar, line 191: This can also be a separate function and then we can quick-return when `errors.IsNotFound` is triggered.

tests/e2e/gpu-feature-discovery_test.go
```diff
@@ -143,9 +143,9 @@ var _ = NVDescribe("GPU Feature Discovery", func() {
 			// Delete Helm release
 			err := helmClient.UninstallReleaseByName(helmReleaseName)
 			Expect(err).NotTo(HaveOccurred())
-			// cleanup node
+			// Cleanup node
 			common.CleanupNode(ctx, f.ClientSet)
-			common.CleanupNodeFeatureRules(ctx, nfdClient, f.Namespace.Name)
+			common.CleanupNodeFeatures(ctx, nfdClient, f.Namespace.Name)
```
elezar, line 148: Question: Since we delete both `NodeFeatures` and `NodeFeatureRules` in this function is there a better name for the collection of these two?

elezar (commented):
I think factoring the two deletion blocks into two separate blocks will make things much clearer and reduce nesting levels.

tests/e2e/gpu-feature-discovery_test.go
ArangoGutierrez: `NFDObjects` ?

ArangoGutierrez:
> I think factoring the two deletion blocks into two separate blocks will make things much clearer and reduce nesting levels.

How about now?

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,53 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
-func CleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
-	// Drop NodeFeatureRule objects
-	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
-	Expect(err).NotTo(HaveOccurred())
+// cleanupNodeFeatures deletes all NodeFeature objects in the given namespace
+func cleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
+	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
+	if errors.IsNotFound(err) {
+		// Omitted error, nothing to do.
+	} else {
```
elezar, line 173:
```suggestion
	if errors.IsNotFound(err) {
		return
	}
```

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,53 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
+}
 
-	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
-	Expect(err).NotTo(HaveOccurred())
+// cleanupNodeFeatureRules deletes all NodeFeatureRule objects
+func cleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset) {
+	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
+	if errors.IsNotFound(err) {
+		// Omitted error, nothing to do.
+	} else {
```
elezar, line 196:
```suggestion
	if errors.IsNotFound(err) {
		return
	}
```

elezar (changes requested):
Last comment. Let's quick-return to reduce nesting.

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,55 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
-	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
-	Expect(err).NotTo(HaveOccurred())
+// cleanupNodeFeatures deletes all NodeFeature objects in the given namespace
+func cleanupNodeFeatures(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
+	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
+	if errors.IsNotFound(err) {
+		// Omitted error, nothing to do.
+		return
+	} else {
+		Expect(err).NotTo(HaveOccurred())
```
elezar, line 175:
```suggestion
	}
	Expect(err).NotTo(HaveOccurred())
```

Since we return we don't need the `else` block.

elezar (changes requested):
Remove the `else` blocks. This is what I mean by removing indentation.

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,53 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
+
+	if len(nfs.Items) != 0 {
+		By("Deleting NodeFeature objects from namespace " + namespace)
+		for _, nf := range nfs.Items {
+			err = cli.NfdV1alpha1().NodeFeatures(namespace).Delete(ctx, nf.Name, metav1.DeleteOptions{})
+			if errors.IsNotFound(err) {
+				// Omitted error
+				continue
+			} else {
+				Expect(err).NotTo(HaveOccurred())
```
elezar, line 185: Nit: technically we can drop this else block too. Not a blocker.

tests/e2e/common/kubernetes.go
```diff
@@ -164,27 +165,53 @@ func CleanupNode(ctx context.Context, cs clientset.Interface) {
...
-	Expect(err).NotTo(HaveOccurred())
-
-	if len(nfs.Items) != 0 {
-		By("Deleting NodeFeature objects from namespace " + namespace)
-		for _, nf := range nfs.Items {
-			err = cli.NfdV1alpha1().NodeFeatures(namespace).Delete(ctx, nf.Name, metav1.DeleteOptions{})
-			Expect(err).NotTo(HaveOccurred())
-		}
-	}
+func CleanupNFDObjects(ctx context.Context, cli *nfdclient.Clientset, namespace string) {
```
elezar, line 214: nit: If we have the public function before the others the diff might align better and the file would be easier to skim from top to bottom.

elezar (approved):
I only have some minor nits left.

Thanks for the patience.

elezar (approved):

PR #650, opened 2024-04-15
E2E Test: Create a DiagnosticCollector

ArangoGutierrez:
The `DiagnosticCollector` is an enhanced version of the previously introduced MustGather function, the result is still the same, but now we have granular control over the objects that we want to capture logs for each test and add the option to manually specify which objects to collect from for when running the e2e manually and the user is only interested in the behavior of a particular object. 

It also defines "extensions" referring to CRD's, opening the door for future CRD's we want to check during testing. this initial iteration only defines NFD for `NodeFeature` and `NodeFeatureRule`.


tests/e2e/device-plugin_test.go
```diff
@@ -71,6 +85,13 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 			},
 		}
 
+		// check Collector objects
+		if *CollectLogs == "default" {
+			collectLogsFrom = strings.Join(defaultCollectorObjects, ",")
+		} else {
+			collectLogsFrom = *CollectLogs
+		}
+
```
elezar, line 94: if we change the parse function to accept a `...string` then we can rather do:
```suggestion
		collectLogsFrom = defaultCollectorObjects
		// check Collector objects
		if *CollectLogs != nil && *CollectLogs != "default" {
			collectLogsFrom = strings.Split(*CollectLogs, ",")
		}
```

tests/e2e/device-plugin_test.go
```diff
@@ -57,8 +58,21 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 			chartSpec       helm.ChartSpec
 			helmReleaseName string
 			kubeconfig      []byte
+
+			collectLogsFrom     string
+			diagnosticCollector *common.DiagnosticCollector
 		)
 
+		defaultCollectorObjects := []string{
```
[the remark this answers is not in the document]
elezar: Should this be a package-level constant?

tests/e2e/device-plugin_test.go
```diff
@@ -104,13 +125,16 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 			Expect(err).NotTo(HaveOccurred())
 			_, err = helmClient.InstallChart(ctx, &chartSpec, nil)
 			Expect(err).NotTo(HaveOccurred())
+			diagnosticCollector, err = common.NewDiagnosticCollector(f.Namespace.Name, filepath.Join(*LogArtifactDir, f.UniqueName))
+			diagnosticCollector.ParseObjects(collectLogsFrom)
```
elezar, line 129: If we always call `ParseObjects` here, does it make sense to use funcational arguments instead and rewrite this as:
```
common.NewDiagnosticCollector(
	common.WithNamespace(f.Namespace.Name), 
	common.WithArtifactDir(filepath.Join(*LogArtifactDir, f.UniqueName)),
	common.WithObjects(collectLogsFrom...),
)
```

tests/e2e/e2e_test.go
```diff
@@ -37,6 +37,7 @@ var (
 	ImageRepo             = flag.String("image.repo", "", "Image repository to fetch image from")
 	ImageTag              = flag.String("image.tag", "", "Image tag to use")
 	ImagePullPolicy       = flag.String("image.pull-policy", "IfNotPresent", "Image pull policy")
+	CollectLogs           = flag.String("collect-logs", "default", "Comma separated list of objects to collect logs from after test")
```
elezar, line 40: nit: collect-logs sounds like a verb. Could we clarify?

tests/e2e/gpu-feature-discovery_test.go
```diff
@@ -63,6 +64,15 @@ var _ = NVDescribe("GPU Feature Discovery", func() {
 		"nvidia.com/mps.capable":          "[true|false]",
 	}
 
+	defaultCollectorObjects := []string{
+		"pods",
+		"nodes",
+		"namespaces",
+		"deployments",
+		"demonsets",
+		"nodefeatures",
```
elezar, line 73: Is the reason for a different default just because we don't launch a job here? Is there harm in also defining jobs here?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+		d.Add(d.NodeFeatureRules)
+	}
+}
+
+// ParseObjects reads from a comma separated string and sets the DiagnosticCollector objects
+// This is a helper function to set the DiagnosticCollector objects from a comma separated string
+// of objects, for making it easier to set the objects to collect when manually running the
+// diagnostic collector.
+func (d *DiagnosticCollector) ParseObjects(objects string) {
+	for _, obj := range strings.Split(objects, ",") {
```
elezar, line 139: Is there a reason that we don't just call `d.Add()` here? What is the benefit of storing a set of strings and THEN converting that to collectors?

elezar: Also, do we need to deduplicate the objects?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+	case NodeFeatureRule:
+		d.Add(d.NodeFeatureRules)
+	}
+}
+
+// ParseObjects reads from a comma separated string and sets the DiagnosticCollector objects
+// This is a helper function to set the DiagnosticCollector objects from a comma separated string
+// of objects, for making it easier to set the objects to collect when manually running the
+// diagnostic collector.
+func (d *DiagnosticCollector) ParseObjects(objects string) {
```
elezar, line 138: How about explicitly accepting `objects ...string`?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+		d.Add(d.Jobs)
+	}
+
+	// Collect the extensions
+	for i := range d.Extensions {
+		d.AddExtension(d.Extensions[i])
+	}
+
+	if len(d.collectors) == 0 {
+		return fmt.Errorf("no collectors specified")
```
elezar, line 198: Is this really an error state? Why not just issue a warning?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+		}
+	}
+	return nil
+}
+
+// Nodes is a collectorFunc that collects node objects
+func (d *DiagnosticCollector) Nodes(ctx context.Context) error {
+	nodes, err := d.clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		if _, lerr := d.errLogFile.WriteString(fmt.Sprintf("Error getting nodes: %v\n", err)); lerr != nil {
```
elezar, line 214: Why not call `fmt.Fprintf` directly? `d.errLogFile` should be an `io.Writer`.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+			err = fmt.Errorf("%v+ error writing to stderr log file: %v", err, lerr)
```
elezar, line 215: should we wrap the errors? Or just call `errors.Join` here?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+		}
+		return err
+	}
+
+	nodesFile, err := os.Create(filepath.Join(d.artifactDir, "nodes.yaml"))
+	if err != nil {
+		if _, lerr := d.errLogFile.WriteString(fmt.Sprintf("Error creating nodes.yaml: %v\n", err)); lerr != nil {
+			err = fmt.Errorf("%v+ error writing to stderr log file: %v", err, lerr)
+		}
+		return err
```
elezar, line 225: This is definitely a patter that we're repeating a lot. Can we reduce the boilerplate here?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+func (d *DiagnosticCollector) Nodes(ctx context.Context) error {
+	nodes, err := d.clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		if _, lerr := d.errLogFile.WriteString(fmt.Sprintf("Error getting nodes: %v\n", err)); lerr != nil {
+			err = fmt.Errorf("%v+ error writing to stderr log file: %v", err, lerr)
+		}
+		return err
+	}
+
+	nodesFile, err := os.Create(filepath.Join(d.artifactDir, "nodes.yaml"))
```
elezar, line 220: what about:
```
func (d *DiagnosticCollector) createFile(fp string) (io.WriteCloser,  error) {
    outfile, err := os.Create(filepath.Join(d.artifactDir, fp))
    if err != nil {
        return nil, d.logError("Error creating %v: %w", fp, err)
    }
    return outfile, nil
}
```

and then using `createFile` here and elsewhere?

tests/e2e/common/diagnosticcollector.go
elezar, line 225: Note that we lose the filename here.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+			err = fmt.Errorf("%v+ error writing to stderr log file: %v", err, lerr)
+		}
+		return err
+	}
+
+	return nil
+}
+
+// Namespace is a collectorFunc that collects namespace objects
+func (d *DiagnosticCollector) Namespaces(ctx context.Context) error {
```
elezar, line 249: All these `collectorFunc`  function all seem to follow a similar pattern:
1. Fetch objects
2. Create file
3. Marshal objects to file

Can this be factored out into common functions?

Also, since you use `collectorFunc` here is this actually an interface?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,564 @@
...
+	jobsFile, err := os.Create(filepath.Join(d.artifactDir, "jobs.yaml"))
+	if err != nil {
+		if _, lerr := d.errLogFile.WriteString(fmt.Sprintf("Error creating jobs.yaml: %v\n", err)); lerr != nil {
+			err = fmt.Errorf("%v+ error writing to stderr log file: %v", err, lerr)
+		}
+		return err
+	}
+	defer jobsFile.Close()
+
+	data, err := yaml.Marshal(jobs)
```
elezar, line 457: Do jobs also have logs like pods?

tests/e2e/Makefile
```diff
@@ -40,4 +41,5 @@ e2e-test:
 		-log-artifacts=$(LOG_ARTIFACTS) \
 		-helm-chart=$(HELM_CHART) \
 		-ginkgo.focus="\[nvidia\]" \
-		-test.timeout=1h
+		-test.timeout=1h \
+		-collect-logs=$(COLLECT_LOGS_FROM)
```
elezar, line 45: Does it make sense to optionally add this argument if `COLLECT_LOGS_FROM` is defined instead? `default` is already defined as the default value in the go code.

elezar (commented):
I definitely like the rename!

I think the implementation can be cleaned up a bit. Please see my suggestions for reusing more functionality.

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: jobs create pods, so the function Pods will collect the logs. no need to do a log collection here

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: Part 1 of the question: got it. will address
Part 2: `collectorFunc` is a type so we can then call funcs just by name

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: check

ArangoGutierrez: check

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: check

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: check

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: check

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: check

tests/e2e/device-plugin_test.go
```diff
@@ -104,13 +125,16 @@ var _ = NVDescribe("GPU Device Plugin", func() {
...
+			Expect(err).NotTo(HaveOccurred())
 		})
 
 		// Cleanup before next test run
 		AfterEach(func(ctx context.Context) {
 			// Gather logs
 			if CurrentSpecReport().Failed() {
-				_ = common.MustGather("k8s-device-plugin", f.Namespace.Name, filepath.Join(*LogArtifactDir, f.UniqueName))
+				_ = diagnosticCollector.RunDiagnostics(ctx)
```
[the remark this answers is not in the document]
ArangoGutierrez: done

tests/e2e/gpu-feature-discovery_test.go
ArangoGutierrez: no harm, if no jobs, it will simply do nothing because it won't find any jobs objects.

My idea is to make this modular so it can be used for any component e2e test. as discussed previously, I am already thinking this `DiagnosticsCollector` should live at `k8s-test-infra` so it can be imported by all the projects. so it might seem irrelevant here, I agree, but moving forward the `DiagnosticCollector` could also have `ServiceAccount` `role` and many other k8s objects we consider important during  debugging

ArangoGutierrez: no harm, if no jobs, it will simply do nothing because it won't find any jobs objects.

My idea is to make this modular so it can be used for any component e2e test. as discussed previously, I am already thinking this `DiagnosticsCollector` should live at `k8s-test-infra` so it can be imported by all the projects. so it might seem irrelevant here, I agree, but moving forward the `DiagnosticCollector` could also have `ServiceAccount` `role` and many other k8s objects we consider important during  debugging

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: yes, good catch, users could by mistake have something like `pods,pods,namespaces`

tests/e2e/Makefile
ArangoGutierrez: I see no harm, since `COLLECT_LOGS_FROM` is set to `"default"` , we won't trigger anything special.

tests/e2e/device-plugin_test.go
ArangoGutierrez: check

tests/e2e/device-plugin_test.go
ArangoGutierrez: check

tests/e2e/device-plugin_test.go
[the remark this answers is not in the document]
ArangoGutierrez: I think it makes sense to have 1 for Device-plugin and 1 for GFD.

tests/e2e/device-plugin_test.go
```diff
@@ -57,8 +58,21 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 			chartSpec       helm.ChartSpec
 			helmReleaseName string
 			kubeconfig      []byte
+
+			collectLogsFrom     string
```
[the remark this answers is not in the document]
ArangoGutierrez: done

tests/e2e/common/diagnosticcollector.go
elezar: As I mentioned offline. I think what I'm seeing here is that if whe introduce the following interface:
```
type Collector interface {
    Collect(Context) error
}
```
then each of the `collectorFunc`s is just an implementation of such an interface. Does it make sense to implement each of these `Collectors` (e.g. for Namespaces) as a separate concrete type that implements the interface? The top-level `DiagnosticsCollector` would then define a slice of these collectors -- and we could also nest them futher if required. One place where this may be reasonable is Pod logs, for example.

tests/e2e/common/diagnosticcollector.go
elezar: Out of scope for this PR: Do we need to store the relationship between the job and the pods that it may create?

tests/e2e/Makefile
elezar: My question was more around defining the default behaviour in two places -- meaning that we need to keep these in sync.

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: ummm it would require some work, but it could be helpful once we move into more complex text matrix with multiple tests and jobs. totally a followup

tests/e2e/Makefile
ArangoGutierrez: I get it, yeah makes sense to not have the flag set in the makefile.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+	namespacesFile, err := c.createFile("namespaces.yaml")
+	if err != nil {
+		return err
+	}
+	defer namespacesFile.Close()
+
+	if err = c.writeToFile(namespacesFile, namespaces); err != nil {
+		return err
+	}
+
```
elezar, line 169: If we introduce the following:
```suggestion
func (c *Config) outputTo(filename string, objects interface{}) error {
	outputfile, err := c.createFile(filename)
	if err != nil {
		return err
	}
	defer outputfile.Close()

	if err = c.writeToFile(outputfile, objects); err != nil {
		return err
	}
	return nil
}
```
we can use `c.outputTo("namespace.yaml", namespaces)` for each of these.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+	return nil
+}
+
+func (c namespaces) Collect(ctx context.Context) error {
+	namespaces, err := c.Config.clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		if _, wErr := fmt.Fprintf(c.errLog, "Error getting namespaces: %v\n", err); wErr != nil {
+			return errors.Join(err, wErr)
+		}
+	}
```
elezar, line 158: Assuming that we can configure `klog` to use `errLog` we could just do:
```suggestion
	if err != nil {
		klog.ErrorS(err, "Error getting namepaces")
	}
```

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+
+	namespacesFile, err := c.createFile("namespaces.yaml")
```
elezar, line 160: Note that this is executed even if we fail to retrieve the namespaces.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+	if err = c.writeToFile(nodesFile, nodes); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c namespaces) Collect(ctx context.Context) error {
+	namespaces, err := c.Config.clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
```
elezar, line 154: Should we explicitly handle `errors.IsNotExists`?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+	for _, pod := range pods.Items {
+		podLogFile, err := c.createFile(fmt.Sprintf("%s_logs.log", pod.Name))
+		if err != nil {
+			return err
+		}
+		defer podLogFile.Close()
+
+		podLogOpts := v1.PodLogOptions{}
+		req := c.clientset.CoreV1().Pods(c.namespace).GetLogs(pod.Name, &podLogOpts)
+		podLogs, err := req.Stream(ctx)
```
elezar, line 243: Question: At which point is this stream closed? Is it only if the pod is stopped?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+
+	// Get all logs per pod
+	for _, pod := range pods.Items {
+		podLogFile, err := c.createFile(fmt.Sprintf("%s_logs.log", pod.Name))
+		if err != nil {
+			return err
+		}
+		defer podLogFile.Close()
+
+		podLogOpts := v1.PodLogOptions{}
```
elezar, line 241: Any reason for this local?

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+			n, err := podLogs.Read(buf)
+			if err != nil {
+				break
+			}
+			_, err = podLogFile.Write(buf[:n])
+			if err != nil {
+				if _, wErr := fmt.Fprintf(c.errLog, "Error writing pod logs: %v\n", err); wErr != nil {
+					return errors.Join(err, wErr)
+				}
+			}
```
elezar, line 261: Let's use a `bufio.Scanner` for the buffering instead of doing it manually: https://pkg.go.dev/bufio#NewScanner
```suggestion
        buf := bufio.NewScanner(podLogs)
        for buf.Scan() {
            _, err := podLogFile.Write(buf.Bytes())
        }
        if err := buf.Err(); err != nil {
            klog.ErrorS(err, "error reading pod log")
        }
```

elezar (changes requested):
Thanks for the rework, @ArangoGutierrez. 

I have made some suggestions for improvements. I think it would also make sense to pull this into a separate package under `common` at this stage already -- especially since I think we can benefit from different files for the implementation.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+type nodeFeatureRules struct {
+	*Config
+	NfdClient *nfdclient.Clientset
+}
+
+type Config struct {
+	clientset *kubernetes.Clientset
+	NfdClient *nfdclient.Clientset
```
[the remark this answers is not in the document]
ArangoGutierrez: is not duplicated, one is a kubernetes client and the other is a nfd client

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+		d.NfdClient = nfdClient
+	}
+}
+
+func WithObjects(objects ...string) func(*DiagnosticCollector) {
+	// First let's remove duplicates
+	objectsMap := make(map[string]struct{})
+	for _, obj := range objects {
+		objectsMap[obj] = struct{}{}
+	}
```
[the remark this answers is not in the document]
ArangoGutierrez: not really, we will collect all info and dump in individual files, so doesn't matter if we start at pods and end at nodes or backwards

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+func (d *DiagnosticCollector) Collect(ctx context.Context) error {
+	// Redirect stdout and stderr to logs
+	logFile, err := d.createFile("diagnostic_collector.stdout.log")
+	if err != nil {
+		return err
+	}
+	defer logFile.Close()
+	d.log = logFile
```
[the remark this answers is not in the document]
ArangoGutierrez: >Question: Do we need a separate log and error log?
we are printing 2 files, one for stdout and one for stderr. 

For the rest of your comment: We don't wanna do that, since this tool is mostly to generate files (artifacts) that can later be used to debug a failed test run

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+	errLogFile, err := d.createFile("diagnostic_collector.stderr.log")
+	if err != nil {
+		return err
+	}
+	defer errLogFile.Close()
+	d.errLog = errLogFile
+
+	if len(d.collectors) == 0 {
+		// Nothing to collect
+		klog.Warning("No collectors to run")
```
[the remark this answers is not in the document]
ArangoGutierrez: I don't think so, I think is something we can safely log. right now this is not being hit. but I am thinking once we move this to be an importable pkg, and people will start using the `DiagnosticCollector` on their own.

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+
+	// the apps group
+	Deployments = "deployments"
+	DaemonSets  = "daemonsets"
+
+	// the batch group
+	Jobs = "jobs"
+
+	// Supported extensions
+	NodeFeature     = "nodeFeature"
```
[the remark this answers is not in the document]
ArangoGutierrez: No, the CRD name is `NodeFeature`, we could allow non-case sensitive `nodefeature` but not with a hyphen

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: yes, in theory pods should be on a terminated state. but we never know, maybe `ginkgo` failed to clean up the node and the pods are hanging or something for some time

ArangoGutierrez: No, I commented out the call to the log collector and ran the test, same behavior.

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: we are not setting any opts, so is to have the same Opts for each call on the loop when running `req := c.clientset.CoreV1().Pods(c.namespace).GetLogs(pod.Name, &podLogOpts)`

tests/e2e/common/diagnosticcollector.go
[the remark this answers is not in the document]
elezar: Ok. It was not clear that this API (what a user can specify) maps directly to CRD names.

tests/e2e/common/diagnosticcollector.go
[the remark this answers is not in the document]
elezar: I mean that the nodeFeature structs also have a reference.

tests/e2e/device-plugin_test.go
```diff
@@ -108,9 +126,14 @@ var _ = NVDescribe("GPU Device Plugin", func() {
...
-			// Gather logs
+			// Run diagnostic collector if test failed
 			if CurrentSpecReport().Failed() {
-				_ = common.MustGather("k8s-device-plugin", f.Namespace.Name, filepath.Join(*LogArtifactDir, f.UniqueName))
+				diagnosticCollector, _ = common.NewDiagnosticCollector(
+					common.WithNamespace(f.Namespace.Name),
+					common.WithArtifactDir(*LogArtifactDir),
+					common.WithObjects(collectLogsFrom...),
+				)
+				_ = diagnosticCollector.Collect(ctx)
```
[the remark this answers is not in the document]
ArangoGutierrez: check

tests/e2e/common/diagnosticcollector.go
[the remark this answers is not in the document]
ArangoGutierrez: ah good catch

tests/e2e/common/diagnosticcollector.go
```diff
@@ -0,0 +1,465 @@
...
+	NodeFeatureRule = "nodeFeatureRule"
+)
+
+type Collector interface {
+	Collect(context.Context) error
+}
+
+type nodes struct {
```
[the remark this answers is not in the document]
elezar: Sorry. Typo. I mean that we have a file per collector.

tests/e2e/common/diagnosticcollector.go
elezar: OK. I don't really follow, because this isntantiation is IN the loop. Does this not mean that there is no difference between what we have here and:
```
req := c.clientset.CoreV1().Pods(c.namespace).GetLogs(pod.Name, &v1.PodLogOptions{})
```

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: All errors will lead to the same process, printing an error log and returning, bc there's nothing left to do if there's no object to work (to print to a file)

tests/e2e/common/diagnosticcollector.go
ArangoGutierrez: no, the err check goes to return err, so this won't be hit

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,229 @@
...
+	// Redirect stdout and stderr to logs
+	logFile, err := d.createFile("diagnostic_collector.log")
+	if err != nil {
+		return err
+	}
+	defer logFile.Close()
+	d.log = logFile
+
+	// configure klog to write to the log file
+	klog.SetOutput(d.log)
```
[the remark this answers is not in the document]
elezar: My question is whether we want to distinguish between output that is logged to the user and output that is logged for archiving. If we wanted different output for the user, the we would *not* override the global logger. If we wanted all output to end up in the file and not be visible to the tests, for example, then doing it as we do now is fine.

If we extend this to the `kubectl-nv` use case, there are probably some outputs that we want to output to `STDERR` instead of the configured file.

I'm happy to address this once that becomes a concern though.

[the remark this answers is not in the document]
ArangoGutierrez: umm yup, klog takes loggr to enable custom configuration, we could create a MultiWritter to write to stdout and file

tests/e2e/common/diagnostics/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		klog.ErrorS(err, "Error getting nodes")
+		return err
+	}
+
+	if err := c.outputTo("nodes.yaml", nodes); err != nil {
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
elezar: OK. I take this back. I think the API types may cause issues here. Let's rather not include this change in this PR.

tests/e2e/common/diagnostics/collector.go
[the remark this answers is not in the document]
ArangoGutierrez: I will do this one as a follow up, I want to do a PR to enhance the overall logging system of the E2E, I am not happy with how it its right now

tests/e2e/common/diagnostics/options.go
```diff
@@ -0,0 +1,97 @@
...
+
+func WithObjects(objects ...string) func(*Diagnostic) {
+	return func(d *Diagnostic) {
+		seen := make(map[string]bool)
+		for _, obj := range objects {
+			if seen[obj] {
+				continue
+			}
+			seen[obj] = true
+			switch obj {
```
elezar, line 63: Nit: We don't handle the case where an input is provided that's not a supported object.

tests/e2e/common/diagnostics/options.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func New(opts ...Option) (*Diagnostic, error) {
+	c := &Config{}
+	dc := &Diagnostic{
+		Config: c,
+	}
+
+	// use the variadic function to set the options
+	for _, opt := range opts {
```
elezar, line 92: What is the expected behaviour for an empty list of options? Are any of the arguments required?

If we call `.Collect()` on the result of an empty list of options, we're going to try to create a diretory that is the result of `filepath.Join("", "")`. What is the behaviour of this?

tests/e2e/common/kubernetes.go
```diff
@@ -213,3 +215,18 @@ func cleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset) {
 		}
 	}
 }
+
+// CreateKubernetesClient creates a Kubernetes clientset from the given kubeconfig file
+func CreateKubernetesClient(kubeconfig string) (*kubernetes.Clientset, error) {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
```
elezar, line 221: Is this the recommended way to do this? We already construct a client in the `internal/flags` package. Should we reuse that for the tests, or are there concerns with portability?

tests/e2e/common/kubernetes.go
```diff
@@ -213,3 +215,18 @@ func cleanupNodeFeatureRules(ctx context.Context, cli *nfdclient.Clientset) {
 		}
 	}
 }
+
+// CreateKubernetesClient creates a Kubernetes clientset from the given kubeconfig file
+func CreateKubernetesClient(kubeconfig string) (*kubernetes.Clientset, error) {
```
elezar, line 220: Given that our tests already check kubernetes objects using a client, I'm surprised to see this function added **without** additional modifications to the existing code. Can we reuse what we already have to create a client?

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	defer outputfile.Close()
+	if err = c.writeToFile(outputfile, objects); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Diagnostic) Collect(ctx context.Context) error {
+	var err error
```
elezar, line 85: We don't need `err` here?

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	// Run the collectors
+	for _, c := range d.collectors {
+		if err := c.Collect(ctx); err != nil {
+			klog.ErrorS(err, "Error running collector")
+		}
+	}
+
+	return err
```
elezar, line 114: Should this return `nil`?

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return outfile, nil
+}
+
+func (c *Config) writeToFile(w io.Writer, data interface{}) error {
+	// Marshal data to YAML format
+	yamlBytes, err := yaml.Marshal(data)
+	if err != nil {
+		klog.ErrorS(err, "Error marshaling data")
+		return err
```
elezar, line 59: Let's just return an error here and log at the top-level instead.
```suggestion
		return fmt.Errorf("error marshalling data: %w", err)
```

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	// Write marshaled bytes to the provided io.Writer
+	_, err = w.Write(yamlBytes)
+	if err != nil {
+		klog.ErrorS(err, "Error writing to file")
+		return err
```
elezar, line 66:
```suggestion
		return err
```

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return nil
+}
+
+func (c *Config) outputTo(filename string, objects interface{}) error {
+	outputfile, err := c.createFile(filename)
+	if err != nil {
+		return err
```
elezar, line 75:
```suggestion
		return fmt.Errorf("error creating %v: %w", filename, err)
```

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	defer outputfile.Close()
+	if err = c.writeToFile(outputfile, objects); err != nil {
+		return err
```
elezar, line 79:
```suggestion
		return fmt.Errorf("error writing to %v: %w", filename, err)
```

tests/e2e/common/diagnostics/extensions.go
```diff
@@ -0,0 +1,60 @@
...
+
+type nodeFeatureRules struct {
+	*Config
+}
+
+func (c nodeFeatures) Collect(ctx context.Context) error {
+	nfs, err := c.NfdClient.NfdV1alpha1().NodeFeatures(c.namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.ErrorS(err, "Error getting node features")
+		return err
```
elezar, line 38:
```suggestion
		return fmt.Errorf("error collecting %t: %w", c, err)
```
This should ensure that we log the type and the error that was raised.

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Create the artifact directory
+	if err := os.MkdirAll(filepath.Join(d.Config.artifactDir, d.Config.namespace), os.ModePerm); err != nil {
+		return err
+	}
+
+	// Redirect stdout and stderr to logs
+	logFile, err := d.createFile("diagnostic_collector.log")
+	if err != nil {
+		return err
```
elezar, line 95:
```suggestion
		return fmt.Errorf("error creating collector log file: %w", err)
```

tests/e2e/common/diagnostics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return nil
+}
+
+func (d *Diagnostic) Collect(ctx context.Context) error {
+	var err error
+
+	// Create the artifact directory
+	if err := os.MkdirAll(filepath.Join(d.Config.artifactDir, d.Config.namespace), os.ModePerm); err != nil {
+		return err
```
elezar, line 89: Do we want to add context to the error?

tests/e2e/common/diagnostics/extensions.go
elezar: This probably holds for all other `Collect` implementations apart from the top-level one.

tests/e2e/common/diagnostics/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+
+	if err := c.outputTo("pods.yaml", pods); err != nil {
+		return err
+	}
+
+	// Get all logs per pod
+	for _, pod := range pods.Items {
+		podLogFile, err := c.createFile(fmt.Sprintf("%s_logs.log", pod.Name))
```
elezar, line 122:
```suggestion
		podLogFile, err := c.createFile(fmt.Sprintf("%s.log", pod.Name))
```

tests/e2e/common/diagnostics/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		if err != nil {
+			return err
+		}
+		defer podLogFile.Close()
+
+		req := c.Clientset.CoreV1().Pods(c.namespace).GetLogs(pod.Name, &v1.PodLogOptions{})
+		podLogs, err := req.Stream(ctx)
+		if err != nil {
+			klog.ErrorS(err, "error getting pod logs")
```
elezar, line 131:
```suggestion
			klog.ErrorS(err, "error getting pod logs", "podName", pod.Name)
```

tests/e2e/common/diagnostics/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+
+	// Get all logs per pod
+	for _, pod := range pods.Items {
+		podLogFile, err := c.createFile(fmt.Sprintf("%s_logs.log", pod.Name))
+		if err != nil {
+			return err
+		}
+		defer podLogFile.Close()
```
elezar, line 126: Question: Is this executed on each loop iteration?

tests/e2e/device-plugin_test.go
```diff
@@ -108,9 +127,19 @@ var _ = NVDescribe("GPU Device Plugin", func() {
...
 			if CurrentSpecReport().Failed() {
-				_ = common.MustGather("k8s-device-plugin", f.Namespace.Name, filepath.Join(*LogArtifactDir, f.UniqueName))
+				var err error
+				diagnostic, err = diagnostics.New(
+					diagnostics.WithNamespace(f.Namespace.Name),
+					diagnostics.WithArtifactDir(*LogArtifactDir),
+					diagnostics.WithKubernetesClient(f.ClientSet),
+					diagnostics.WithObjects(collectLogsFrom...),
+				)
+				Expect(err).NotTo(HaveOccurred())
```
elezar, line 139: I think we also just want to log this error as we do for the call to `Collect()` below.

tests/e2e/gpu-feature-discovery_test.go
```diff
@@ -79,6 +89,9 @@ var _ = NVDescribe("GPU Feature Discovery", func() {
 			chartSpec       helm.ChartSpec
 			helmReleaseName string
 			kubeconfig      []byte
+
+			collectLogsFrom []string
+			diagnostic      *diagnostics.Diagnostic
```
elezar, line 94:
```suggestion
			diagnosticsCollector      diagnostics.Collector
```

tests/e2e/device-plugin_test.go
```diff
@@ -57,8 +58,20 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 			chartSpec       helm.ChartSpec
 			helmReleaseName string
 			kubeconfig      []byte
+
+			collectLogsFrom []string
+			diagnostic      *diagnostics.Diagnostic
```
elezar, line 63:
```suggestion
			diagnosticsCollector      diagnostics.Collector
```

elezar (commented):
The last missing bit is cleaning up how we're logging errors. 

Here, we should log a the top-level and ensure that sufficient context is added to the errors to provide information to debug issues.

I don't think any of these are blockers, but I would rather address them here before we port this code to `kubectl-nv`.

tests/e2e/common/diagnostics/kubernetes.go
ArangoGutierrez: each loop iteration is a different pod, and will be saved to a diff file.

elezar: Yes, I understand. My question was whether the defer statement is triggered in each loop.

Looking at https://go.dev/play/p/ggT4Sim9X1u, this is not the case. It seems as if we would first create all files before unrolling the defers -- that is to say that the loop body doesn't represent a conext that is exited.

The simplest solution to ensure that the pod log is created and closed in the same iteration is to call a function in the loop body instead. See https://go.dev/play/p/vuQndJk6B0U

We could do the following to ensure that we are closing these files properly:
```
diff --git a/tests/e2e/common/diagnostics/kubernetes.go b/tests/e2e/common/diagnostics/kubernetes.go
index 637a95ac..4dd88d4a 100644
--- a/tests/e2e/common/diagnostics/kubernetes.go
+++ b/tests/e2e/common/diagnostics/kubernetes.go
@@ -19,11 +19,11 @@ package diagnostics
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/klog/v2"
 )
 
 type nodes struct {
@@ -112,32 +112,50 @@ func (c pods) Collect(ctx context.Context) error {
 		return err
 	}
 
-	// Get all logs per pod
+	var podLogCollectors []Collector
 	for _, pod := range pods.Items {
-		podLogFile, err := c.createFile(fmt.Sprintf("%s.log", pod.Name))
-		if err != nil {
-			return err
+		plc := podLogCollector{
+			Config: c.Config,
+			name:   pod.Name,
 		}
-		defer podLogFile.Close()
+		podLogCollectors = append(podLogCollectors, plc)
+	}
 
-		req := c.Clientset.CoreV1().Pods(c.namespace).GetLogs(pod.Name, &v1.PodLogOptions{})
-		podLogs, err := req.Stream(ctx)
-		if err != nil {
-			klog.ErrorS(err, "error getting pod logs")
-			continue
-		}
+	var errs error
+	for _, collector := range podLogCollectors {
+		errs = errors.Join(errs, collector.Collect(ctx))
+	}
 
-		buf := bufio.NewScanner(podLogs)
-		for buf.Scan() {
-			if _, err := podLogFile.Write(buf.Bytes()); err != nil {
-				klog.ErrorS(err, "error getting pod logs", "podName", pod.Name)
-			}
-		}
-		if err := buf.Err(); err != nil {
-			klog.ErrorS(err, "error reading pod log")
-		}
+	return errs
+}
+
+type podLogCollector struct {
+	*Config
+	name string
+}
+
+func (c podLogCollector) Collect(ctx context.Context) error {
+	podLogFile, err := c.createFile(fmt.Sprintf("%s.log", c.name))
+	if err != nil {
+		return err
 	}
+	defer podLogFile.Close()
+
+	req := c.Clientset.CoreV1().Pods(c.namespace).GetLogs(c.name, &v1.PodLogOptions{})
+	podLogs, err := req.Stream(ctx)
+	if err != nil {
+		return fmt.Errorf("error getting logs for pod %v: %w", c.name, err)
 
+	}
+	buf := bufio.NewScanner(podLogs)
+	for buf.Scan() {
+		if _, err := podLogFile.Write(buf.Bytes()); err != nil {
+			return fmt.Errorf("error writing logs for pod %v: %w", c.name, err)
+		}
+	}
+	if err := buf.Err(); err != nil {
+		return fmt.Errorf("error reading logs for pod %v: %w", c.name, err)
+	}
 	return nil
 }
```

ArangoGutierrez: Thanks for the suggestion, I implemented it, with a small mod

elezar (approved):
LGTM.

Issue #657, opened 2024-04-19
When use MPS, add a initContainers to default set compute model


lengrongfu:
Because when using MPS, it usually suggests GPU compute mode is `EXCLUSIVE_PROCESS`, we can use an initContainer to set GPU compute mode.

PR #658, opened 2024-04-19
add initContainer to set compute mode

lengrongfu:
Fixes: https://github.com/NVIDIA/k8s-device-plugin/issues/657

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
@@ -56,6 +56,11 @@ spec:
       shareProcessNamespace: true
       {{- end }}
       initContainers:
+      - name: set-compute-mode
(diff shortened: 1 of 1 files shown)


elezar (commented):
@lengrongfu we cannot unconditionally set the compute mode in the device plugin. This will break use cases such as sharing using time-slicing. We also need to restore the compute mode once the device plugin is uninstalled. Note that the `v0.15.0` releaes includes experimental support for MPS. When this is configured, a separate daemonset is started that **does** [set the compute mode](https://github.com/NVIDIA/k8s-device-plugin/blob/e7be49d3a2153c03ae4e682714cabd1c69f8642c/cmd/mps-control-daemon/mps/daemon.go#L89-L91). It also ensures that this is restored when the MPS daemon is shut down.

If you have a use case that is not covered by this flow, feel free to update #657 describing what is requried. We can assess this and if applicable make changes to our MPS implementation.

PR #666, opened 2024-04-19
Support non-standard driver installs

elezar:
Thes changes add support to the NVIDIA GPU Device Plugin for driver
installations that have the following properties:
* driver files are installed at a `hostDriverRoot`, say: `/host/nvidia/driver`
* device nodes are created at `/dev`

In addition to the mounts of driver libraries (e.g. `libnvidia-ml.so.VERSION`)
and device nodes that are handled by the NVIDIA Container Runtime or CDI to
allow the device plugin to detect and enumerate devices, the `hostDriverRoot`
is mounted into the device plugin at `/driver-root`. This allows the detection
of driver files that are not required for the device plugin to fuction, but may
be required by specific workloads.

From the perspective of the CDI spec generation being run in the device plugin
container, all driver files are rooted at `/driver-root`. Furthermore, since
no device nodes are detected at `/driver-root/dev` we assume that these are
at `/dev` in the container (and on the host). The generated CDI specifications
need to be transformed so that they are valid for container workloads started
on the host. This means that occurrences of `/driver-root` need to be
transformed to `hostDriverRoot` or `/host/nvidia/driver` as per our example.

### DIFF
--- a/api/config/v1/flags.go
+++ b/api/config/v1/flags.go
@@ -148,7 +148,7 @@ func (f *Flags) UpdateFromCLIFlags(c *cli.Context, flags []cli.Flag) {
 				updateFromCLIFlag(&f.Plugin.DeviceIDStrategy, c, n)
 			case "cdi-annotation-prefix":
 				updateFromCLIFlag(&f.Plugin.CDIAnnotationPrefix, c, n)
-			case "nvidia-ctk-path":
+			case "nvidia-cdi-hook-path", "nvidia-ctk-path":
 				updateFromCLIFlag(&f.Plugin.NvidiaCTKPath, c, n)
 			case "container-driver-root":
 				updateFromCLIFlag(&f.Plugin.ContainerDriverRoot, c, n)
--- a/api/config/v1/flags.go
+++ b/api/config/v1/flags.go
@@ -59,6 +59,7 @@ type CommandLineFlags struct {
 	FailOnInitError         *bool                   `json:"failOnInitError"            yaml:"failOnInitError"`
 	MpsRoot                 *string                 `json:"mpsRoot,omitempty"          yaml:"mpsRoot,omitempty"`
 	NvidiaDriverRoot        *string                 `json:"nvidiaDriverRoot,omitempty" yaml:"nvidiaDriverRoot,omitempty"`
+	NvidiaDevRoot           *string                 `json:"nvidiaDevRoot,omitempty"    yaml:"nvidiaDevRoot,omitempty"`
 	GDSEnabled              *bool                   `json:"gdsEnabled"                 yaml:"gdsEnabled"`
 	MOFEDEnabled            *bool                   `json:"mofedEnabled"               yaml:"mofedEnabled"`
 	UseNodeFeatureAPI       *bool                   `json:"useNodeFeatureAPI"          yaml:"useNodeFeatureAPI"`
@@ -121,8 +122,10 @@ func (f *Flags) UpdateFromCLIFlags(c *cli.Context, flags []cli.Flag) {
 				updateFromCLIFlag(&f.FailOnInitError, c, n)
 			case "mps-root":
 				updateFromCLIFlag(&f.MpsRoot, c, n)
-			case "nvidia-driver-root":
+			case "driver-root", "nvidia-driver-root":
 				updateFromCLIFlag(&f.NvidiaDriverRoot, c, n)
+			case "dev-root", "nvidia-dev-root":
+				updateFromCLIFlag(&f.NvidiaDevRoot, c, n)
 			case "gds-enabled":
 				updateFromCLIFlag(&f.GDSEnabled, c, n)
 			case "mofed-enabled":
--- a/api/config/v1/strategy.go
+++ b/api/config/v1/strategy.go
@@ -48,12 +48,22 @@ func (s DeviceListStrategies) Includes(strategy string) bool {
 	return s[strategy]
 }
 
-// IsCDIEnabled returns whether any of the strategies being used require CDI.
-func (s DeviceListStrategies) IsCDIEnabled() bool {
+// AnyCDIEnabled returns whether any of the strategies being used require CDI.
+func (s DeviceListStrategies) AnyCDIEnabled() bool {
 	for k, v := range s {
 		if strings.HasPrefix(k, "cdi-") && v {
 			return true
 		}
 	}
 	return false
+}
+
+// AllCDIEnabled returns whether all strategies being used require CDI.
+func (s DeviceListStrategies) AllCDIEnabled() bool {
+	for k, v := range s {
+		if !strings.HasPrefix(k, "cdi-") && v {
+			return false
+		}
+	}
+	return true
 }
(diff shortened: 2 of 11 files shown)


api/config/v1/flags.go
```diff
@@ -59,6 +59,7 @@ type CommandLineFlags struct {
 	FailOnInitError   *bool                   `json:"failOnInitError"            yaml:"failOnInitError"`
 	MpsRoot           *string                 `json:"mpsRoot,omitempty"          yaml:"mpsRoot,omitempty"`
 	NvidiaDriverRoot  *string                 `json:"nvidiaDriverRoot,omitempty" yaml:"nvidiaDriverRoot,omitempty"`
+	HostDevRoot       *string                 `json:"hostDevRoot,omitempty"      yaml:"hostDevRoot,omitempty"`
```
cdesiniotis, line 62: Thoughts on naming this field `NvidiaDevRoot` to closer align with `NvidiaDriverRoot`?

internal/plugin/server.go
```diff
@@ -354,8 +354,8 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 	if plugin.config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
 		plugin.updateResponseForMPS(response)
 	}
-	if *plugin.config.Flags.Plugin.PassDeviceSpecs {
-		response.Devices = append(response.Devices, plugin.apiDeviceSpecs(*plugin.config.Flags.NvidiaDriverRoot, requestIds)...)
+	if *plugin.config.Flags.Plugin.PassDeviceSpecs && !plugin.deviceListStrategies.IsNonCDIEnabled() {
```
cdesiniotis, line 357: This double-negative is a bit difficult to read. Would we be able to update this to `*plugin.config.Flags.Plugin.PassDeviceSpecs && plugin.deviceListStrategies.IsCDIEnabled()`?

Also, I am not quite sure I am following this check. Why does it matter that CDI is enabled when `--pass-device-specs` is set?

internal/cdi/cdi.go
```diff
@@ -152,15 +166,23 @@ func (cdi *cdiHandler) CreateSpecFile() error {
...
 		}
 
+		// The transformation above may change the required spec version.
+		// TODO: This should be integrated into the transformer.
+		minVersion, err := cdiapi.MinimumRequiredVersion(spec.Raw())
+		if err != nil {
+			return fmt.Errorf("failed to get minimum required CDI spec version: %v", err)
+		}
+		spec.Raw().Version = minVersion
+
```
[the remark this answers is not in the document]
elezar: The reason this is required is that devices with different host paths to device paths require a v0.5.0 spec. This means that the transform changes the initial spec version. I will look at how we can handle this through transformer options instead.

api/config/v1/flags.go
elezar: Yes, I think we could. Will update.

cmd/nvidia-device-plugin/plugin-manager.go
```diff
@@ -46,7 +51,8 @@ func NewPluginManager(config *spec.Config) (manager.Interface, error) {
 
 	cdiHandler, err := cdi.New(
 		cdi.WithDeviceListStrategies(deviceListStrategies),
-		cdi.WithDriverRoot(*config.Flags.Plugin.ContainerDriverRoot),
+		cdi.WithDriverRoot(string(driverRoot)),
+		cdi.WithDevRoot(driverRoot.getDevRoot()),
```
cdesiniotis, line 55: Question -- why are we not using the `NVIDIA_DEVICE_ROOT` option here that was added to the device plugin CLI? From what I gather, `driverRoot.getDevRoot()` will always resolve to either `driverRoot` or `/`. What if a custom dev root is configured that is neither of these paths?

elezar: The `devRoot` from the context of CDI in this case is the dev root in the container. This is determined from the `driverRoot` (in the container) by checking if `{{ .driverRoot }}/dev` exists and using that, otherwise using `/dev`.

cmd/nvidia-device-plugin/plugin-manager.go
```diff
@@ -39,15 +39,20 @@ func NewPluginManager(infolib info.Interface, nvmllib nvml.Interface, devicelib
...
 		return nil, fmt.Errorf("invalid device list strategy: %v", err)
 	}
 
 	cdiHandler, err := cdi.New(infolib, nvmllib, devicelib,
 		cdi.WithDeviceListStrategies(deviceListStrategies),
-		cdi.WithDriverRoot(*config.Flags.Plugin.ContainerDriverRoot),
+		cdi.WithDriverRoot(string(driverRoot)),
+		cdi.WithDevRoot(driverRoot.getDevRoot()),
 		cdi.WithTargetDriverRoot(*config.Flags.NvidiaDriverRoot),
+		cdi.WithTargetDevRoot(*config.Flags.NvidiaDevRoot),
```
elezar, line 55: Note that the `TargetDevRoot` is the dev root on the host. This is set to `NvidiaDevRoot`.

internal/cdi/cdi.go
```diff
@@ -66,6 +69,18 @@ func newHandler(opts ...Option) (Interface, error) {
 	if !c.deviceListStrategies.IsCDIEnabled() {
 		return &null{}, nil
 	}
+	if c.driverRoot == "" {
```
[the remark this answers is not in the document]
elezar: Yes, that's probably better to do, but is redundant since we 1. set it in the caller, and 2. also set it in the CDI api. Let me clean it up regardless.

internal/plugin/server.go
elezar: The reason for the check is that if *only* CDI is used, then we don't want to inject the device specs in this way since CDI is used to inject the device nodes.

Let me see if I can come up with something that's clearer.

api/config/v1/strategy.go
```diff
@@ -48,12 +48,22 @@ func (s DeviceListStrategies) Includes(strategy string) bool {
...
+func (s DeviceListStrategies) AnyCDIEnabled() bool {
 	for k, v := range s {
 		if strings.HasPrefix(k, "cdi-") && v {
 			return true
 		}
 	}
 	return false
 }
+
+// IsOnlyCDIEnabled returns whether all strategies being used require CDI.
```
cdesiniotis, line 61:
```suggestion
// AllCDIEnabled returns whether all strategies being used require CDI.
```

cmd/nvidia-device-plugin/plugin-manager.go
cdesiniotis: Got it. Let's keep this as is for now.

One thing to consider in the context of the GPU Operator -- I currently set `DEV_ROOT_CTR_PATH` in the `driver-ready` status file, to indicate the container path of the device root. If it makes sense to do so, we could consider leveraging this here in the device-plugin (and other operands).

cdesiniotis (approved):
LGTM

cmd/nvidia-device-plugin/plugin-manager.go
elezar: One thought that I had was that we could set this in the management CDI spec. Will have to check what that would mean though.

PR #674, opened 2024-04-22
Don't require node-name to be set if not needed

elezar:
The `node-name` is only needed if the NodeFeatureAPI is used.

### DIFF
--- a/internal/lm/output.go
+++ b/internal/lm/output.go
@@ -42,15 +42,22 @@ type Outputer interface {
 }
 
 // TODO: Replace this with functional options.
-func NewOutputer(config *spec.Config, nodeConfig flags.NodeConfig, clientSets flags.ClientSets) Outputer {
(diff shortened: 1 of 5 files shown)


internal/lm/output.go
```diff
@@ -153,6 +153,9 @@ type nodeFeatureObject struct {
 // UpdateNodeFeatureObject creates/updates the node-specific NodeFeature custom resource.
 func (n *nodeFeatureObject) Output(labels Labels) error {
 	nodename := n.nodeConfig.Name
+	if nodename == "" {
+		return fmt.Errorf("required flag %q not set", "node-name")
```
ArangoGutierrez, line 157: if we are passing `"node-name"` as a `string` is it really necessary to set `%q` couldn't we simply just have a flat string `"required flag node-name not set" ` ?

Issue #670, opened 2024-04-22
allPossibleMigStrategiesAreNone is false when using default values


jakubkrzykowski:
### 1. Quick Debug Information

N/A

### 2. Issue or feature description

When trying to deploy the latest (also tested on v0.13.0) version of the helm chart with the default values file I can see that `NVIDIA_MIG_MONITOR_DEVICES` env var are being set to `all` and `securityContext.capabilities.add` are being set to `SYS_ADMIN`. 
`allPossibleMigStrategiesAreNone` checks if `migStrategy` is set to "none" but doesn't cover the scenario when this value is null (not set at all), which is the default behavior of the helm chart. 

Is this intentional?

elezar:
@jakubkrzykowski this sounds like a bug. Looking at the helm helper, we definitely don't check for the `null` value.

Would you be able to submit a PR with a fix?

jakubkrzykowski:
Hi @elezar !
I was able to workaround it by setting it explicitly to `none` as you suggested. Just wanted to be sure if this is not intentional.
Happy to raise a PR with a fix.

jakubkrzykowski:
I've checked the _helpers.tpl file thoroughly and found that the problem lies with {{- else if ne (include "nvidia-device-plugin.configMapName" .) "true" -}} in the allPossibleMigStrategiesAreNone definition. With the default values, it will never be true because configMapName will always be an empty string, not the string "true", according to this:
```
{{- define "nvidia-device-plugin.configMapName" -}}
{{- $result := "" -}}
{{- if .Values.config.name -}}
  {{- $result = .Values.config.name -}}
{{- else if not (empty .Values.config.map) -}}
  {{- $result = printf "%s-%s" (include "nvidia-device-plugin.fullname" .) "configs" -}}
{{- end -}}
{{- $result -}}
{{- end -}}
```

jakubkrzykowski:
@elezar here is the PR to address that - https://github.com/NVIDIA/k8s-device-plugin/pull/675

andy108369:
looks like this introduced a bug https://github.com/NVIDIA/k8s-device-plugin/issues/856 where `SYS_ADMIN` cap gets dropped after upgrading to the helm-chart after https://github.com/NVIDIA/k8s-device-plugin/commit/4ca56ae17533724910f6e246f08dd0db7b3c37cd was added.

PR #675, opened 2024-04-22
Fixing allPossibleMigStrategiesAreNone helm chart helper.

jakubkrzykowski:
`configMapName` is the empty string `""` by default and it breaks `allPossibleMigStrategiesAreNone` function with default values.

See https://github.com/NVIDIA/k8s-device-plugin/issues/670 for details.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
+++ b/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
@@ -130,7 +130,7 @@ Check if migStrategy (from all possible configurations) is "none"
   {{- if ne .Values.migStrategy "none" -}}
     {{- $result = false -}}
   {{- end -}}
-{{- else if ne (include "nvidia-device-plugin.configMapName" .) "true" -}}
+{{- else if eq (include "nvidia-device-plugin.hasConfigMap" .) "true" -}}
     {{- $result = false -}}
 {{- else -}}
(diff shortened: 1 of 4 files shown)


elezar (commented):
Sorry, these were pendings.

deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
```diff
@@ -130,7 +130,7 @@ Check if migStrategy (from all possible configurations) is "none"
   {{- if ne .Values.migStrategy "none" -}}
     {{- $result = false -}}
   {{- end -}}
-{{- else if ne (include "nvidia-device-plugin.configMapName" .) "true" -}}
+{{- else if ne (include "nvidia-device-plugin.configMapName" .) "" -}}
```
elezar, line 133: Let's use the following instead:
```suggestion
{{- else if eq (include "nvidia-device-plugin.hasConfigMap" .) "true" -}}
```

elezar (changes requested):

elezar (approved):
Thanks @jakubkrzykowski. These look good now.

I rebased and incorporated the changes from #749

adammw:
having problems with this using the built in config.map option - since a configMapName is set it immediately goes to `$result = false` instead of falling through to evaluate the config.map profiles

PR #676, opened 2024-04-23
Adding support to populate node labels when GPUs are configured for passthrough

visheshtanksale:
Adding support to populate node labels when GPUs are configured for passthrough
GFD does not populate GPU related node labels in passthrough mode because it reads GPU details from NVML. This PR adds support to read GPU details from sysfs when GPU are configured for passthrough

To Do: Adding unit tests


api/config/v1/flags.go
```diff
@@ -101,11 +101,12 @@ func (f *deviceListStrategyFlag) UnmarshalJSON(b []byte) error {
...
-	NoTimestamp     *bool     `json:"noTimestamp"     yaml:"noTimestamp"`
-	SleepInterval   *Duration `json:"sleepInterval"   yaml:"sleepInterval"`
-	OutputFile      *string   `json:"outputFile"      yaml:"outputFile"`
-	MachineTypeFile *string   `json:"machineTypeFile" yaml:"machineTypeFile"`
+	Oneshot               *bool     `json:"oneshot"         yaml:"oneshot"`
+	NoTimestamp           *bool     `json:"noTimestamp"     yaml:"noTimestamp"`
+	SleepInterval         *Duration `json:"sleepInterval"   yaml:"sleepInterval"`
+	OutputFile            *string   `json:"outputFile"      yaml:"outputFile"`
+	MachineTypeFile       *string   `json:"machineTypeFile" yaml:"machineTypeFile"`
+	VMPassthroughWorkload *bool     `json:"vmPassthroughWorkload" yaml:"vmPassthroughWorkload"`
```
klueska, line 109: I would drop the `Workload` and maybe rename the flag completely to something with `vfio` in the name

internal/resource/sysfs-lib.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Shutdown is a no-op for the vfio manager
+func (l *vfioLib) Shutdown() (err error) {
+	return nil
+}
+
+// GetDevices returns the devices with VFIO PCI driver available on the system
+func (l *vfioLib) GetDevices() ([]Device, error) {
+	var devices []Device
+	nvpci := nvpci.New()
```
elezar, line 45: We may want to allow for injection of an `nvpci.Interface` for testing.

internal/resource/sysfs-lib.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, err
+	}
+
+	for _, dev := range nvdevices {
+		if dev.Driver != "vfio-pci" {
+			return nil, fmt.Errorf("device not bound to 'vfio-pci'; device: %s driver: '%s'", dev.Address, dev.Driver)
+		} else {
+			vfioDev := vfioDevice{dev}
+			devices = append(devices, vfioDev)
+		}
```
elezar, line 57: Since we `return` in the previous block, we don't need the explicit else.
```suggestion
		}
		vfioDev := vfioDevice{dev}
		devices = append(devices, vfioDev)
```

internal/resource/sysfs-lib.go
```diff
@@ -0,0 +1,71 @@
...
+	var devices []Device
+	nvpci := nvpci.New()
+	nvdevices, err := nvpci.GetGPUs()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, dev := range nvdevices {
+		if dev.Driver != "vfio-pci" {
+			return nil, fmt.Errorf("device not bound to 'vfio-pci'; device: %s driver: '%s'", dev.Address, dev.Driver)
```
elezar, line 53: Question: Is this an error, or should we just skip the device?

internal/resource/factory.go
```diff
@@ -39,7 +39,7 @@ func WithConfig(manager Manager, config *spec.Config) Manager {
 }
 
 // getManager returns the resource manager depending on the system configuration.
-func getManager() Manager {
+func getManager(config *spec.Config) Manager {
```
elezar, line 42: See https://github.com/NVIDIA/k8s-device-plugin/pull/673 where I reworked this to use go-nvlib functionality that I'm trying to add.

If we add a `mode` flag, I would recommend  only passing the `mode` here and then making a construction decision based on that.

elezar (commented):
Thanks @visheshtanksale. I think this is a good start.

I think adding a more generic mode will allow for more flexibility going forward, and also make it clear that NVML, CUDA and vfio labels are mutually exclusive.

api/config/v1/flags.go
elezar: @klueska as I suggested below, adding a top-level mode flag to switch between this, `nvml`, and `cuda` explicitly may be useful. Do you have specific thoughts on this?

internal/resource/factory.go
```diff
@@ -64,11 +64,18 @@ func getManager() Manager {
...
 		return NewNVMLManager()
-	} else if isTegra {
+	}
+
+	if isTegra {
 		klog.Info("Using CUDA manager")
 		return NewCudaManager()
 	}
 
+	if *config.Flags.GFD.VMPassthroughWorkload {
```
cdesiniotis: I agree with Evan's comment: https://github.com/NVIDIA/k8s-device-plugin/pull/676/files#r1575766699

We should add a new `vfio` mode to the `go-nvlib/info` package, which represents a system where all GPUs are bound to the `vfio-pci` driver. We can construct a `VfioManager` if we detect a `vfio` system, so no need to introduce a new `VMPassthroughWorkload` flag.

internal/resource/sysfs-lib.go
visheshtanksale: I had return by mistake. Fixed it

internal/resource/sysfs-lib.go
visheshtanksale: Yes we should skip the device. Fixed it.

internal/resource/factory.go
visheshtanksale: @elezar This makes sense. How do you want to handle this? Since the go-nvlib PR is still open do you want to update the info interface or should I wait for it to be merged?

elezar: I think the addition of the mode flag is separate from that PR. We can already add the config option and command line flags here and then ensure that it is plumbed through. I would suggest.

1. Add a new PR that adds the mode flag / config option.
2. Rebase these changes to include the mode flag commits, adding support for a `vfio` mode.

Once the go-nvlib changes and #673 are merged, we can update the checks for the existing modes here.

As a follow-up we can pull the `vfio` detection into `go-nvlib`, but I don't see that as a blocker for this PR.

internal/resource/factory.go
elezar: To clarify. The `mode` flag I suggested in the linked comment is local. We could (and probably should) pull some of the functionality into the `go-nvlib/info` package.

The important functionality from the perspective of GFD and the device plugin is that a user can explicitly override the mode if our autodetection fails. This already happens on newer Tegra-based systems, for example.

api/config/v1/flags.go
```diff
@@ -162,6 +163,8 @@ func (f *Flags) UpdateFromCLIFlags(c *cli.Context, flags []cli.Flag) {
 				updateFromCLIFlag(&f.GFD.NoTimestamp, c, n)
 			case "machine-type-file":
 				updateFromCLIFlag(&f.GFD.MachineTypeFile, c, n)
+			case "vfio-pci-driver-mode":
```
ArangoGutierrez, line 166: Could you add this to the Helm templates as well

ArangoGutierrez (changes requested):

internal/resource/sysfs-lib.go
visheshtanksale: Fixed!

api/config/v1/flags.go
```diff
@@ -101,12 +101,12 @@ func (f *deviceListStrategyFlag) UnmarshalJSON(b []byte) error {
...
-	SleepInterval     *Duration `json:"sleepInterval"   yaml:"sleepInterval"`
-	OutputFile        *string   `json:"outputFile"      yaml:"outputFile"`
-	MachineTypeFile   *string   `json:"machineTypeFile" yaml:"machineTypeFile"`
-	VfioPciDriverMode *bool     `json:"vfioPciDriverMode" yaml:"vfioPciDriverMode"`
+	Oneshot         *bool     `json:"oneshot"         yaml:"oneshot"`
+	NoTimestamp     *bool     `json:"noTimestamp"     yaml:"noTimestamp"`
+	SleepInterval   *Duration `json:"sleepInterval"   yaml:"sleepInterval"`
+	OutputFile      *string   `json:"outputFile"      yaml:"outputFile"`
+	MachineTypeFile *string   `json:"machineTypeFile" yaml:"machineTypeFile"`
+	Mode            *string   `json:"mode" yaml:"mode"`
```
elezar, line 109: Question: Is this a top-level flag instead? That is to say, would this mode control both GFD and the device plugin?

cmd/gpu-feature-discovery/main.go
```diff
@@ -99,11 +99,10 @@ func main() {
...
 			EnvVars: []string{"GFD_USE_NODE_FEATURE_API", "USE_NODE_FEATURE_API"},
 		},
-		&cli.BoolFlag{
-			Name:    "vfio-pci-driver-mode",
-			Value:   false,
-			Usage:   "Publish labels when GPUs are configured for passthrough",
-			EnvVars: []string{"GFD_VFIO_PCI_DRIVER_MODE"},
+		&cli.StringFlag{
+			Name:    "mode",
+			Usage:   "Select GFD mode between 'NVML','TEGRA' and 'VFIO'",
```
elezar, line 104: `auto (default)`, `nvml`, `tegra`, `vfio`.

cmd/gpu-feature-discovery/main_test.go
```diff
@@ -209,12 +209,12 @@ func TestRunSleep(t *testing.T) {
...
-					SleepInterval:     ptr(spec.Duration(time.Second)),
-					NoTimestamp:       ptr(false),
-					MachineTypeFile:   ptr(testMachineTypeFile),
-					VfioPciDriverMode: ptr(false),
+					Oneshot:         ptr(false),
+					OutputFile:      ptr("./gfd-test-loop"),
+					SleepInterval:   ptr(spec.Duration(time.Second)),
+					NoTimestamp:     ptr(false),
+					MachineTypeFile: ptr(testMachineTypeFile),
+					Mode:            ptr(""),
```
elezar, line 217: This should ideally default to `auto` -- although we can use `""` to mean this too.

internal/resource/factory.go
```diff
@@ -39,7 +41,7 @@ func WithConfig(manager Manager, config *spec.Config) Manager {
 }
 
 // getManager returns the resource manager depending on the system configuration.
-func getManager(config *spec.Config) Manager {
+func getManager(mode *string) Manager {
```
elezar, line 44: This should accept a `string` and not a `*string`

cmd/gpu-feature-discovery/main.go
elezar: We also want to add a check to `validateFlags` that checks for these valid values.

internal/resource/factory.go
```diff
@@ -71,7 +73,7 @@ func getManager(config *spec.Config) Manager {
 		return NewCudaManager()
 	}
 
-	if *config.Flags.GFD.VfioPciDriverMode {
+	if strings.EqualFold(*mode, "vfio") {
```
elezar, line 76: Let's just use `mode == "vfio"` here and require lower case input.

internal/resource/sysfs-lib.go
```diff
@@ -21,11 +21,18 @@ import (
...
 
-type vfioLib struct{}
+type vfioLib struct {
+	nvpci.Interface
+}
 
 // NewVfioManager returns an resource manger for devices with VFIO PCI driver
 func NewVfioManager() Manager {
-	return &vfioLib{}
+
```
elezar, line 30: Remove the newline.

internal/resource/sysfs-lib.go
```diff
@@ -21,11 +21,18 @@ import (
...
+	nvpcilib := nvpci.New()
```
elezar, line 31: Do we need to pass the driver root here?

internal/resource/sysfs-lib.go
```diff
@@ -21,11 +21,18 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type vfioLib struct{}
+type vfioLib struct {
+	nvpci.Interface
```
elezar, line 25: Do we really want to expose the entire `nvpci.Interface` here? can we use:
```
type vfioLib struct {
    nvpcilib nvpci.Interface
}
```
instead to ensure that we don't do so?

internal/resource/factory.go
```diff
@@ -64,11 +66,18 @@ func getManager() Manager {
 	if hasNVML {
 		klog.Info("Using NVML manager")
 		return NewNVMLManager()
-	} else if isTegra {
+	}
+
+	if isTegra {
```
elezar, line 71: I would recommend first adding the mode flag so that we can properly handle the `nvml` and `tegra` modes and then rebase this PR onto those changes.

internal/resource/sysfs-device.go
```diff
@@ -0,0 +1,66 @@
...
+package resource
+
+import (
+	"fmt"
+
+	"github.com/NVIDIA/go-nvlib/pkg/nvpci"
+)
+
+type vfioDevice struct {
+	*nvpci.NvidiaPCIDevice
```
elezar, line 26: let's rather make this a private member.

elezar (commented):
Thanks @visheshtanksale. I think this is moving in the right direction. It may be good to add the mode flag -- supporting the values `nvml` and `tegra` in a separate PR and then updating this to *add* the `vfio` mode. 

Apart from my implementation-specific coments, the one question I have is wheter the `mode` makes sense as a top-level (i.e. not GFD-specific) option.

internal/resource/factory.go
visheshtanksale: Added this

api/config/v1/flags.go
visheshtanksale: Does running in the TEGRA mode mean anything to device plugin? Also currently we dont support NVML as well. So I feel it makes sense to keep it here.

cmd/gpu-feature-discovery/main_test.go
visheshtanksale: Fixed

cmd/gpu-feature-discovery/main.go
visheshtanksale: Fixed both!

internal/resource/factory.go
visheshtanksale: Fixed

internal/resource/factory.go
visheshtanksale: Fixed

internal/resource/sysfs-lib.go
visheshtanksale: Fixed

internal/resource/sysfs-lib.go
visheshtanksale: Fixed

internal/resource/sysfs-device.go
visheshtanksale: Fixed

internal/resource/sysfs-lib.go
visheshtanksale: What do you mean by driver root?

elezar: I mean whether one of the `Options`:
```
// Option defines a function for passing options to the New() call.
type Option func(*nvpci)

// WithLogger provides an Option to set the logger for the library.
func WithLogger(logger logger) Option {
	return func(n *nvpci) {
		n.logger = logger
	}
}

// WithPCIDevicesRoot provides an Option to set the root path
// for PCI devices on the system.
func WithPCIDevicesRoot(root string) Option {
	return func(n *nvpci) {
		n.pciDevicesRoot = root
	}
}

// WithPCIDatabasePath provides an Option to set the path
// to the pciids database file.
func WithPCIDatabasePath(path string) Option {
	return func(n *nvpci) {
		n.pcidbPath = path
	}
}
```
that `nvpci.New()` accepts is required. I suppose since we're always running in a container we don't expect a different root for pci devices or a different pcidb path.

api/config/v1/flags.go
elezar: We have similar logic in the device plugin here: https://github.com/NVIDIA/k8s-device-plugin/blob/cf082174cda71b023030df58a6639572aaaea040/internal/plugin/manager/factory.go#L97-L133

This also changes the way resources are disovered based on autodetecting the platform (nvml, and tegra). I would assume that we would do something similar to allow the device plugin to also serve vfio devices?

elezar:
```suggestion
	Mode            *string   `json:"mode"            yaml:"mode"`
```

cmd/gpu-feature-discovery/main_test.go
elezar: How did you fix this? I would have expected this to change to `ptr("auto")` instead.

visheshtanksale: Apologies I missed this. Fixed it now.

internal/resource/sysfs-lib.go
visheshtanksale: Yes we dont expect both of the paths to be different, so not setting them.

api/config/v1/flags.go
```diff
@@ -63,6 +63,7 @@ type CommandLineFlags struct {
 	MOFEDEnabled      *bool                   `json:"mofedEnabled"               yaml:"mofedEnabled"`
 	UseNodeFeatureAPI *bool                   `json:"useNodeFeatureAPI"          yaml:"useNodeFeatureAPI"`
 	Plugin            *PluginCommandLineFlags `json:"plugin,omitempty"           yaml:"plugin,omitempty"`
+	Mode              *string                 `json:"mode"                       yaml:"mode"`
```
elezar, line 66: Nit: Let's add this above `Plugin` to so that the plugin and gfd-specific flags are last.

internal/resource/factory.go
```diff
@@ -39,7 +39,30 @@ func WithConfig(manager Manager, config *spec.Config) Manager {
...
+		return NewNVMLManager()
+	case "tegra":
+		klog.Info("Using CUDA manager")
+		return NewCudaManager()
+	case "vfio":
+		klog.Info("Using Vfio manager")
+		return NewVfioManager()
+	}
+
+	klog.Warning("No valid resources detected; using empty manager.")
```
elezar, line 57: nit: We may want to include `resolved` here:
```suggestion
	klog.Warningf("Unsupported mode detected: %v; using empty manager.", resolved)
```

internal/resource/sysfs-lib.go
```diff
@@ -0,0 +1,75 @@
+/**
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```
elezar, line 2: nit: 2022 -> 2024

cmd/gpu-feature-discovery/main.go
```diff
@@ -99,6 +100,12 @@ func main() {
 			Usage:   "Use NFD NodeFeature API to publish labels",
 			EnvVars: []string{"GFD_USE_NODE_FEATURE_API", "USE_NODE_FEATURE_API"},
 		},
+		&cli.StringFlag{
+			Name:    "mode",
+			Value:   "auto",
+			Usage:   "Select GFD mode between 'auto','nvml','tegra' or 'vfio'",
+			EnvVars: []string{"GFD_MODE"},
```
elezar, line 107:
```suggestion
			EnvVars: []string{"MODE", "GFD_MODE"},
```
or even just `MODE`?

elezar (changes requested):
Thanks @visheshtanksale. I think this looks a lot better now. A minor comment on the ordering of the config structs. Once that's been addressed, I think we can move ahead.

internal/resource/factory.go
visheshtanksale: Fixed

internal/resource/sysfs-lib.go
visheshtanksale: Fixed

cmd/gpu-feature-discovery/main.go
visheshtanksale: Fixed

api/config/v1/flags.go
visheshtanksale: Done

ArangoGutierrez (approved):

visheshtanksale:
> PR looks in good shape now, but could be squash the commits, there are 6 commits with the same title, and all for review comments, also edit PR description to provide some context, not a JIRA link.
> 
> Thanks

@ArangoGutierrez Squashed the commits and edited PR description.

ArangoGutierrez (approved):

deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -35,6 +35,7 @@ deviceIDStrategy: null
 nvidiaDriverRoot: null
 gdsEnabled: null
 mofedEnabled: null
+gfdMode: "auto"
```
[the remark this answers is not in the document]
elezar: Created #710

PR #687, opened 2024-04-30
Add release helper scripts 

ArangoGutierrez:

PR #691, opened 2024-05-03
GitHub Actions: Add Release workflow

ArangoGutierrez:


ArangoGutierrez:
Depends on #687

.github/workflows/release.yaml
```diff
@@ -0,0 +1,44 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+        run: |
+          GH_EXTRA_ARGS=""
+          if [[ ${{ github.ref }} == *rc* ]]; then
+            GH_EXTRA_ARGS="--prerelease"
+          fi
+          gh release create ${{ github.ref }} --draft \
```
elezar, line 39:
```suggestion
          gh release create ${{ github.ref }} \
              --draft \
```

elezar (commented):
We also spoke about uploading the helm packages and then publishing those to gh-pages. Is that still a goal for this PR?

.github/workflows/release.yaml
ArangoGutierrez: Can be this PR, a separate commit. will add it here. don't know why I was thinking on a followup PR

hack/update-helm-index.sh
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o pipefail
+
+asset_urls="$@"
+
+git checkout gh-pages
```
elezar, line 21: Check https://github.com/NVIDIA/k8s-device-plugin/pull/688/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R303-R308 where I checkout a local copy of the repo in some other folder to avoid conflicts with the branch that is currently checked out.

hack/update-helm-index.sh
```diff
@@ -0,0 +1,63 @@
...
+    download_baseurl=`dirname $asset_url`
+done
+
+if [ -z "$download_baseurl" ]; then
+    echo "No Helm chart release assets found"
+    exit 0
+fi
+
+echo "Updating helm index"
+helm repo index stable --merge index.yaml --url https://nvidia.github.io/k8s-device-plugin/stable
```
elezar, line 43: Is this what the `build-index.sh` script in `gh-pages` does?

hack/update-helm-index.sh
```diff
@@ -0,0 +1,63 @@
...
+cp -f stable/index.yaml index.yaml
+
+# Check if there were any changes in the repo
+if [ -z "`git status --short`" ]; then
+    echo "No changes in Helm repo incex, gh-pages branch already up-to-date"
+    exit 0
+fi
+
+# Create a new commit
+commit_msg="Update Helm repo index for release `basename $download_baseurl`
```
elezar, line 56: Should we specify the version explicitly?

hack/update-helm-index.sh
```diff
@@ -0,0 +1,63 @@
...
+echo "Updating helm index"
+helm repo index stable --merge index.yaml --url https://nvidia.github.io/k8s-device-plugin/stable
+
+popd > /dev/null
+
+cp -f stable/index.yaml index.yaml
+
+# Check if there were any changes in the repo
+if [ -z "`git status --short`" ]; then
+    echo "No changes in Helm repo incex, gh-pages branch already up-to-date"
```
elezar, line 51:
```suggestion
    echo "No changes in Helm repo index, gh-pages branch already up-to-date"
```

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,44 @@
...
+# Run this workflow on new tags
+name: Helm
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  update-helm-charts:
+    name: Update gh-pages helm repo index
```
elezar, line 24: This also publishes the packages to `gh-pages`, correct?

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,44 @@
...
+        fetch-depth: 0
+
+    - name: Update repo index
+      run: |
+        git config user.name "Github Actions"
+        git config user.email "no-reply@github.com"
+        ./hack/update-helm-index.sh  ${{ join(github.event.release.assets.*.browser_download_url, ' ') }}
+
+    - name: Push
+      run: git push -f https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }} gh-pages
```
elezar, line 44: Can this not leak the token? What is the recommended way to push to a branch in a github action?

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Check out repo
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Update repo index
+      run: |
+        git config user.name "Github Actions"
+        git config user.email "no-reply@github.com"
+        ./hack/update-helm-index.sh  ${{ join(github.event.release.assets.*.browser_download_url, ' ') }}
```
tariq1890, line 41: Can you move the update-helm-index.sh command argument to a variable so that it's more readable?

.github/workflows/helm.yaml
tariq1890: Let's move this CLI argument to a variable as well

.github/workflows/helm.yaml
tariq1890: Can we run these scripts as makefile targets instead?

hack/update-helm-index.sh
ArangoGutierrez: no, `build-index.sh` is

```bash
#!/usr/bin/env bash

helm repo index stable --url https://nvidia.github.io/k8s-device-plugin/stable

cp -f stable/index.yaml index.yaml
```

ArangoGutierrez: `--merge string   merge the generated index into the given index`

.github/workflows/helm.yaml
ArangoGutierrez: When handling GH_SECRETS is better to leave GitHub parser to handle the env vars, I prefer leaving it like this.

.github/workflows/helm.yaml
ArangoGutierrez: Since I want to take arguments, and let GitHub to build the argument from secrets, I prefer to leave it as a script

hack/update-helm-index.sh
ArangoGutierrez: Done

.github/workflows/helm.yaml
ArangoGutierrez: Yes, this action will push to `gh-pages` branch, which will trigger the action there.

hack/update-helm-index.sh
ArangoGutierrez: done

hack/update-helm-index.sh
ArangoGutierrez: done

hack/update-helm-index.sh
```diff
@@ -0,0 +1,66 @@
...
+
+pushd $HELM_REPO_PATH > /dev/null
+# Check if there were any changes in the repo
+if [ -z "`git status --short`" ]; then
+    echo "No changes in Helm repo index, gh-pages branch already up-to-date"
+    exit 0
+fi
+
+# Create a new commit
+commit_msg="Update Helm repo index for release $VERSION"
```
elezar, line 60: We're doing more than this. We're also adding the actual packages.

hack/update-helm-index.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Committing changes..."
+git commit -am "$commit_msg"
```
elezar, line 63: Does it make sense to explicitly add the files we're adding and then use `git add -u` here instead?

I would not use `git commit -a` in automated tooling.

hack/update-helm-index.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Updating helm index"
+helm repo index stable --merge index.yaml --url https://nvidia.github.io/k8s-device-plugin/stable
+popd > /dev/null
+
+cp -f $HELM_REPO_PATH/stable/index.yaml $HELM_REPO_PATH/index.yaml
+
+pushd $HELM_REPO_PATH > /dev/null
+# Check if there were any changes in the repo
+if [ -z "`git status --short`" ]; then
```
elezar, line 54: Does it make sense to use `$()` instead of the backticks?

.github/workflows/helm.yaml
ArangoGutierrez: ah sure, will add

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Push updated index to gh-pages branch
+      env:
+        HELM_REPO_PATH: releases/helm-${{ github.event.release.tag_name }}/
+      run: |
+        cd $HELM_REPO_PATH
+        git push -f https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }} gh-pages
```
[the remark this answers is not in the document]
ArangoGutierrez: no, `gh` is for GitHub interactions, not `git` operations

hack/update-helm-index.sh
ArangoGutierrez: ideally, we could remove `build-index.sh` @ `gh-pages` once this PR is in.

hack/update-helm-index.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ -z "$download_baseurl" ]; then
+    echo "No Helm chart release assets found"
+    exit 0
+fi
+
+echo "Updating helm index"
+helm repo index stable --merge index.yaml --url https://nvidia.github.io/k8s-device-plugin/stable
+popd > /dev/null
+
+cp -f $HELM_REPO_PATH/stable/index.yaml $HELM_REPO_PATH/index.yaml
```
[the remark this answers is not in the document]
ArangoGutierrez: No hard reason, it will simply change the absolute path, so it could be run before or after the `popd`

hack/update-helm-index.sh
ArangoGutierrez: ah sure, will address

hack/update-helm-index.sh
ArangoGutierrez: will enhance

hack/update-helm-index.sh
ArangoGutierrez: ok, will split

ArangoGutierrez:
@elezar PTAL

hack/update-helm-index.sh
```diff
@@ -0,0 +1,71 @@
...
+    if ! echo "$asset_url" | grep -q '.*tgz$'; then
+        echo "Skipping $asset_url, does not look like a Helm chart archive"
+        continue
+    fi
+
+    echo "Downloading $asset_url..."
+    curl -sSfLO $asset_url
+    # We rely on all release assets having the same baseurl
+    download_baseurl=`dirname $asset_url`
+done
```
elezar, line 39: QUestion: Does it make sense to also allow this to be triggered outside of github. That is to say to separate the download of the assets from the update of the index and publishing them?

hack/update-helm-index.sh
elezar: My question is shouldn't we keep the functionality but move it to main?

hack/update-helm-index.sh
ArangoGutierrez: Sure, makes total sense

ArangoGutierrez: Done PTAL

hack/update-helm-index.sh
```diff
@@ -0,0 +1,127 @@
...
+
+#
+# Parse command line
+#
+asset_url=
+asset_local=
+while [[ $# -gt 0 ]]; do
+	key="$1"
+	case $key in
+		--from-url)
```
elezar, line 50: Could we check for the `https://` | `http://` prefix instead of requiring that a user set this explicitly?

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,52 @@
...
+      en:
+        VERSION: ${{ github.event.release.tag_name }}
+        HELM_REPO_PATH: releases/helm-${{ github.event.release.tag_name }}/
+        DOWNLOAD_URL: ${{ join(github.event.release.assets.*.browser_download_url, ' ') }}
+      run: |
+        git config user.name "Github Actions"
+        git config user.email "no-reply@github.com"
+        ./hack/update-helm-index.sh $DOWNLOAD_URL
+
+    - name: Push updated index to gh-pages branch
```
elezar, line 47: Nit: it also pushes the packages.

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,52 @@
...
+      env:
+        HELM_REPO_PATH: releases/helm-${{ github.event.release.tag_name }}/
+      run: |
+        cd $HELM_REPO_PATH
+        git push -f https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }} gh-pages
```
elezar, line 52: Let's use `git -C $HELM_REPO_PATH` instead.

Also, let's not `push -f`. If something goes wrong here we don't want to overwrite the repo contents.

I also find it strange that `gh-pages` is implied in the previous step, but explicitly listed here.

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Check out repo
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Update helm index
+      en:
+        VERSION: ${{ github.event.release.tag_name }}
+        HELM_REPO_PATH: releases/helm-${{ github.event.release.tag_name }}/
```
elezar, line 40: Should the `HELM_REPO_PATH` be an argument?

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        version: 3.14.4
+
+    - name: Check out repo
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Update helm index
+      en:
```
elezar, line 38:
```suggestion
      env:
```

hack/update-helm-index.sh
```diff
@@ -0,0 +1,127 @@
...
+# Create a new commit
+commit_msg="Update helm charts for release $VERSION
+
+This commit was created by the k8s-device-plugin release workflow.
+It updates the Helm charts in the gh-pages branch to include the $VERSION release assets.
+Also updates the Helm index.yaml file to include the new chart versions."
+
+echo "Committing changes..."
+git add index.yaml stable
+git commit -m "$commit_msg"
```
elezar, line 124: Do we need to add `-s`?

hack/update-helm-index.sh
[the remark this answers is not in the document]
elezar: Let's keep it with the other one since they're both concerned with updating the `index.yaml`.

elezar (changes requested):
Thanks for iterating on this.

Please also update `RELEASE.md` to reflect the changes here.

hack/update-helm-index.sh
ArangoGutierrez: no... use of habit will remove

hack/update-helm-index.sh
ArangoGutierrez: sure, I was not really sure how to make this friendly and defaulted for flags, but we can make it nicer, thanks for the recommendation

hack/update-helm-index.sh
elezar: I mean do we want to sign-off the commits, not sign the commits.

.github/workflows/helm.yaml
ArangoGutierrez: is not implied in the previous step, the first step at  `./hack/update-helm-index.sh` is to git clone and git checkout to `gh-pages` branch that's why is not visible at the GitHub action script

.github/workflows/helm.yaml
ArangoGutierrez: thanks, good catch

.github/workflows/helm.yaml
ArangoGutierrez: done

.github/workflows/helm.yaml
ArangoGutierrez: done

.github/workflows/helm.yaml
ArangoGutierrez: set as an optional argument

hack/update-helm-index.sh
ArangoGutierrez: they will be signed off with `no-reply@github.com`, done

RELEASE.md
```diff
@@ -13,8 +13,6 @@ Publishing the helm chart is currently manual, and we should move to an automate
 - [ ] Run the `./hack/generate-changelog.sh` script to generate the a draft changelog
 - [ ] Update the CHANGELOG.md file with the generated changelog
 - [ ] Commit, Tag and Push to Gitlab
-- [ ] Switch to the `gh-pages` branch and move the newly generated package to the `stable` helm repo
-- [ ] While on the `gh-pages` branch, run the `./build-index.sh` script to rebuild the indices for each repo
-- [ ] Commit and push the `gh-pages` branch to GitHub
-- [ ] Wait for the [CI job associated with your tag] (https://gitlab.com/nvidia/kubernetes/device-plugin/-/pipelines) to complete
-- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) on Github with the changelog
+- [ ] Wait for the `Release` GitHub Action to complete
+- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases)from the draft release created by the GitHub Action
```
elezar, line 17: Do we want to document how to do this if things go wrong?

hack/update-helm-index.sh
```diff
@@ -0,0 +1,130 @@
...
+  --helm-repo-path      specify the path to the Helm repo (defaults to HELM_REPO_PATH)
+  --help/-h             show this help and exit
+
+NOTE: only one of --from-url or --from-local can be specified
+
+Example:
+  - from a local path:
+	$this --from-local /path/to/nvidia-device-plugin-0.15.0.tgz
+  - from a URL:	
+    $this --from-url https://github.com/NVIDIA/k8s-device-plugin/archive/refs/tags/nvidia-device-plugin-0.15.0.tgz
```
elezar, line 35: Out of date.

RELEASE.md
ArangoGutierrez: done! good idea

hack/update-helm-index.sh
ArangoGutierrez: fixed

.github/workflows/helm.yaml
ArangoGutierrez: done

hack/update-helm-index.sh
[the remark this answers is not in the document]
ArangoGutierrez: done

.github/workflows/helm.yaml
ArangoGutierrez: done

ArangoGutierrez:
@elezar / @tariq1890  PTAL

.github/workflows/helm.yaml
elezar: Let's make it a required argument and not use the envvar.

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,51 @@
...
+        fetch-depth: 0
+
+    - name: Update helm index
+      env:
+        VERSION: ${{ github.event.release.tag_name }}
+        HELM_REPO_PATH: releases/helm-${{ github.event.release.tag_name }}/
+        DOWNLOAD_URL: ${{ join(github.event.release.assets.*.browser_download_url, ' ') }}
+      run: |
+        git config user.name "Github Actions"
+        git config user.email "no-reply@github.com"
```
elezar, line 44: Question (not a blocker): can this be set in the action somehow?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+          if [[ ${{ github.ref }} == *rc* ]]; then
+            GH_EXTRA_ARGS="--prerelease"
+          fi
+          gh release create ${{ github.ref }} \
+            --draft \
+            -t "${{ github.ref }}" \
+            -n $(./hack/generate-changelog.sh --version ${{ github.ref }}) \
+            -R $OWNER/$REPO \
+            --verify-tag \
+            $GH_EXTRA_ARGS
```
elezar, line 45: Is this something that we want to have in a script instead?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+          ./hack/package-helm-charts.sh ${{ github.event.release.tag_name }}
+
+      - name: Upload Release Artifacts
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+        run: |
+          gh release upload ${{ github.ref }} ./nvidia-device-plugin-${{ github.event.release.tag_name }}.tgz -R $OWNER/$REPO
+          gh release upload ${{ github.ref }} ./gpu-feature-discovery-${{ github.event.release.tag_name }}.tgz -R $OWNER/$REPO
```
elezar, line 63: also a script?

RELEASE.md
```diff
@@ -13,8 +13,14 @@ Publishing the helm chart is currently manual, and we should move to an automate
 - [ ] Run the `./hack/generate-changelog.sh` script to generate the a draft changelog
 - [ ] Update the CHANGELOG.md file with the generated changelog
 - [ ] Commit, Tag and Push to Gitlab
-- [ ] Switch to the `gh-pages` branch and move the newly generated package to the `stable` helm repo
-- [ ] While on the `gh-pages` branch, run the `./build-index.sh` script to rebuild the indices for each repo
-- [ ] Commit and push the `gh-pages` branch to GitHub
-- [ ] Wait for the [CI job associated with your tag] (https://gitlab.com/nvidia/kubernetes/device-plugin/-/pipelines) to complete
-- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) on Github with the changelog
+- [ ] Wait for the `Release` GitHub Action to complete
+- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) from the draft release created by the GitHub Action
```
elezar, line 17: Nit:
```suggestion
- [ ] Publish the [draft release](https://github.com/NVIDIA/k8s-device-plugin/releases) created by the GitHub Action
```

RELEASE.md
```diff
@@ -13,8 +13,14 @@ Publishing the helm chart is currently manual, and we should move to an automate
 - [ ] Run the `./hack/generate-changelog.sh` script to generate the a draft changelog
 - [ ] Update the CHANGELOG.md file with the generated changelog
 - [ ] Commit, Tag and Push to Gitlab
```
elezar, line 15: It seems as if some fo the steps were lost somewhere. We no longer mention the steps to publish the image. Could we readd those?

RELEASE.md
```diff
@@ -13,8 +13,14 @@ Publishing the helm chart is currently manual, and we should move to an automate
...
-- [ ] Wait for the [CI job associated with your tag] (https://gitlab.com/nvidia/kubernetes/device-plugin/-/pipelines) to complete
-- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) on Github with the changelog
+- [ ] Wait for the `Release` GitHub Action to complete
+- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) from the draft release created by the GitHub Action
+- [ ] Wait for the `Publish Helm Chart` GitHub Action to complete
+
+## Troubleshooting
+- If the `Release` GitHub Action fails:
+    - check the logs for the error first.
+    - Manually run `./hack/package-helm-charts.sh` and updload the generated charts to the release assets
```
elezar, line 23: Should we list the assets?

RELEASE.md
```diff
@@ -13,8 +13,14 @@ Publishing the helm chart is currently manual, and we should move to an automate
...
-- [ ] Commit and push the `gh-pages` branch to GitHub
-- [ ] Wait for the [CI job associated with your tag] (https://gitlab.com/nvidia/kubernetes/device-plugin/-/pipelines) to complete
-- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) on Github with the changelog
+- [ ] Wait for the `Release` GitHub Action to complete
+- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) from the draft release created by the GitHub Action
+- [ ] Wait for the `Publish Helm Chart` GitHub Action to complete
+
+## Troubleshooting
+- If the `Release` GitHub Action fails:
+    - check the logs for the error first.
```
elezar, line 22: Is rerunning the job an option?

hack/package-helm-charts.sh
```diff
@@ -16,6 +16,13 @@
 
 set -o pipefail
 
+# if arg1 is set, it will be used as the version number
+if [ -z "$1" ]; then
+  VERSION=$(git describe --tags --always --dirty)
+else
+  VERSION=$1
+fi
```
elezar, line 24: Is it simpler to make this a required argument? Alternatively should we extract the version from `versions.mk`?

hack/update-helm-index.sh
```diff
@@ -0,0 +1,128 @@
...
+pushd $HELM_REPO_PATH/stable > /dev/null
+# Download charts from release assets
+
+if [ -n "$asset_local" ]; then
+	echo "Copying $asset_local..."
+	cp -f $asset_local $HELM_REPO_PATH/stable
+fi
+
+if [ -n "$asset_url" ]; then
+	for asset_url in $asset_urls; do
```
elezar, line 86: Where is `asset_urls` defined?

hack/update-helm-index.sh
```diff
@@ -0,0 +1,128 @@
...
+	    if ! echo "$asset_url" | grep -q '.*tgz$'; then
+	        echo "Skipping $asset_url, does not look like a Helm chart archive"
+	        continue
+	    fi
+
+	    echo "Downloading $asset_url..."
+	    curl -sSfLO $asset_url
```
elezar, line 93: Where do we download these to?

hack/update-helm-index.sh
```diff
@@ -0,0 +1,128 @@
...
+git -C $(HELM_REPO_PATH) fetch github gh-pages
+git -C $(HELM_REPO_PATH) checkout gh-pages
+
+pushd $HELM_REPO_PATH/stable > /dev/null
+# Download charts from release assets
+
+if [ -n "$asset_local" ]; then
+	echo "Copying $asset_local..."
+	cp -f $asset_local $HELM_REPO_PATH/stable
+fi
```
elezar, line 83: I don't think this works the way we want. We are in `$HELM_REPO_PATH/stable` meaning that `$asset_local` would be interpreted relative to that path.

hack/update-helm-index.sh
```diff
@@ -0,0 +1,128 @@
...
+	done
+
+	if [ -z "$download_baseurl" ]; then
+	    echo "No Helm chart release assets found"
+	    exit 0
+	fi
+fi
+
+echo "Updating helm index"
+helm repo index stable --merge index.yaml --url https://nvidia.github.io/k8s-device-plugin/stable
```
elezar, line 105: We're in `$HELM_REPO_PATH/stable`. Does that mean that this should be:
```
```suggestion
helm repo index . --merge index.yaml --url https://nvidia.github.io/k8s-device-plugin/stable
```

hack/update-helm-index.sh
```diff
@@ -0,0 +1,128 @@
...
+cp -f stable/index.yaml index.yaml
```
elezar, line 106: Since we're in `$HELM_REPO_PATH/stable` this command is invalid. Should it be:
```suggestion
cp -f index.yaml ../index.yaml
```

hack/update-helm-index.sh
```diff
@@ -0,0 +1,128 @@
...
+popd > /dev/null
+
+pushd $HELM_REPO_PATH > /dev/null
```
elezar, line 109: Should we use `git -C` for the git commands instead of `pushd`/`popd`?

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,51 @@
...
+        ./hack/update-helm-index.sh $DOWNLOAD_URL
+
+    - name: Push updated Helm charts and index to gh-pages branch
+      env:
+        HELM_REPO_PATH: releases/helm-${{ github.event.release.tag_name }}/
```
elezar, line 49: Question: Could we set `HELM_REPO_PATH` as a job-level envvar?

elezar (changes requested):
I think there are some finer points in the implementation that still need to be looked at.

hack/update-helm-index.sh
ArangoGutierrez: Line 66

elezar: No, `asset_url` is set there, not `assert_urls`:
```
if [[ $asset_path =~ ^https?:// ]]; then
	asset_url=$asset_path
else
```

hack/update-helm-index.sh
ArangoGutierrez: `HELM_REPO_PATH` is the absolute path, so it doesn't matter where we are, the `cp` will perform the copy to the given path.

RELEASE.md
ArangoGutierrez: is just 2 files with `tgz` sufix. I am not against it. up to you

.github/workflows/release.yaml
ArangoGutierrez: ummm I think this one and the next one(same comment) are good to have here since they are not functionality we would do manually. and is good to have GitHub actions that are readable. 
So I would say no, I prefer to keep this as is

.github/workflows/helm.yaml
ArangoGutierrez: you mean like as an argument to the action? 
The documentation and examples I have found all play it like this.

.github/workflows/release.yaml
ArangoGutierrez: answer on the comment above

RELEASE.md
ArangoGutierrez: I don't follow, current https://raw.githubusercontent.com/NVIDIA/k8s-device-plugin/main/RELEASE.md RELEASE doesn't hold steps for image release. ALl we have is line 7
```md
Publishing the container is automated through gitlab-ci and only requires one to tag the commit and push it to gitlab.
```

hack/update-helm-index.sh
elezar: What if we have to run these locally as in the "troubleshooting" case?

RELEASE.md
elezar: Sorry, I think I'm misremembering the steps we outlined in github issues / JIRA and the steps in this file. (see for example https://github.com/NVIDIA/cloud-native-team/issues/59).

The `Commit, Tag and Push` to Gitlab is the image trigger though, so we may want to reword this.

Not a blocker for this PR though.

ArangoGutierrez:
Updated as discussed @elezar

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Run this workflow on new tags
+name: Publish Helm Chart
+
+on:
+  release:
+    types: [published]
```
tariq1890, line 20: Let's be consistent in how we specify yaml arrays
```suggestion
    types: 
      - published
```

hack/update-helm-index.sh
```diff
@@ -0,0 +1,120 @@
...
+
+this=`basename $0`
+
+usage () {
+cat << EOF
+Usage: $this [-h]
+
+Options:
+  --helm-repo-path      specify the path to the Helm repo (defaults to HELM_REPO_PATH)
+  --help/-h             show this help and exit
```
tariq1890, line 27:
```suggestion
  --help/-h             show help for this command and exit
```

RELEASE.md
```diff
@@ -5,16 +5,22 @@ The device plugin consists in two artifacts:
...
-- [ ] Create a [new release](https://github.com/NVIDIA/k8s-device-plugin/releases) on Github with the changelog
+- [ ] Wait for the `Release` GitHub Action to complete
+- [ ] Publish the [draft release](https://github.com/NVIDIA/k8s-device-plugin/releases) created by the GitHub Action
+- [ ] Wait for the `Publish Helm Chart` GitHub Action to complete
+
+## Troubleshooting
+- If the `Release` GitHub Action fails:
+    - check the logs for the error first.
+    - Manually run the `Release` GitHub Action locally with the same inputs.
+    - If the action continues to fail, manually run `./hack/package-helm-charts.sh` and updload the generated charts to the release assets.
```
tariq1890, line 23:
```suggestion
    - If the action fails, manually run `./hack/package-helm-charts.sh` and upload the generated charts to the release assets.
```

RELEASE.md
```diff
@@ -5,16 +5,22 @@ The device plugin consists in two artifacts:
...
+- If the `Publish Helm Chart` GitHub Action fails:
+    - check the logs for the error first.
```
tariq1890, line 25:
```suggestion
    - Check the logs for the error.
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - uses: actions/checkout@v4
+        name: Check out code
+      - name: Create Draft Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+        run: |
+          GH_EXTRA_ARGS=""
+          if [[ ${{ github.ref }} == *rc* ]]; then
```
elezar, line 36: strictly-speaking this should probably be `v*-rc.*` to reduce false positives.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        name: Check out code
+      - name: Create Draft Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+        run: |
+          GH_EXTRA_ARGS=""
```
elezar, line 35: This does not work as expected. See https://github.com/NVIDIA/nvidia-container-toolkit/actions/runs/9549456867/job/26319106625

Also when running the gh command line as given, it asks a number of questions. Not sure whether that would be skipped in the action.

ArangoGutierrez: Fixed with `github.ref_name`

ArangoGutierrez: It should not ask questions, as it should know is running on GItHub Actions mode https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          ./hack/package-helm-charts.sh ${{ github.event.release.tag_name }}
+
+      - name: Upload Release Artifacts
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+        run: |
+          gh release upload ${{ github.ref_name	}} ./nvidia-device-plugin-${{ github.event.release.tag_name }}.tgz -R $OWNER/$REPO
```
elezar, line 62: What value does `tag_name` have? Do we strip the leading `v`?

ArangoGutierrez: responds to `refs/tags/<tag_name>` so it will include the `v`

elezar: Ok. Our packages don't include the v though. Can we update this?

Also, as a matter of interest, why do we use `tag_name` instead of just using `ref_name` consistently?

ArangoGutierrez: how about now?

elezar (approved):
I have made the required changes to the branch.

PR #699, opened 2024-05-07
Allow /dev/shm size to be specified

elezar:
This change allows the size of the /dev/shm created for MPS to be specified as part of the helm deployment.

### DIFF
--- a/cmd/mps-control-daemon/mount/mount-shm.go
+++ b/cmd/mps-control-daemon/mount/mount-shm.go
@@ -20,23 +20,55 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"regexp"
 
 	"github.com/urfave/cli/v2"
 	"k8s.io/mount-utils"
 )
 
+type options struct {
+	shmSize string
+}
+
 // NewCommand constructs a mount command.
 func NewCommand() *cli.Command {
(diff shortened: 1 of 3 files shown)


cmd/mps-control-daemon/mount/mount-shm.go
```diff
@@ -20,23 +20,55 @@ import (
...
+
+	c := cli.Command{
+		Name:  "mount-shm",
+		Usage: "Set up the /dev/shm mount required by the MPS daemon",
+		Before: func(ctx *cli.Context) error {
+			return validateFlags(ctx, &opts)
+		},
+		Action: func(ctx *cli.Context) error {
+			return mountShm(ctx, &opts)
+		},
```
klueska, line 45: I has assumed we would do this as a configuration parameter in the config file

elezar: Yes, I had that in mind too and was looking into that as a follow up to this commit.

Would you ONLY expose it in the config file?

klueska: I think we definitely want it in the config -- I don't have a strong opinion about whether it is *also* a command line flag. Probably not though, since other MPS options aren't.

klueska (changes requested):

PR #710, opened 2024-05-16
Use gfd.mode instead of gfdMode

elezar:
See https://github.com/NVIDIA/k8s-device-plugin/pull/676#discussion_r1602916576 from #676 

In #709 we add a `devicePlugin.mode`.


elezar:
Superseded by #709

PR #713, opened 2024-05-16
Raise error in GFD when MPS sharing is enabled for MIG devices

elezar:
This change raises an error in GFD if MPS sharing is enabled for MIG devices instead of (siliently) setting `nvidia.com/mps.capable=false`.

This aligns GFD behaviour with that of the device plugin.

### DIFF
--- a/internal/lm/nvml.go
+++ b/internal/lm/nvml.go
@@ -17,15 +17,16 @@
 package lm
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
-	"k8s.io/klog/v2"
-
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 	"github.com/NVIDIA/k8s-device-plugin/internal/resource"
 )
+
+var errMPSSharingNotSupported = errors.New("MPS sharing is not supported")
 
 // NewDeviceLabeler creates a new labeler for the specified resource manager.
 func NewDeviceLabeler(manager resource.Manager, config *spec.Config) (Labeler, error) {
@@ -60,6 +61,11 @@ func NewDeviceLabeler(manager resource.Manager, config *spec.Config) (Labeler, e
 		return nil, fmt.Errorf("error creating mig capability labeler: %v", err)
 	}
 
+	sharingLabeler, err := newSharingLabeler(manager, config)
+	if err != nil {
+		return nil, fmt.Errorf("error creating sharing labeler: %w", err)
+	}
+
 	resourceLabeler, err := NewResourceLabeler(manager, config)
 	if err != nil {
 		return nil, fmt.Errorf("error creating resource labeler: %v", err)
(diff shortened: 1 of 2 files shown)


internal/lm/resource.go
```diff
@@ -90,6 +91,10 @@ func NewMIGResourceLabeler(resourceName spec.ResourceName, config *spec.Config,
 	}
 
 	resourceLabeler := newResourceLabeler(resourceName, config)
+	if strategy := resourceLabeler.sharing.SharingStrategy(); strategy == spec.SharingStrategyMPS {
+		klog.Warningf("Ignoring unsupported sharing strategy for MIG devices: %v", strategy)
```
elezar, line 95: Question: instead of issuing a warning and removing the setting, should we error out GFD instead?

klueska:
Instead of doing this, do we want to add support for MIG devices with MPS?

elezar:
> Instead of doing this, do we want to add support for MIG devices with MPS?

Note that the code change here is relatively straightforward and just ensures that we don't label a node a `mps.capable` or adjust sharing labels when MIG is enabled. This is simple to revert once we do support MPS with MIG.

klueska:
> Note that the code change here is relatively straightforward and just ensures that we don't label a node a `mps.capable` or adjust sharing labels when MIG is enabled. This is simple to revert once we do support MPS with MIG.

That's fair. Let's do that then (I assume the tests will need to be reverted at that point too).

internal/lm/resource.go
elezar: The plugin raises an error.

klueska: What would be the reason to not just do the same here then?

elezar: This does mean that no other labels are generated. In the case where a user is just deploying GFD -- possibly with a faulty config -- we would end up with no labels at all.

klueska: You mean running GFD alongside a different device plugin that doesn't support our config file? I still think we would want to "fail-fast" in this case -- prompting an admin to check the logs and update the config file. Unless I am missing some case where updating the config wouldn't be possible, but we want to allow them to continue anyway.

internal/lm/nvml.go
```diff
@@ -151,26 +155,27 @@ func newMigCapabilityLabeler(manager resource.Manager) (Labeler, error) {
...
 
-	return Labels{
+	labels := Labels{
 		"nvidia.com/mps.capable": strconv.FormatBool(capable),
 	}
+	return labels, nil
 }
 
+var errMPSSharingNotSupported = errors.New("MPS sharing is not supported")
+
```
klueska, line 178: Can we put this at the top of the file instead of embedded here in the middle of the file?

elezar: Sure. I had it close to where it was used, but it makes more sense further up.

klueska (approved):

PR #730, opened 2024-05-27
Make construction of infolib, nvmllib, and devicelib explicit

elezar:
This change ensures the infolib, nvmllib, and devicelib instances
are constructed once and passed by reference to functions that
need them.

### DIFF
--- a/cmd/gpu-feature-discovery/main.go
+++ b/cmd/gpu-feature-discovery/main.go
@@ -14,6 +14,10 @@ import (
 	"github.com/urfave/cli/v2"
 	"k8s.io/klog/v2"
 
+	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
+	nvinfo "github.com/NVIDIA/go-nvlib/pkg/nvlib/info"
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 	"github.com/NVIDIA/k8s-device-plugin/internal/flags"
 	"github.com/NVIDIA/k8s-device-plugin/internal/info"
@@ -166,7 +170,14 @@ func start(c *cli.Context, cfg *Config) error {
 		}
 		klog.Infof("\nRunning with config:\n%v", string(configJSON))
 
-		manager := resource.NewManager(config)
+		nvmllib := nvml.New()
+		devicelib := device.New(nvmllib)
+		infolib := nvinfo.New(
+			nvinfo.WithNvmlLib(nvmllib),
+			nvinfo.WithDeviceLib(devicelib),
+		)
+
+		manager := resource.NewManager(infolib, nvmllib, devicelib, config)
 		vgpul := vgpu.NewVGPULib(vgpu.NewNvidiaPCILib())
 
 		var clientSets flags.ClientSets
--- a/cmd/mps-control-daemon/main.go
+++ b/cmd/mps-control-daemon/main.go
@@ -27,6 +27,10 @@ import (
 	"github.com/urfave/cli/v2"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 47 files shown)


ArangoGutierrez (approved):

cmd/gpu-feature-discovery/main.go
```diff
@@ -166,7 +170,15 @@ func start(c *cli.Context, cfg *Config) error {
 		}
 		klog.Infof("\nRunning with config:\n%v", string(configJSON))
 
-		manager := resource.NewManager(config)
+		// TODO: If applicable, an nvmllib needs to be passed to the infolib constructor.
```
klueska, line 173: What do you mean by "if applicable"? When is this necessary vs. not?

cmd/gpu-feature-discovery/main.go
```diff
@@ -166,7 +170,15 @@ func start(c *cli.Context, cfg *Config) error {
...
+		nvmllib := nvml.New()
+		devicelib := device.New(device.WithNvml(nvmllib))
```
klueska, line 175: Didn't we change this recently to a required parameter (as opposed to an optional one via `WithNvml()`).

cmd/mps-control-daemon/main.go
```diff
@@ -168,9 +172,16 @@ func startDaemons(c *cli.Context, cfg *Config) ([]*mps.Daemon, bool, error) {
 	}
 	spec.DisableResourceNamingInConfig(logger.ToKlog, config)
 
+	nvmllib := nvml.New()
+	devicelib := device.New(device.WithNvml(nvmllib))
+	infolib := nvinfo.New(
+		nvinfo.WithNvmlLib(nvmllib),
+		nvinfo.WithDeviceLib(devicelib),
+	)
```
klueska, line 180: Same questions as above apply here.

cmd/mps-control-daemon/main.go
```diff
@@ -168,9 +172,16 @@ func startDaemons(c *cli.Context, cfg *Config) ([]*mps.Daemon, bool, error) {
...
+
 	// Update the configuration file with default resources.
 	klog.Info("Updating config with default resource matching patterns.")
-	err = rm.AddDefaultResourcesToConfig(config)
+	err = rm.AddDefaultResourcesToConfig(infolib, nvmllib, devicelib, config)
```
klueska, line 184: Does it make sense to introduce a new struct to hold all of these instead of as individual parameters?

cmd/gpu-feature-discovery/main.go
elezar: I think this comment is not relevant anymore. This was an earlier itteration.

cmd/gpu-feature-discovery/main.go
elezar: Yes. I have not yet included that change. I will pull it in.

cmd/mps-control-daemon/main.go
elezar: I think I would like to do this as a follow-up if you're ok with that.

cmd/gpu-feature-discovery/main.go
elezar: Removed the comment.

cmd/nvidia-device-plugin/main.go
```diff
@@ -262,9 +260,21 @@ func startPlugins(c *cli.Context, flags []cli.Flag) ([]plugin.Interface, bool, e
 	}
 	spec.DisableResourceNamingInConfig(logger.ToKlog, config)
 
+	nvmllib := nvml.New()
+	devicelib := device.New(device.WithNvml(nvmllib))
+	infolib := nvinfo.New(
+		nvinfo.WithNvmlLib(nvmllib),
+		nvinfo.WithDeviceLib(devicelib),
+	)
```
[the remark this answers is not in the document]
elezar: Updated.

PR #747, opened 2024-05-30
add ngc image signing job for auto signing

shivakunv:
add ngc image signing job for auto signing

### DIFF
--- a/.common-ci.yml
+++ b/.common-ci.yml
@@ -28,6 +28,7 @@ stages:
   - test
   - scan
   - release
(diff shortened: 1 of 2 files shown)


shivakunv:
@cdesiniotis @elezar PTAL

elezar (changes requested):
Comments from https://github.com/NVIDIA/nvidia-container-toolkit/pull/519 are applicable here too.

elezar (approved):

PR #748, opened 2024-05-31
Fix usage of hasConfigMap

### LINKED ISSUE #689: nvidia-device-plugin.hasConfigMap returns a string
The helm chart helper `nvidia-device-plugin.hasConfigMap` attempts to return true/false but an `include` will always return a string.

Some places handled this properly by comparing to `true` while other places used the string as a guard.

### 3. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) 
https://github.com/NVIDIA/k8s-device-plugin/blob/main/deployments/helm/nvidia-device-plugin/templates/role.yml#L1
https://github.com/NVIDIA/k8s-device-plugin/blob/main/deployments/helm/nvidia-device-plugin/templates/service-account.yml#L1
https://github.com/NVIDIA/k8s-device-plugin/blob/main/deployments/helm/nvidia-device-plugin/templates/role-binding.yml#L1

elezar:
The helm include statement always generates a string. This was leading to inconsistent handling of the hasConfigMap option.

This change adds a options helper that can be used to pass derrived values as JSON so that the type information is retained when including these.

Fixes #689

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
+++ b/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
@@ -217,3 +217,15 @@ Pod annotations for the plugin and GFD
 {{- end -}}
 {{- toYaml $annotations }}
 {{- end -}}
+
+
+{{/*
+Collection of typed options for the device plugin.
+
+We convert this to JSON so that it can be included and converted to an object using fromJson.
+*/}}
+{{- define "nvidia-device-plugin.options" -}}
+{{- $options := dict "" "" -}}
+{{- $_ := set $options "hasConfigMap" ( eq ( (include "nvidia-device-plugin.hasConfigMap" . ) | trim ) "true" ) -}}
+{{- mustToJson $options -}}
+{{- end -}}
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -14,8 +14,8 @@
 
 {{- if .Values.devicePlugin.enabled }}
 ---
-{{- $hasConfigMap := (include "nvidia-device-plugin.hasConfigMap" .) | trim }}
-{{- $useServiceAccount := eq $hasConfigMap "true" }}
+{{- $options := (include "nvidia-device-plugin.options" . | fromJson) }}
+{{- $useServiceAccount := $options.hasConfigMap }}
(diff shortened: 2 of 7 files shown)


deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
```diff
@@ -217,3 +217,15 @@ Pod annotations for the plugin and GFD
...
 {{- toYaml $annotations }}
 {{- end -}}
+
+
+{{/*
+Collection of typed options for the device plugin.
+
+We convert this to JSON so that it can be included and converted to an object using fromJson.
+*/}}
+{{- define "nvidia-device-plugin.options" -}}
```
tariq1890, line 227: Why can't this function return the typed options object instead of the Json? It seems a bit redundant the current way since we seem to couple all the calls to `nvidia-device-plugins.options` with `fromJson`

elezar: I think it could and I will spend some time checking out how this should be done.

For what it's worth this is from: https://github.com/helm/helm/issues/11231#issuecomment-1234506859

tariq1890: Understood, thanks for the context

deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
```diff
@@ -217,3 +217,15 @@ Pod annotations for the plugin and GFD
...
+{{- $options := dict "" "" -}}
+{{- $_ := set $options "hasConfigMap" ( eq ( (include "nvidia-device-plugin.hasConfigMap" . ) | trim ) "true" ) -}}
+{{- mustToJson $options -}}
```
tariq1890, line 230:
```suggestion
{{- mustToJson $options -}}
{{- fromJson $options -}}
```

elezar: I don't think that works as expected. The point is that `options` here must be a JSON string. The reason being that when we include this template it is **always** rendered as a string. This is why we use `fromJson` at the call site to construct a typed object there.

tariq1890: ahh yes, makes sense. So part of the workaround is to pipe the output of the `include` call to `fromJson` at the point of invocation.

Unfortunate that we have to resort to this workaround, but I see there's no other way around it. I will resolve the comment

tariq1890 (approved):

PR #749, opened 2024-05-31
Fix all possible mig strategies

elezar:
This depends on #675 and #748


elezar:
Incorportated in #675

PR #756, opened 2024-06-03
Adding security context to initContainer

kndoni:
**Changes** 

1. Adding security context to init container as keyverno policies implies that initContainer should have securityContext defined as well
2. Adding resources for every container and initContainer on each daemonset.
3. Adding property devicePluginMps in values that controlles enabling or disabling mps daemonset.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -92,6 +92,8 @@ spec:
             mountPath: /available-configs
           - name: config
             mountPath: /config
+        securityContext:
+          {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
       {{- end }}
       containers:
       {{- if $options.hasConfigMap }}
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -194,6 +194,9 @@ spec:
             value: compute,utility
         securityContext:
           {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
+          capabilities:
+            add:
+              - SYS_ADMIN
         volumeMounts:
           - name: device-plugin
             mountPath: /var/lib/kubelet/device-plugins
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -92,6 +92,10 @@ spec:
             mountPath: /available-configs
           - name: config
             mountPath: /config
+        {{- with .Values.resources }}
+        resources:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
         securityContext:
           {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -92,6 +92,8 @@ spec:
             mountPath: /available-configs
           - name: config
             mountPath: /config
+        securityContext:
```
elezar, line 95: @klueska @cdesiniotis do you have any specifics on why this was omitted?

elezar (changes requested):
@kndoni please also add this to the other daemonsest in the helm package.

kndoni:
> @kndoni please also add this to the other daemonsest in the helm package.

I have added security context for other containers in daemonsets as well

elezar (approved):

deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
```diff
@@ -61,6 +61,7 @@ spec:
         command: [mps-control-daemon, mount-shm]
         securityContext:
           privileged: true
+          {{- include "mps-control-daemon.securityContext" . | nindent 10 }}
```
elezar, line 64: Note that this helper template is not defined.

deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
```diff
@@ -169,6 +172,7 @@ spec:
             value: compute,utility
           securityContext:
             privileged: true
+            {{- include "mps-control-daemon.securityContext" . | nindent 10 }}
```
elezar, line 175: Since we're already running the container as `privileged`, I don't think this is relevant here.

elezar (changes requested):
I don't think that adding a security context if we're already running a container as priveleved makes sense. We also do not have the template used for the mps-control-daemon defined.

deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
kndoni: Removed this as well

kndoni:
@elezar Let me know if something else should be changed as well.
Thanks in advance!

elezar (changes requested):
@kndoni please remove the changes to the MPS API from this PR.

I'm not opposed to making the change, but would like to separate the security-context specific changes from the API changes.

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -92,6 +92,12 @@ spec:
             mountPath: /available-configs
           - name: config
             mountPath: /config
+        {{- with .Values.resources }}
```
elezar, line 95: A general question: Does it make sense to apply the same resources to each of the containers?

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -184,6 +194,9 @@ spec:
             value: compute,utility
         securityContext:
           {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
+          capabilities:
+            add:
+              - SYS_ADMIN
```
elezar, line 199: Why was this added? There was an issue #856 which should be resolved in the latest release.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -175,7 +185,10 @@ spec:
           - name: GFD_MODE
             value: "{{ .Values.gfdMode }}"
         securityContext:
-          {{- include "gpu-feature-discovery.securityContext" . | nindent 10 }}
+          {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
+          capabilities:
+            add:
+              - SYS_ADMIN
```
elezar, line 191: Similar question here. See #856

PR #768, opened 2024-06-14
Add support for node labels to report GPU mode

visheshtanksale:
Some GPUs support switching between graphics mode and compute mode.

The mode is switched by a utility called [displaymodeselector](https://developer.nvidia.com/displaymodeselector)

This change identifies the mode on the GPU and labels the node to help end user schedule workloads

The assumption is that all the GPUs on the node have the same mode, if that is not the case then the label `nvidia.com/gpu.mode` value is `unknown`


internal/resource/cuda-device.go
```diff
@@ -96,3 +96,5 @@ func (d *cudaDevice) IsMigCapable() (bool, error) {
 func (d *cudaDevice) IsMigEnabled() (bool, error) {
 	return false, nil
 }
+
+func (d *cudaDevice) GetDisplayMode() (string, error) { return "unknown", nil }
```
elezar, line 100: nit: let's have this span multiple lines.

internal/resource/device_mock.go
```diff
@@ -324,3 +324,5 @@ func (mock *DeviceMock) IsMigEnabledCalls() []struct {
 	mock.lockIsMigEnabled.RUnlock()
 	return calls
 }
+
+func (mock *DeviceMock) GetDisplayMode() (string, error) { return "unknown", nil }
```
elezar, line 328: This is not correct: Run `make generate` to regenerate the mocks.

internal/resource/nvml-device.go
```diff
@@ -86,3 +90,37 @@ func (d nvmlDevice) GetTotalMemoryMB() (uint64, error) {
 	}
 	return info.Total / (1024 * 1024), nil
 }
+
+func (d nvmlDevice) GetDisplayMode() (string, error) {
+	info, retVal := d.Device.GetPciInfo()
```
elezar, line 95: What is the difference between this logic and the value returned by `nvmlDeviceGetDisplayMode` for the device?

internal/resource/nvml-device.go
```diff
@@ -86,3 +90,37 @@ func (d nvmlDevice) GetTotalMemoryMB() (uint64, error) {
...
+		return "", retVal
+	}
+	var bytes []byte
+	for _, char := range info.BusId {
+		if char == 0 {
+			break
+		}
+		bytes = append(bytes, byte(char))
+	}
+	pciID := strings.ToLower(strings.TrimPrefix(string(bytes), "0000"))
```
elezar, line 106: I recently added a `GetBusID` function to `go-nvml`: https://github.com/NVIDIA/go-nvlib/pull/41 does it make sense to merge that first?

internal/resource/nvml-device.go
```diff
@@ -86,3 +90,37 @@ func (d nvmlDevice) GetTotalMemoryMB() (uint64, error) {
...
+	return resolvePCIAddressToMode(pciID)
+}
+
+func resolvePCIAddressToMode(addr string) (string, error) {
```
elezar, line 110: This seems like a function that we should have in `go-nvpci` for instead of reimplementing it here.

internal/resource/sysfs-device.go
```diff
@@ -64,3 +64,7 @@ func (d vfioDevice) IsMigEnabled() (bool, error) {
 func (d vfioDevice) IsMigCapable() (bool, error) {
 	return false, nil
 }
+
+func (d vfioDevice) GetDisplayMode() (string, error) {
+	return resolvePCIAddressToMode(d.nvidiaPCIDevice.Address)
```
elezar, line 69: in a similar question to the nvml case, does it make sense to use the `d.nvidiaPCIDevice.` functions here directly?

tests/expected-output-mig-none.txt
```diff
@@ -18,6 +18,7 @@ nvidia\.com\/gpu\.sharing-strategy=[none|mps|time-slicing]
 nvidia\.com\/gpu\.product=[A-Za-z_-]+
 nvidia\.com\/gpu\.memory=[0-9]+
 nvidia\.com\/gpu\.family=[a-z]+
+nvidia\.com\/gpu\.mode=unknown
```
elezar, line 21: Should this be: `[unknown|compute|graphics]`?

docs/gpu-feature-discovery/README.md
```diff
@@ -209,6 +209,7 @@ their meaning:
 | nvidia.com/gpu.machine         | String     | Machine type                                 | DGX-1          |
 | nvidia.com/gpu.memory          | Integer    | Memory of the GPU in Mb                      | 2048           |
 | nvidia.com/gpu.product         | String     | Model of the GPU                             | GeForce-GT-710 |
+| nvidia.com/gpu.mode            | String     | Display or Compute Mode of the GPU           | compute        |
```
elezar, line 212: Could we provide more information on what this value means?

elezar (commented):
As a general comment, could we update the commit message to indicate that we are reporting the mode and not specifying it.

Then, should MIG devices have `unknown` reported? Should these not always be `compute`? What is the intended use of this label?

internal/resource/nvml-device.go
visheshtanksale: `nvmlDeviceGetDisplayMode` method indicates whether a physical display (like monitor) is currently connected to any of the device's connectors

internal/resource/nvml-device.go
visheshtanksale: Yes having that logic in go-nvlib is better

jojimt:
> Then, should MIG devices have `unknown` reported? Should these not always be `compute`? What is the intended use of this label?
MIG devices should report compute. The intended use of the label is to let an application select a worker node based on the mode available on the node.

internal/resource/cuda-device.go
visheshtanksale: Fixed.

internal/resource/device_mock.go
visheshtanksale: Fixed

internal/resource/sysfs-device.go
visheshtanksale: I changed the implementation here. `Class` is the property of the PCI device which is changed when the GPU is switched between modes. So we should just be getting `class` of the device. The interpretation of label based on class is the labelers job.

tests/expected-output-mig-none.txt
visheshtanksale: Fixed

internal/resource/nvml-device.go
visheshtanksale: I changed the implementation here to just read the `class`  based on the pci address. Maybe we should still push getting the class function to go-nvml.

visheshtanksale:
> As a general comment, could we update the commit message to indicate that we are reporting the mode and not specifying it.

 Fixed this
> Then, should MIG devices have `unknown` reported? Should these not always be `compute`? What is the intended use of this label?

Based on current list of GPUs that support mode switch it will be always compute. But fixed this to just pull it from the actual device.

internal/lm/nvml.go
```diff
@@ -193,3 +199,43 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
 	}
 	return true, nil
 }
+
+func newGPUModeLabeler(manager resource.Manager) (Labeler, error) {
+
```
elezar, line 204: Remove the empty line.

internal/resource/cuda-device.go
```diff
@@ -96,3 +96,7 @@ func (d *cudaDevice) IsMigCapable() (bool, error) {
 func (d *cudaDevice) IsMigEnabled() (bool, error) {
 	return false, nil
 }
+
+func (d *cudaDevice) GetClass() (string, error) {
```
elezar, line 100: Question: Does it make sense for the class to be an `int` instead? Do we really need to make a string comparison?

internal/resource/sysfs-device.go
```diff
@@ -64,3 +64,8 @@ func (d vfioDevice) IsMigEnabled() (bool, error) {
 func (d vfioDevice) IsMigCapable() (bool, error) {
 	return false, nil
 }
+
+func (d vfioDevice) GetClass() (string, error) {
+	class := fmt.Sprintf("0x0%x", d.nvidiaPCIDevice.Class)
```
elezar, line 69: Does  the following make more sense here:
```suggestion
	class := fmt.Sprintf("%#06x\n", d.nvidiaPCIDevice.Class)
```
https://go.dev/play/p/bPr3-R1GtPY

internal/resource/types.go
```diff
@@ -39,4 +39,5 @@ type Device interface {
 	GetTotalMemoryMB() (uint64, error)
 	GetDeviceHandleFromMigDeviceHandle() (Device, error)
 	GetCudaComputeCapability() (int, int, error)
+	GetClass() (string, error)
```
elezar, line 42: Should this be:
```suggestion
	GetPCIClass() (int, error)
```
Alternatively we can just have this be `GetMode() (string, error)`

internal/resource/nvml-device.go
```diff
@@ -86,3 +90,28 @@ func (d nvmlDevice) GetTotalMemoryMB() (uint64, error) {
...
+	return resolvePCIAddressToClass(pciID)
+}
+
+func resolvePCIAddressToClass(addr string) (string, error) {
+	class, err := os.ReadFile(fmt.Sprintf("/sys/bus/pci/devices/%s/class", addr))
+	if err != nil {
+		klog.Errorf("Error getting gpu class: %v", err)
+		return "unknown", fmt.Errorf("Error getting gpu class: %v", err)
+	}
+	return strings.TrimSpace(string(class)), nil
```
elezar, line 116: What about reusing the logic from `nvpci`:
```suggestion
	// TODO: The nvpci.Interface instance should be a member of 
	// nvmlDevice.
	device, err := nvpci.New().GetGPUByPciBusID(addr)
	if err != nil {
		return "", err
	}
	return fmt.Sprintf("%#06x", device.Class), nil
```

internal/lm/nvml.go
```diff
@@ -193,3 +199,43 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+		val, err := d.GetClass()
+		if err != nil {
+			return nil, err
+		}
+		if class != "" && val != class {
+			break
+		} else {
+			class = val
+		}
+	}
```
elezar, line 226: Let's move this to a different function that is a little more generic:
```
func getDevicePCIEClasses(devices []resource.Device) ([]string, error) {
	seenClasses := make(map[string]bool)
	for _, d := range devices {
		class, err := d.GetClass()
		if err != nil {
			return nil, err
		}
		seenClasses[class] = true
	}

	var classes []string
	for class := range seenClasses {
		classes = append(classes, class)
	}

	return classes, nil
}
```

We may also want to introduce a function that returns the mode from these classes (or for a slice of devices) and then use this to set the label value.

elezar (changes requested):

internal/resource/nvml-device.go
elezar: The place for this to live would always be `go-nvlib`. The package at `go-nvml` is a wrapper for NVML specifically and we don't really add additional functionality there.

internal/resource/types.go
visheshtanksale: Aren't all these devices PCIE. I feel  `GetClass() (int, error)` makes more sense.

internal/resource/sysfs-device.go
visheshtanksale: Fixed

internal/resource/nvml-device.go
visheshtanksale: Fixed this

internal/lm/nvml.go
visheshtanksale: Fixed this

internal/resource/types.go
visheshtanksale: All the devices are PCIE so I feel GetClass makes more sense.

internal/lm/nvml.go
visheshtanksale: Fixed

internal/resource/nvml-device.go
elezar: The PR has merged, so we should be able to update the dependecies and use it here.

internal/resource/types.go
elezar: The point is that from the point of view of the caller Class is very generic. We have NVML devices, CUDA Devices -- which include Tegra iGPUs, and SYSFS devices. Adding PCIE to the function name clarifies that we're talking about a class.

internal/resource/nvml-device.go
visheshtanksale: Updated it

internal/resource/types.go
visheshtanksale: Renamed the method

internal/resource/cuda-device.go
visheshtanksale: Used `uint32` because thats what nv-golib is returning

internal/lm/nvml.go
```diff
@@ -193,3 +201,59 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+		return "compute"
+	default:
+		return "unknown"
+	}
+}
+
+func getDeviceClasses(devices []resource.Device) ([]uint32, error) {
+	seenClasses := make(map[uint32]bool)
+	for _, d := range devices {
+		class, err := d.GetPIEClass()
```
elezar, line 247:
```suggestion
		class, err := d.GetPCIClass()
```

internal/resource/cuda-device.go
```diff
@@ -96,3 +96,7 @@ func (d *cudaDevice) IsMigCapable() (bool, error) {
 func (d *cudaDevice) IsMigEnabled() (bool, error) {
 	return false, nil
 }
+
+func (d *cudaDevice) GetPIEClass() (uint32, error) {
```
elezar, line 100:
```suggestion
func (d *cudaDevice) GetPCIClass() (uint32, error) {
```

internal/resource/nvml-device.go
```diff
@@ -86,3 +87,15 @@ func (d nvmlDevice) GetTotalMemoryMB() (uint64, error) {
 	}
 	return info.Total / (1024 * 1024), nil
 }
+
+func (d nvmlDevice) GetPIEClass() (uint32, error) {
```
elezar, line 91:
```suggestion
func (d nvmlDevice) GetPCIClass() (uint32, error) {
```

internal/resource/nvml-mig-device.go
```diff
@@ -132,3 +133,23 @@ func totalMemory(attr map[string]interface{}) (uint64, error) {
 		return 0, fmt.Errorf("unsupported attribute type %v", t)
 	}
 }
+
+func (d nvmlMigDevice) GetPIEClass() (uint32, error) {
+	info, retVal := d.MigDevice.GetPciInfo()
```
elezar, line 138: Question: Is PCI info valid for a MIG device? How does the busid differ from that of the parent?

internal/resource/nvml-mig-device.go
```diff
@@ -132,3 +133,23 @@ func totalMemory(attr map[string]interface{}) (uint64, error) {
...
+	if retVal != nvml.SUCCESS {
+		return 0, retVal
+	}
+	var bytes []byte
+	for _, char := range info.BusId {
+		if char == 0 {
+			break
+		}
+		bytes = append(bytes, byte(char))
+	}
```
elezar, line 148: Instead of implemention this here this should be pulled into `go-nvlib` if it is valid.

internal/resource/testing/resource-testing.go
```diff
@@ -51,6 +51,7 @@ func NewDeviceMock(migEnabled bool) *DeviceMock {
 		IsMigEnabledFunc:     func() (bool, error) { return migEnabled, nil },
 		IsMigCapableFunc:     func() (bool, error) { return migEnabled, nil },
 		GetMigDevicesFunc:    func() ([]resource.Device, error) { return nil, nil },
+		GetPIEClassFunc:      func() (uint32, error) { return 0x030000, nil },
```
elezar, line 54:
```suggestion
		GetPCIClassFunc:      func() (uint32, error) { return 0x030000, nil },
```

internal/resource/types.go
```diff
@@ -39,4 +39,5 @@ type Device interface {
 	GetTotalMemoryMB() (uint64, error)
 	GetDeviceHandleFromMigDeviceHandle() (Device, error)
 	GetCudaComputeCapability() (int, int, error)
+	GetPIEClass() (uint32, error)
```
elezar, line 42:
```suggestion
	GetPCIClass() (uint32, error)
```

internal/resource/nvml-mig-device.go
elezar: Happy to do this in a follow-up though.

internal/lm/nvml.go
```diff
@@ -193,3 +201,59 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
 	}
 	return true, nil
 }
+
+func newGPUModeLabeler(manager resource.Manager) (Labeler, error) {
```
elezar, line 205: let's add a resonable docstring to the constructor.

Also, we could consider using:
```suggestion
func newGPUModeLabeler(devices []resource.Device) (Labeler, error) {
```
instead since we already have a slice of devices as the callsite.

elezar, line 205: Does it make sense to add a unit test for this labeler?

internal/lm/nvml.go
```diff
@@ -193,3 +201,59 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+	devices, err := manager.GetDevices()
+	if err != nil {
+		return nil, err
+	}
+	if len(devices) == 0 {
+		// no devices, return empty labels
+		return empty{}, nil
```
elezar, line 212: Note that if `devices` is empty we return no labels, but if `classes` is empty, we return  `unknown` as the label. Which is the desired behaviour?

docs/gpu-feature-discovery/README.md
elezar: We still need to add a reference as to what this means. From the documentation it is also only applicable to specific device types. Do we want to link to the relevant documentation here?

internal/lm/nvml.go
```diff
@@ -193,3 +201,59 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+	}
+	classes, err := getDeviceClasses(devices)
+	if err != nil {
+		return nil, err
+	}
+	gpuMode := getModeForClasses(classes)
+	labels := Labels{
+		"nvidia.com/gpu.mode": gpuMode,
```
elezar, line 220: Question: Since we also extract this label for a MIG device do we expect different labels per MIG profile?

internal/lm/nvml.go
```diff
@@ -193,3 +201,59 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+		return "unknown"
+	}
+}
+
+func getDeviceClasses(devices []resource.Device) ([]uint32, error) {
+	seenClasses := make(map[uint32]bool)
+	for _, d := range devices {
+		class, err := d.GetPIEClass()
+		if err != nil {
+			return nil, err
```
elezar, line 249: Question: Do we want to treat errors in getting the class as fatal? This will crash GFD and cause NO labels to be generated. Should we rather return `unknown` as the label in this case?

internal/resource/sysfs-device.go
```diff
@@ -64,3 +64,7 @@ func (d vfioDevice) IsMigEnabled() (bool, error) {
 func (d vfioDevice) IsMigCapable() (bool, error) {
 	return false, nil
 }
+
+func (d vfioDevice) GetPIEClass() (uint32, error) {
```
elezar, line 68:
```suggestion
func (d vfioDevice) GetPCIClass() (uint32, error) {
```

elezar (changes requested):
Thanks @visheshtanksale.

There are some typos in the function name and I'm also not clear on what the behaviour should be for a MIG device.

internal/lm/nvml.go
visheshtanksale: I wanted to keep this behavior consistent with other labeler, that is the reason I am returning the error instead instead of labeling it `unknown` .

internal/lm/nvml.go
visheshtanksale: Fixed

internal/resource/nvml-device.go
visheshtanksale: Fixed

internal/resource/cuda-device.go
visheshtanksale: Fixed

internal/resource/sysfs-device.go
visheshtanksale: Fixed

internal/resource/testing/resource-testing.go
visheshtanksale: Fixed

internal/resource/types.go
visheshtanksale: Fixed

internal/lm/nvml.go
visheshtanksale: Used devices from the calling function

visheshtanksale: Added

internal/lm/nvml.go
visheshtanksale: They both are different behaviours. 

Devices being empty specifies no devices are present on the node so no need to label the node.

Classes are not being present (which is highly unlikely) is a state that devices are there on the node but we are not able to determine their mode. So, the label is `unknown`

internal/lm/nvml.go
visheshtanksale: Devices that support MIG do not support changing the GPU mode. They are always going to return compute PCI class. We can actually hard code the class of MIG devices to be compute.

internal/resource/nvml-mig-device.go
visheshtanksale: We can decide where to put this behavior once we decide what behavior is expected for MIG devices.

internal/lm/nvml.go
elezar: Yes, let's rather do that. It would simplify the implementation quite a bit.

docs/gpu-feature-discovery/README.md
visheshtanksale: FIxed

internal/lm/nvml.go
visheshtanksale: Updated the implementation

internal/resource/nvml-mig-device.go
visheshtanksale: Because of the hardcoding of the PCI class for MIG devices this code is removed

internal/lm/nvml.go
```diff
@@ -193,3 +201,53 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+	return labels, nil
+}
+
+func getModeForClasses(classes []uint32) string {
+	if len(classes) == 0 {
+		return "unknown"
+	}
+	for _, class := range classes {
+		if class != classes[0] {
+			return "unknown"
```
elezar, line 225: Not a blocker for this PR, but we may want to log the content of `classes` here as a warning.

internal/lm/nvml_test.go
```diff
@@ -204,3 +204,89 @@ func TestSharingLabeler(t *testing.T) {
 		})
 	}
 }
+
+func TestGPUModeLabeler(t *testing.T) {
```
elezar, line 208: Thanks for adding the tests!

internal/resource/testing/resource-testing.go
```diff
@@ -51,6 +51,14 @@ func NewDeviceMock(migEnabled bool) *DeviceMock {
 		IsMigEnabledFunc:     func() (bool, error) { return migEnabled, nil },
 		IsMigCapableFunc:     func() (bool, error) { return migEnabled, nil },
 		GetMigDevicesFunc:    func() ([]resource.Device, error) { return nil, nil },
+		GetPCIClassFunc:      func() (uint32, error) { return 0x030000, nil },
```
elezar, line 54: nit: Should this return `0` by default since it's "unknown" / "undefined"?

tests/expected-output-mig-single.txt
```diff
@@ -30,4 +30,5 @@ nvidia\.com\/gpu\.engines\.jpeg=[0-9]+
 nvidia\.com\/gpu\.engines\.ofa=[0-9]+
 nvidia\.com\/gpu\.slices\.gi=[0-9]+
 nvidia\.com\/gpu\.slices\.ci=[0-9]+
+nvidia\.com\/gpu\.mode=[unknown|compute|graphics]
```
elezar, line 33: nit: this can only be `compute` now, correct? Not a blocker.

elezar (approved):
Thanks @visheshtanksale.

I have some minor comments, but these can be addressed in a follow-up.

internal/resource/testing/resource-testing.go
visheshtanksale: Fixed

tests/expected-output-mig-single.txt
visheshtanksale: Fixed

PR #770, opened 2024-06-14
Set /dev/shm size from /proc/meminfo

elezar:
This change reads MemTotal from /proc/meminfo and sets
the dev/shm size to half this value. This aligns with the
default for creating tmpfs mounts.

If reading the value fails, the previous default of 65536k is used.

### DIFF
--- a/cmd/mps-control-daemon/mount/mount-shm.go
+++ b/cmd/mps-control-daemon/mount/mount-shm.go
@@ -17,22 +17,27 @@
 package mount
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
+	"strings"
 
 	"github.com/urfave/cli/v2"
+	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 )
 
 // NewCommand constructs a mount command.
 func NewCommand() *cli.Command {
-	// Create the 'generate-cdi' command
-	return &cli.Command{
+	c := cli.Command{
 		Name:   "mount-shm",
 		Usage:  "Set up the /dev/shm mount required by the MPS daemon",
 		Action: mountShm,
 	}
+
+	return &c
 }
 
 // mountShm creates a tmpfs mount at /mps/shm to be used by the mps control daemon.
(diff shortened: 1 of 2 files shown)


cmd/mps-control-daemon/mount/mount-shm.go
```diff
@@ -17,26 +17,46 @@
...
+		Name:  "mount-shm",
+		Usage: "Set up the /dev/shm mount required by the MPS daemon",
+		Action: func(ctx *cli.Context) error {
+			return mountShm(ctx, &opts)
+		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "shm-size",
+				Destination: &opts.shmSize,
+				Hidden:      true,
```
klueska, line 50: What does `hidden` do?

cmd/mps-control-daemon/mount/mount-shm.go
```diff
@@ -17,26 +17,46 @@
...
-		Usage:  "Set up the /dev/shm mount required by the MPS daemon",
-		Action: mountShm,
+	opts := options{}
+
+	c := cli.Command{
+		Name:  "mount-shm",
+		Usage: "Set up the /dev/shm mount required by the MPS daemon",
+		Action: func(ctx *cli.Context) error {
+			return mountShm(ctx, &opts)
+		},
```
klueska, line 45: Can we have this component directly use the plugins `config` file and avoid introducing flags?

cmd/mps-control-daemon/mount/mount-shm.go
```diff
@@ -54,11 +74,58 @@ func mountShm(c *cli.Context) error {
...
 	if err := mounter.Mount("shm", shmDir, "tmpfs", mountOptions); err != nil {
 		return fmt.Errorf("error mounting %v as tmpfs: %w", shmDir, err)
 	}
 
 	return nil
 }
+
+// getShmSize returns the default shm size.
+// This reads /proc/meminfo to get the total memory to calculate this.
+func (o *options) getShmSize() (size string) {
```
klueska, line 88:
```suggestion
func (o *options) getDefaultShmSize() (size string) {
```

cmd/mps-control-daemon/mount/mount-shm.go
elezar: It doesn't show it in the help text. I can remove this if required but I was still half-way between updating the default and allowing a user to configure this as per #699

cmd/mps-control-daemon/mount/mount-shm.go
```diff
@@ -54,11 +74,58 @@ func mountShm(c *cli.Context) error {
...
+	defer func() {
+		if size == "" {
+			size = "65536k"
+		}
+		klog.InfoS("Calculated /dev/shm size", "size", size)
+	}()
```
klueska, line 94: Why this instead of:
```
fallbackSize := "65536k"
```
And then explicitly return this on errors instead of `""`

deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -150,3 +150,9 @@ mps:
   # directories.
   # Pipe directories will be created at {{ mps.root }}/{{ .ResourceName }}
   root: "/run/nvidia/mps"
+  # devShmSize specifies the size in bytes of the tmpfs created for the MPS
+  # control daemon.
+  # The format should match that of the size argument (excluding the size=)
+  # prefix when mounting a tmpfs:
+  #    See https://man7.org/linux/man-pages/man5/tmpfs.5.html
+  devShmSize: null
```
klueska, line 158: So this is a helm chart setting not a plugin config setting?

cmd/mps-control-daemon/mount/mount-shm.go
elezar: Sure.

cmd/mps-control-daemon/mount/mount-shm.go
elezar: It doesn't get the default shm size. It also returns the size if set.

klueska: Then maybe the comment needs to change?

cmd/mps-control-daemon/mount/mount-shm.go
elezar: I have removed this option since it's not used.

cmd/mps-control-daemon/mount/mount-shm.go
elezar: I will remove the flags. I will work in the use of the config as part of #699 if we determine that this needs to be set.

One thing to note is that allowing this option to be set complicates the syncronisation between the various components considerably since this container would also need to be restarted on a config change.

deployments/helm/nvidia-device-plugin/values.yaml
elezar: Sorry, this is not applicable to this PR and should only be in #699

cmd/mps-control-daemon/mount/mount-shm.go
elezar: It doesn't get the default shm size. It also returns the size if set. Note that this assumes that we want to allow this to be set. I could remove the interactions with `options` to simplify.

elezar: Removed the addition of `options`. I will add that once it becomes relevant.

cmd/mps-control-daemon/mount/mount-shm.go
elezar: Updated.

deployments/helm/nvidia-device-plugin/values.yaml
elezar: These have been removed.

klueska (approved):

PR #773, opened 2024-06-17
Bump version v0.16.0-rc.1

elezar:


hack/generate-changelog.sh
```diff
@@ -77,4 +77,4 @@ echo ""
 echo "### Version $VERSION"
 
 # Iterate over the commit messages and ignore the ones that start with "Merge" or "Bump"
-git log --pretty=format:"%s" $REFERENCE..@ | grep -Ev "(^Merge )|(^Bump)" |  sed 's/^\(.*\)/- \1\\n/g' 
+git log --pretty=format:"%s" $REFERENCE..@ | grep -Ev "(^Merge )|(^Bump)" |  sed 's/^\(.*\)/- \1/g'
```
cdesiniotis, line 80: This is my first time seeing this. Nice!

cdesiniotis (approved):

hack/generate-changelog.sh
elezar: It does generate a very VERBOSE changelog, but it's probably better to add better checks and commit messages than to manually generate those anyway.

ArangoGutierrez (approved):
:)

CHANGELOG.md
```diff
@@ -1,5 +1,50 @@
 ## Changelog
 
+### Version v1.16.0-rc.1
+
+- Skip container updates if only CDI is selected
```
elezar, line 5: We definitely need to filter these a bit more strictly.

ArangoGutierrez: is just a helper script, to filter them better we would need to be more strict with Commit naming so we can use regex filters, or have an LLM action as a helper for this

PR #789, opened 2024-06-27
Modify SELinux context of MPS pipe directory

elezar:


cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -97,6 +98,12 @@ func (d *Daemon) Start() error {
 		return fmt.Errorf("error creating directory %v: %w", pipeDir, err)
 	}
 
+	if selinux.EnforceMode() == selinux.Enforcing {
+		if err := selinux.Chcon(pipeDir, "container_file_t", true); err != nil {
+			return fmt.Errorf("error setting SELinux context: %w", err)
+		}
+	}
```
empovit, line 105: It's a one-time initialization step, isn't it? I'm wondering if it's possible that a user sets SELinux to permissive (e.g. for debugging), installs the driver, and then changes SELinux back to enforcing. Won't it be safer to rely on `selinux.GetEnabled()` or `selinux.DefaultEnforceMode()`?

empovit: I've just tested `chcon` on a system with SELinux disabled, and it works. I guess it won't have any effect, but won't fail either.

elezar: I'm happy to switch to `GetEnabled()` if we think that's cleaner.

> It's a one-time initialization step, isn't it?

This step is only run as the MPS daemon container is started -- or if the configuration changes. What is the expected behaviour if a user changes the SELinux mode?

elezar: Just to clarify, `EnforceMode` contains the logic to extract the CURRENT mode.
```
// enforceMode returns the current SELinux mode Enforcing, Permissive, Disabled
func enforceMode() int {
	var enforce int

	enforceB, err := os.ReadFile(selinuxEnforcePath())
	if err != nil {
		return -1
	}
	enforce, err = strconv.Atoi(string(enforceB))
	if err != nil {
		return -1
	}
	return enforce
}
```
Whereas `GetEnabled()` checks whether SELinux is available at all -- regardless of the enforcing mode.

empovit:
LGTM

PR #790, opened 2024-06-27
Accept kubelet socket from command line

elezar:
This also makes it possible to test device plugin logic that does not depend on the kubelet.

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"syscall"
 	"time"
 
@@ -104,6 +105,12 @@ func main() {
 			Name:    "mofed-enabled",
 			Usage:   "ensure that containers are started with NVIDIA_MOFED=enabled",
 			EnvVars: []string{"MOFED_ENABLED"},
+		},
+		&cli.StringFlag{
+			Name:    "kubelet-socket",
+			Value:   pluginapi.KubeletSocket,
+			Usage:   "specify the socket for communicating with the kubelet; if this is empty, no connection with the kubelet is attempted",
+			EnvVars: []string{"KUBELET_SOCKET"},
 		},
 		&cli.StringFlag{
 			Name:        "config-file",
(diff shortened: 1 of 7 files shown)


cmd/nvidia-device-plugin/main.go
```diff
@@ -105,6 +106,12 @@ func main() {
 			Usage:   "ensure that containers are started with NVIDIA_MOFED=enabled",
 			EnvVars: []string{"MOFED_ENABLED"},
 		},
+		&cli.StringFlag{
+			Name:    "kubelet-socket",
+			Value:   pluginapi.KubeletSocket,
+			Usage:   "specify the socket for communicating with the kubelet; if this is empty, no connection with the kublet is attempted",
```
tariq1890, line 112:
```suggestion
			Usage:   "specify the socket for communicating with the kubelet; if this is empty, no connection with the kubelet is attempted",
```

cmd/nvidia-device-plugin/main.go
```diff
@@ -196,8 +203,10 @@ func loadConfig(c *cli.Context, flags []cli.Flag) (*spec.Config, error) {
 }
 
 func start(c *cli.Context, flags []cli.Flag) error {
-	klog.Info("Starting FS watcher.")
-	watcher, err := watch.Files(pluginapi.DevicePluginPath)
+	kubeletSocket := c.String("kubelet-socket")
+	kubeletSocketDir := filepath.Dir(kubeletSocket)
+	klog.Infof("Starting FS watcher for %v", kubeletSocketDir)
```
tariq1890, line 208:
```suggestion
	klog.Infof("Starting FS watcher for %s", kubeletSocketDir)
```

internal/plugin/api.go
```diff
@@ -21,6 +21,6 @@ import "github.com/NVIDIA/k8s-device-plugin/internal/rm"
 // Interface defines the API for the plugin package
 type Interface interface {
 	Devices() rm.Devices
-	Start() error
+	Start(string) error
```
tariq1890, line 24: why not add `kubeletSocket` as a field to the struct that implements this Interface? IMO it's better to avoid modifying the interface method signature if we can help it

cmd/nvidia-device-plugin/main.go
elezar: Will update. Thanks.

cmd/nvidia-device-plugin/main.go
elezar: Since `kubeletSocketDir` is a `string` these two should be equivalent. Is there a specific reason to prefer `%s` over `%v` in this case?

internal/plugin/api.go
elezar: Updated.

tariq1890 (approved):

internal/plugin/api.go
tariq1890: Thanks

cmd/nvidia-device-plugin/main.go
tariq1890: This is just a minor nit. I prefer `%s` as it explicitly says that variable is a string

tariq1890 (approved):

PR #792, opened 2024-06-29
[gfd] Fixing logic of atomically file writing

### LINKED ISSUE #791: [gfd] Incorrect implementation of atomic writing to a file when exporting features by gpu-feature-discovery
* OS/Version: Ubuntu 18.04.6, Ubuntu 22.04.2

### 2. Issue description

We use the node-feature-discovery and gpu-feature-discovery features to monitor GPU issues, including cases when the number of available GPUs on a node unexpectedly decreases: Target Number == `nvidia.com/gpu.count` == Node Allocatable.

We have noticed that sometimes (albeit rarely) all the features (labels `nvidia.com/*`) exported by gpu-feature-discovery disappear from the node. During one of these moments, we noticed the following line in the nfd-worker logs:
```log
E0629 17:29:25.340196       1 local.go:266] source local failed reading file 'gfd': open /etc/kubernetes/node-feature-discovery/features.d/gfd: permission denied
```

We suspect that this error occurs due to an incorrect implementation of the atomic file writing logic in the function [writeFileAtomically](https://github.com/NVIDIA/k8s-device-plugin/blob/35ad18080eded1889dc1eaee1132debddfd6757c/internal/lm/output.go#L105). The issue is that the `os.Chmod` command is executed after the `os.Rename` command, causing a moment when the `features.d/gfd` file has incorrect permissions (`0600` (default for a file created with [os.CreateTemp](https://cs.opensource.google/go/go/+/master:src/os/tempfile.go;l=49?q=CreateTemp&ss=go%2Fgo)) instead of `0644`).

### 3. Additional information that might help better understand your environment and reproduce the bug

The incorrectness of the current implementation of [writeFileAtomically](https://github.com/NVIDIA/k8s-device-plugin/blob/35ad18080eded1889dc1eaee1132debddfd6757c/internal/lm/output.go#L105) can be easily reproduced with the following example:

For more deterministic behavior, I added `time.Sleep(1 * time.Millisecond)` between the `os.Rename` and `os.Chmod` calls (but the error also reproduces without this, it just takes a bit longer to occur):

<details>
<summary>writer/main.go</summary>

```go
...
const (
	filePath = "/<abs/path/to>/gfd"
	content  = `
nvidia.com/gpu.machine=VMware7,1
nvidia.com/gpu.product=Tesla-V100-PCIE-32GB
`
)

func main() {
	for {
		_ = writeFileAtomically(filePath, []byte(content), 0644)
	}
}

// from https://github.com/NVIDIA/k8s-device-plugin/blob/35ad18080eded1889dc1eaee1132debddfd6757c/internal/lm/output.go#L105
func writeFileAtomically(path string, contents []byte, perm os.FileMode) error {
	absPath, err := filepath.Abs(path)
	if err != nil {
		return fmt.Errorf("failed to retrieve absolute path of output file: %v", err)
	}

	absDir := filepath.Dir(absPath)
	tmpDir := filepath.Join(absDir, "gfd-tmp")

	err = os.MkdirAll(tmpDir, os.ModePerm)
	if err != nil && !os.IsExist(err) {
		return fmt.Errorf("failed to create temporary directory: %v", err)
	}
	defer func() {
		if err != nil {
			os.RemoveAll(tmpDir)
		}
	}()

	tmpFile, err := os.CreateTemp(tmpDir, "gfd-")
	if err != nil {
		return fmt.Errorf("fail to create temporary output file: %v", err)
	}
	defer func() {
		if err != nil {
			tmpFile.Close()
			os.Remove(tmpFile.Name())
		}
	}()

	err = os.WriteFile(tmpFile.Name(), contents, perm)
	if err != nil {
		return fmt.Errorf("error writing temporary file '%v': %v", tmpFile.Name(), err)
	}

	err = os.Rename(tmpFile.Name(), path)
	if err != nil {
		return fmt.Errorf("error moving temporary file to '%v': %v", path, err)
	}

	time.Sleep(1 * time.Millisecond) // added for better error reproduction

	err = os.Chmod(path, perm)
	if err != nil {
		return fmt.Errorf("error setting permissions on '%v': %v", path, err)
	}

	return nil
}

```

</details>

The [nfd-worker when reading the file](https://github.com/kubernetes-sigs/node-feature-discovery/blob/0418e7ddf33424b150c68ca8fe71fcfc98440039/source/local/local.go#L438) simply calls the `os.ReadFile` function:

<details>
<summary>reader/main.go</summary>

```go
...
const filePath = "/<abs/path/to>/gfd"

func main() {
	for {
		_, err := os.ReadFile(filePath)
		if err != nil {
			log.Println(err)
		}
	}
}

```

</details>

Let's run it and see the expected error:

```bash
#terminal 1
go run writer/main.go

#terminal 2: run as another user since tmpFile has mode 0600 by default
sudo -u <another user> go run reader/main.go

#2024/06/30 01:55:00 open /<abs/path/to>/gfd: permission denied
#2024/06/30 01:55:00 open /<abs/path/to>/gfd: permission denied
#2024/06/30 01:55:00 open /<abs/path/to>/gfd: permission denied
#...
```

### 4. Additional information

Fixing this issue is simple. Just change the order of the `os.Rename` and `os.Chmod` calls to `os.Chmod -> os.Rename` in the [writeFileAtomically](https://github.com/NVIDIA/k8s-device-plugin/blob/35ad18080eded1889dc1eaee1132debddfd6757c/internal/lm/output.go#L105) function - #792 

I found another Issue that mentions this problem - #325. However, it also mentions an issue with selinux files, so I decided to open this separate Issue.

belo4ya:
fix #791

The current implementation of the [writeFileAtomically](https://github.com/NVIDIA/k8s-device-plugin/blob/35ad18080eded1889dc1eaee1132debddfd6757c/internal/lm/output.go#L105) function contains a bug - the `os.Chmod` function is called after `os.Rename`, causing a moment when the target file (`features.d/gfd`) has incorrect permissions (`0600` instead of `0644`):
```bash
E0629 17:29:25.340196       1 local.go:266] source local failed reading file 'gfd': open /etc/kubernetes/node-feature-discovery/features.d/gfd: permission denied
```

To fix this issue, the order of the calls needs to be changed to: `os.Chmod -> os.Rename`


belo4ya:
What do you think about using [github.com/google/renameio](https://github.com/google/renameio) instead of our own implementation of atomic file writing?
```go
import "github.com/google/renameio"

func writeFileAtomically(path string, contents []byte, perm os.FileMode) error {
	return renameio.WriteFile(path, contents, perm)
}
```

[github.com/google/renameio](https://github.com/google/renameio) is stable, has no dependencies, is quite popular ([Used by 4.3k](https://github.com/google/renameio/network/dependents)), and performs its task well.

elezar:
@belo4ya thanks for the fix. I think we would be ok with using `renameio`. Feel free to update this PR to switch to that instead. Alternatively we can do so as a follow-up.

belo4ya:
Done. I replaced the function `writeFileAtomically` with `renameio.WriteFile`.

The only thing is, the temporary file will be created in `$TMPDIR` instead of `/etc/kubernetes/node-feature-discovery/features.d/gfd-tmp/` (I don't think this is a problem, but it's worth mentioning).

elezar (approved):
Thanks for this fix @belo4ya.

elezar (changes requested):
With regards to:
> The only thing is, the temporary file will be created in $TMPDIR instead of /etc/kubernetes/node-feature-discovery/features.d/gfd-tmp/ (I don't think this is a problem, but it's worth mentioning).

Let's squash these two commits into a single commit and include this note in the commit message along with additional context on the fixed bug (something like your description would be great).

belo4ya:
@elezar, done.

elezar (approved):

PR #806, opened 2024-07-08
Warn instead of error when a GPU has no MIG devices with mixed strategy

klueska:


internal/rm/device_map.go
```diff
@@ -165,8 +166,7 @@ func (b *deviceMapBuilder) assertAllMigDevicesAreValid(uniform bool) error {
 			return err
 		}
 		if len(migDevices) == 0 {
-			i := 0
-			return fmt.Errorf("device %v has an invalid MIG configuration", i)
+			klog.Warningf("device %v has no MIG devices configured", i)
```
elezar, line 169: I think we need to make the return of an error conditional on the `uniform` argument. 

```
	var requireUniformMIGDevices bool
	if *b.migStrategy == spec.MigStrategySingle {
		requireUniformMIGDevices = true
	}
```
thus something like:
```suggestion
			if uniform {
				return fmt.Errorf("device %v has no MIG devices configured", i)
			}
			klog.Warningf("device %v has no MIG devices configured", i)
```

elezar (changes requested):

PR #814, opened 2024-07-11
[no-relnotes] Use dependabot to update go toolchain

elezar:
This change uses Dependabot to update the golang toolcain.

The `deployments/devel/Dockfile` is used as a reference for defining the golang version through the `FROM` clause.


ArangoGutierrez (approved):
proven to work at toolkit repo

PR #823, opened 2024-07-15
Add node name environment variable to the GFD deployments

lalitadithya:
Without the `NODE_NAME` environment variable, GFD was failing to start with an error. The `NODE_NAME` env var was present in the helm chart but was missing in the static deployments. This change adds the missing `NODE_NAME` env var to the static deployment.


ArangoGutierrez (commented):
LGTM
@elezar ?

PR #870, opened 2024-08-05
[no-relnote] Change context during E2E call

ArangoGutierrez:
This PR enables E2E after https://github.com/NVIDIA/k8s-device-plugin/pull/855

### DIFF
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -31,7 +31,7 @@ e2e-test:
 		echo "[ERR] KUBECONFIG missing, must be defined"; \
 		exit 1; \
 	fi
-	$(GO_CMD) test -v $(CURDIR)/tests/e2e -args \
+	cd $(CURDIR)/tests/e2e && $(GO_CMD) test -v . -args \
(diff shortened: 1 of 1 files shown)


tariq1890 (approved):

tests/e2e/Makefile
```diff
@@ -31,7 +31,7 @@ e2e-test:
 		echo "[ERR] KUBECONFIG missing, must be defined"; \
 		exit 1; \
 	fi
-	$(GO_CMD) test -v $(CURDIR)/tests/e2e -args \
+	cd $(CURDIR)/tests/e2e && $(GO_CMD) test -v . -args \
```
elezar, line 34: Question: Do we need to cd back to the project root afterwards?

elezar (approved):
Thanks @ArangoGutierrez 

one minor question, but probably not a blocker

tests/e2e/Makefile
ArangoGutierrez: no, that's handled by Make, I had the same question my self, but after a manual run I saw my self on the same PATH

PR #871, opened 2024-08-05
[no-relnote] Run a parallel Staging E2E test job

ArangoGutierrez:


ArangoGutierrez:
Depends on https://github.com/NVIDIA/k8s-device-plugin/pull/870

elezar (commented):
As a general question, what benefit does a nightly run give us here?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,52 @@
...
+          echo "VSPHERE_SSH_KEY=${VSPHERE_SSH_KEY}" >> $GITHUB_ENV
+      - name: Set up Holodeck
+        uses: NVIDIA/holodeck@v0.2.1
+        with:
+          holodeck_config: "tests/e2e/infra/vsphere.yaml"
+      - name: Run e2e tests
+        env:
+          KUBECONFIG: ${{ github.workspace }}/kubeconfig
+          E2E_IMAGE_REPO: ghcr.io/nvidia/k8s-device-plugin
+          E2E_IMAGE_TAG: ${{ github.base_ref }}
```
elezar, line 42: Why not just use `COMMIT_SHORT_SHA` here?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Nightly End-to-end Tests
+
+on:
+  schedule:
+    - cron: '00 09 * * *' # scheduled job
```
elezar, line 19: Nit. I think we can drop the comment.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
```
elezar, line 21: We seem to be duplicating the e2e job that we already have defined. Is there a way to reuse that instead?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Nightly End-to-end Tests
+
+on:
+  schedule:
```
elezar, line 18: Can't we just also include a `schedule` in the existing e2e test workflow?

.github/workflows/nightly.yaml
ArangoGutierrez: Done

.github/workflows/nightly.yaml
ArangoGutierrez: Done

.github/workflows/nightly.yaml
ArangoGutierrez: Since we can't have a conditional on the `runs-on` everything after it has to be duplicated

.github/workflows/nightly.yaml
ArangoGutierrez: Done (with the caveat of duplicated code, Since we can't have a conditional on the runs-on everything after it has to be duplicated)

.github/workflows/e2e.yaml
```diff
@@ -40,7 +42,39 @@ jobs:
...
+    if: ${{ github.event.workflow_run.event == 'schedule' }}
+    steps:
+      - uses: actions/checkout@v4
+        name: Check out code
+      - name: Calculate build vars
+        id: vars
+        run: |
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
+          echo "HOLODECK_VCENTER_USERNAME=${HOLODECK_VCENTER_USERNAME}" >> $GITHUB_ENV
+          echo "HOLODECK_VCENTER_PASSWORD=${HOLODECK_VCENTER_PASSWORD}" >> $GITHUB_ENV
```
elezar, line 72: Doesnt' this mean it's already in the env?

.github/workflows/image.yaml
```diff
@@ -70,6 +70,7 @@ jobs:
         env:
           IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/k8s-device-plugin
           VERSION: ${COMMIT_SHORT_SHA}
+          IMAGE_EXTRA_TAG_NAMES: ${{ github.base_ref }}
```
elezar, line 73: Should be removed?

.github/workflows/e2e.yaml
```diff
@@ -40,7 +42,39 @@ jobs:
...
+        run: |
+          make -f tests/e2e/Makefile e2e-test
+      - name: Archive test logs
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-test-logs
+          path: ./e2e_logs/
+          retention-days: 15
+  nightly-e2e:
```
elezar, line 61: Instead of doing this nightly, why not just also trigger it on push? As I mentioned, I don't see the value of nightly runs until we have better notifications of failures.

.github/workflows/e2e.yaml
ArangoGutierrez: since `holodeck` is a container action, we need to pass the env to the container env.

.github/workflows/e2e.yaml
ArangoGutierrez: how about now, I called it `staging` to refer as a job that is not a prod job run

.github/workflows/e2e.yaml
```diff
@@ -40,7 +40,39 @@ jobs:
           aws_access_key_id: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
elezar, line 40: Why are there not equivalent constructs for the vsphere settings? Using different mechanisms to configure the two different infrastructure providers makes it seem as if something is missing.

.github/workflows/e2e.yaml
elezar: For the aws kas we use `with` instead of setting the env. Why the difference?

ArangoGutierrez: Because these are ENV VARS at the runner level, `with` is to pass GitHub stored secrets.  Since the idea for this Staging job is to test the internal runners, this marks a difference.

.github/workflows/e2e.yaml
ArangoGutierrez: https://github.com/NVIDIA/k8s-device-plugin/pull/871/files#r1706671562

.github/workflows/e2e.yaml
elezar: I don't think that really explains why we can't do:
```
with:
  vcenter_username: ${{ secrets.VCENTER_USERNAME }}
  vcenter_password: ${{ secrets.VCENTER_PASSWORD }}
  vsphere_ssh_key: ${{ secrets.VSPHERE_SSH_KEY }}
```

It is my understanding that these are converted to `INPUT_*` environment variables in any case and will be accessible to holodeck.

Maybe I'm missing something though.

ArangoGutierrez: GitHub don't inject variables during a non `Post` job, or a job that is forked. 
This Staging job is to test that we can run e2e during `pull` events, meaning we won't have secrets injected by GitHub in the default way, but we are doing secret injection via GitHub ARC config

elezar: But this job is:
```
if: ${{ github.event.workflow_run.conclusion == 'success' }} && ${{ github.event.workflow_run.event == 'push' }}
```
which is not a PR trigger, or am I missing something?

tests/e2e/infra/vsphere.yml
```diff
@@ -0,0 +1,30 @@
...
+    privateKey: /home/nvidia/.ssh/gpu_runner
+  VsphereVirtualMachine:
+    vCenterURL: "pdx01egxvc01.nvidia.com"
+    datacenter: "PDX01"
+    datastore: "PDX01EGX_vSAN_001"
+    cluster: "PDX01EGX001"
+    network: "PDX01EGXDVS001_Internal_307_10.185.111.0_24"
+    vmFolder: "GitHub ARC"
+    resoursePool: "CNT-ARC"
+    templateImage: "GPU_UBUNTU_TEMPLATE"
```
elezar, line 19: Not critical for this PR, but this seems like the incorrect place to define this. If this is shared across projects, should it not be defined centrally -- or even on the runners themselves -- so that required changes do not need to be made to every project where this is used.

ArangoGutierrez: Makes sense, this could be handled by Holodeck or the k8s-test-infra repo

.github/workflows/e2e.yaml
```diff
@@ -40,15 +40,15 @@ jobs:
...
+          holodeck_config: "tests/e2e/infra/aws.yaml"
       - name: Run e2e tests
         env:
           KUBECONFIG: ${{ github.workspace }}/kubeconfig
           E2E_IMAGE_REPO: ghcr.io/nvidia/k8s-device-plugin
           E2E_IMAGE_TAG: ${COMMIT_SHORT_SHA}-ubuntu22.04
           LOG_ARTIFACTS: ${{ github.workspace }}/e2e_logs
         run: |
-          make test-e2e
+          make -f tests/e2e/Makefile e2e-test
```
elezar, line 51: This should be `make test-e2e`

tests/e2e/infra/vsphere.yml
elezar: I'm ok to do that as a follow up.

.github/workflows/staging_e2e.yaml
```diff
@@ -0,0 +1,58 @@
...
+    types:
+      - completed
+    branches:
+      - main
+      - release-*
+
+jobs:
+  staging-e2e:
+    runs-on: pdx01-arc-runners
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
elezar, line 29: Question: How is the image made accessible to the tests in this context?

.github/workflows/e2e.yaml
ArangoGutierrez: Moved to it's own workflow file, and now runs on PR's (if the image workflow passes)

elezar (changes requested):

.github/workflows/staging_e2e.yaml
ArangoGutierrez: umm we will have to push it. meaning we can't test during PR's unless we modify the image building. I totally forgot about that. Should we leave it for Merge events, as the currently running e2e?

elezar: Yes, let's start there.

ArangoGutierrez: done

elezar (approved):

PR #872, opened 2024-08-06
Add IMAGE_EXTRA_TAGS for extra image tags

ArangoGutierrez:
Moving this bit out of #871 

This will allow us to have a tagged image for the development branch, for test purposes


.github/workflows/image.yaml
```diff
@@ -70,6 +70,7 @@ jobs:
         env:
           IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/k8s-device-plugin
           VERSION: ${COMMIT_SHORT_SHA}
+          IMAGE_EXTRA_TAG_NAMES: ${{ github.base_ref }}
```
elezar, line 73: Why is the SHA-tagged image not sufficient?

ArangoGutierrez: For nightly tests we won't have a SHA, Nightly test will be run against a branch tag (main)

deployments/container/Makefile
```diff
@@ -79,6 +82,7 @@ $(IMAGE_TARGETS): image-%:
 		$(DOCKER_BUILD_OPTIONS) \
 		$(DOCKER_BUILD_PLATFORM_OPTIONS) \
 		--tag $(IMAGE) \
+		$(foreach tag,$(IMAGE_EXTRA_TAGS),-t $(tag)) \
```
elezar, line 85: nit: If we're going to do this we should at least use `--tag` for consistency.

deployments/container/Makefile
```diff
@@ -29,13 +29,16 @@ endif
 
 IMAGE_VERSION := $(VERSION)
 
+IMAGE_EXTRA_TAG_NAMES ?=
 IMAGE_TAG ?= $(IMAGE_VERSION)-$(DIST)
 IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
+IMAGE_EXTRA_TAGS := $(foreach tag,$(IMAGE_EXTRA_TAG_NAMES),$(IMAGE_NAME):$(tag))
```
elezar, line 35: We lose the dist-specific tags? This means that for cases where we do produce two distributions they specified tags will be overridden.

PR #880, opened 2024-08-07
[no-relnote] Increase timeout for GPU job test to 10 minutes

elezar:
The device plugin e2e tests regularly fail due to a timeout of 5 minutes:
```
[FAILED] Timed out after 300.001s.
```
This introduced flakiness into the tests. 

This change increased the timeout for the device plugin job checks to 10 minutes instead.


ArangoGutierrez (approved):
Some GPU' are faster than others...

elezar:
> Some GPU' are faster than others...

I don't think this is GPU speed -- especially not for the sample that we consider.

PR #885, opened 2024-08-08
[no-relnote] Set GH_TOKEN for helm publish step

elezar:


ArangoGutierrez (approved):

elezar:
One note: Should we move to `github.token` over `secrets.GITHUB_TOKEN`?

PR #891, opened 2024-08-09
[no-relnote] Add CodeQL conde scanning

ArangoGutierrez:
Inspired in https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-starter-workflows-for-code-scanning


ArangoGutierrez:
First run of `CodeQL / Analyze Go code with CodeQL` will fail, until the file is merged.

ArangoGutierrez:
Commit message has also been updated to fix the typo

.github/workflows/code_scanning.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - synchronize
+    branches:
+      - main
+      - release-*
+  push:
+    branches:
+      - main
+      - release-*
+  schedule:
+    - cron: '31 11 * * 4'
```
elezar, line 30: Question: What is the benefit of a `cron` job?

elezar:
Question: Does this annotate the code in the PR or do we need to check the Security tab?

.github/workflows/code_scanning.yaml
ArangoGutierrez: Catch a CVE or a new malpractice

ArangoGutierrez:
> Question: Does this annotate the code in the PR or do we need to check the Security tab?

https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/triaging-code-scanning-alerts-in-pull-requests#code-scanning-results-check

.github/workflows/code_scanning.yaml
elezar: For projects such as the device plugin where we're already triggering this on PRs -- including dependency bumps -- and on merge, I don't think this is required. Let's rather remove this and add scheduled scans as a follow-up once we are sure that the existing signals are not sufficient.

ArangoGutierrez: Done

ArangoGutierrez:
Ready for review, all requested changes have been addressed

elezar (approved):

PR #898, opened 2024-08-11
modify the log level of some errors

googs1025:
- modify the log level of some errors
- fix method comment

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -142,7 +142,7 @@ func (plugin *NvidiaDevicePlugin) Start() error {
 
 	err := plugin.Serve()
 	if err != nil {
-		klog.Infof("Could not start device plugin for '%s': %s", plugin.rm.Resource(), err)
+		klog.Errorf("Could not start device plugin for '%s': %s", plugin.rm.Resource(), err)
 		plugin.cleanup()
 		return err
 	}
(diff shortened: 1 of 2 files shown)


elezar (approved):

internal/plugin/server.go
```diff
@@ -157,7 +157,7 @@ func (plugin *NvidiaDevicePlugin) Start() error {
 		// TODO: add MPS health check
 		err := plugin.rm.CheckHealth(plugin.stop, plugin.health)
 		if err != nil {
-			klog.Infof("Failed to start health check: %v; continuing with health checks disabled", err)
+			klog.Errorf("Failed to start health check: %v; continuing with health checks disabled", err)
```
elezar, line 160: This could also be a warning instead, but I don't think it's a blocker.

googs1025:
do we have to fix this ci error?

```bash
Error: internal/resource/cuda-lib.go:60:15: G115: integer overflow conversion int -> uint (gosec)
  	major := uint(version) / 1000
  	             ^
  Error: internal/resource/cuda-lib.go:61:15: G115: integer overflow conversion int -> uint (gosec)
  	minor := uint(version) % 100 / 10
  	             ^
  Error: internal/resource/nvml-lib.go:44:15: G115: integer overflow conversion int -> uint (gosec)
  	major := uint(v / 1000)
  	             ^
  Error: internal/resource/nvml-mig-device.go:1[31](https://github.com/NVIDIA/k8s-device-plugin/actions/runs/10904087630/job/30259653352?pr=898#step:5:33):16: G115: integer overflow conversion int -> uint64 (gosec)
  		return uint64(totalMemory.(int)), nil

```

internal/plugin/server.go
googs1025: Thanks for your reply :)

googs1025:
friendly ping @elezar

Issue #796, opened 2024-07-01
[gfd] Add option to disable automatic cleanup features file on gpu-feature-discovery exit


belo4ya:
### Issue description

We use the node-feature-discovery and gpu-feature-discovery features to monitor GPU issues, including cases when the number of available GPUs on a node unexpectedly decreases: `Target Number == nvidia.com/gpu.count == Node Allocatable`.

We have noticed that sometimes after restarting gpu-feature-discovery, all the features (labels `nvidia.com/*`) exported by gpu-feature-discovery disappear from the node for a period roughly equal to the nfd-worker `sleepInterval` (in our case, 1 minute). This causes false positives in our monitoring system.

We found that this occurs because gpu-feature-discovery deletes the `features.d/gfd` file before terminating if it is not running in one-shot mode (done using the [removeOutputFile function](https://github.com/NVIDIA/k8s-device-plugin/blob/35ad18080eded1889dc1eaee1132debddfd6757c/cmd/gpu-feature-discovery/main.go#L241)).

This behavior is very inconvenient (and undesirable) for us, especially when updating the gpu-feature-discovery version in the cluster.

### Feature request

I found that this behavior was added with this commit - https://github.com/NVIDIA/gpu-feature-discovery/commit/bc91c4aec84c2bc3e6da47789d6d0a0326330455. However, I did not find an associated Issue justifying the need for this behavior.

Could you please consider:

1. Adding an option to disable automatic cleanup before gpu-feature-discovery terminates using a flag (and/or environment variable) (e.g., `--no-cleanup-on-exit`).
2. Or the refusal to automatically clean up before shutting down the gpu-feature-discovery.

An argument for 2. could be that node-feature-discovery does not do this. Instead, it uses a [prune-job](https://kubernetes-sigs.github.io/node-feature-discovery/v0.15/deployment/uninstallation.html#removing-feature-labels).

belo4ya:
@elezar, @klueska, @ArangoGutierrez, please take a look at this

elezar:
Thanks @belo4ya, I have created #899 to add this option and we can continue this discussion there.

@ArangoGutierrez one thing that I noted is that we don't do any cleanup when the NodeFeatureAPI is used. How are labels removed in this case?

PR #899, opened 2024-08-12
Add --no-cleanup-on-exit option to GFD

elezar:
This change adds a --no-cleanup-on-exit option to GFD that skips the removal of the generated labels file.

See #796


belo4ya:
@elezar, are there any updates or issues related to this fix? We are still eagerly awaiting its inclusion in the next release.

---

One thing: as was rightly noted in comment https://github.com/NVIDIA/k8s-device-plugin/issues/796#issuecomment-2283859991, when using the NodeFeature API, such automatic cleanup does not occur. Considering this, such a general option name as `--no-cleanup-on-exit` may not be entirely clear and could cause some confusion. Perhaps it would be better to choose a more suitable name?

taehwoi:
@elezar are there any updates or issues related to this fix? We would be grateful if this gets included in the next release.

PR #905, opened 2024-08-14
update default go version of the main container image

tariq1890:
As the go.mod file pins the minimum go version to 1.22.2, it no longer makes sense for the container's default go version to be a values lesser than that. Sp we bump the default golang build arg to 1.22.6

Signed-off-by: Tariq Ibrahim <tibrahim@nvidia.com>


elezar (approved):

tariq1890:
Thanks for the review @elezar . I have rebased the PR on top of the latest changes

PR #972, opened 2024-09-30
Fix golang-ci-lint errors

elezar:
This PR makes the following changes to address `golangci-lint` errors:

* updates the `GetCudaDriverVersion` return type to `(int, int, error)` instead of `(*uint, *uint, error)`. This prevents unneeded type conversions and pointer operations.
* adds an explicit check in the testing-only `int` -> `uint64` type conversion for MIG memory.


ArangoGutierrez (approved):
thanks

tariq1890 (approved):

PR #975, opened 2024-10-04
add go mod check for deployments/devel and enable it in CI

tariq1890:


.github/workflows/golang.yaml
```diff
@@ -49,7 +49,9 @@ jobs:
         args: -v --timeout 5m
         skip-cache: true
     - name: Check golang modules
-      run: make check-vendor
+      run: |
+        make check-vendor
+        make -C deployments/devel check-modules
```
elezar, line 54: Question: Why don't we use the same target in all cases?

elezar: OK. Should we then rename the target in `Makefile`?

tariq1890: Sure. So make it `check-modules` everywhere?

elezar: yes, let's do that.

tariq1890: Done, please check now

elezar (approved):

PR #985, opened 2024-10-09
 Add optional injection of the default IMEX channel

elezar:
These changes allow users to op-in to injecting the default IMEX channel (`channel0`) into workloads.

The following config options are added:
```
imex:
  channelIDs: []
  required: false
```
where `imex.channelIDs` indicates the list of channels (if any) to be injected into workload containers. The only valid values are currently `[]` and `[0]`.

If a channel is specified, the plugin will check if the corresponding device node exists and inject the channel into the `ContainerAllocateResponse` if it does. If `imex.required == true` then an error is raised if the device node does not exist.

The mechanism of injection depends on the `deviceListStrategy` as follows:
* `deviceListStrategy=envvar`: the IMEX channel is requested by setting the `NVIDIA_IMEX_CHANNELS` environment variable in `ContainerAllocateResponse`.
* `deviceListStrategy=cdi*`: the device plugin generates a CDI specification for the IMEX channel and adds a CDI device request to the `ContainerAllocateResponse`.
* `deviceListStrategy=volume-mounts`: the IMEX channel is requested by adding a `/var/run/nvidia-container-devices/imex/0` volume mount. This requires the toolkit support from https://github.com/NVIDIA/nvidia-container-toolkit/pull/740

---

**Testing**

Build the image
```shell
make -f deployments/container/Makefile build VERSION=imex-devel ARCH=arm64
```
and ensure that the `nvcr.io/nvidia/k8s-device-plugin:imex-devel-ubi9` is available to each node.

Create the following namespace for the demos:
```
kubectl create ns demo
```

Create the following config files:
```
cat << EOF > dp-default-config.yaml
version: v1
EOF
```

```
cat << EOF > dp-imex-envvar-config.yaml
version: v1
flags:
  plugin:
    deviceListStrategy:
    - envvar
imex:
  channelIDs:
  - 0
EOF
```

```
cat << EOF > dp-imex-cdi-annotations-config.yaml
version: v1
flags:
  plugin:
    deviceListStrategy:
    - cdi-annotations
imex:
  channelIDs:
  - 0
EOF
```

Deploy the device plugin:
```shell
helm upgrade nvidia-device-plugin -i deployments/helm/nvidia-device-plugin \
    --namespace nvidia-device-plugin \
    --create-namespace \
    --set nvidiaDriverRoot=/ \
    --set runtimeClassName=nvidia \
    --set image.tag=imex-devel-ubi9 \
    --set-file config.map.default=dp-default-config.yaml \
    --set-file config.map.imex-envvar-config=dp-imex-envvar-config.yaml \
    --set-file config.map.imex-cdi-annotations-config=dp-imex-cdi-annotations-config.yaml \
```

**Note:** We set `nvidiaDriverRoot` to allow for IMEX devs to be located at `/driver-root/dev`.
This is also required for CDI spec generation to function as expected.

## Default configuration

Confirm that the imex config is empty:
```shell
kubectl logs -n nvidia-device-plugin \
    $(kubectl get pods -n nvidia-device-plugin -o custom-columns=NAME:.metadata.name | tail -1) \
    -c nvidia-device-plugin-ctr | grep -A2 imex
```

```
kubectl logs -n nvidia-device-plugin     $(kubectl get pods -n nvidia-device-plugin -o custom-columns=NAME:.metadata.name | tail -1)     -c nvidia-device-plugin-ctr | grep -A2 imex
  "imex": {}
}
I1011 12:52:06.624171      44 main.go:334] Retrieving plugins.
```

### Deploy a workload

Run a no-op sample requesting a GPU:
```
cat <<EOF | kubectl apply -f -
---
apiVersion: v1
kind: Pod
metadata:
  namespace: demo
  name: demo-pod
spec:
  runtimeClassName: nvidia
  restartPolicy: OnFailure
  containers:
  - name: demo-ctr
    image: ubuntu
    command: ["sleep", "9999"]
    resources:
      limits:
        nvidia.com/gpu: 1
EOF
```

Check the container environment and which NVIDIA devices are injected:
```
kubectl exec -ti -n demo demo-pod -- bash -c "echo 'devices'; ls -al /dev/nv*; echo 'env'; env | grep NVIDIA_"
```
```
devices
crw-rw-rw- 1 root root 502,   0 Oct 10 04:14 /dev/nvidia-uvm
crw-rw-rw- 1 root root 502,   1 Oct 10 04:14 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   0 Oct 10 04:13 /dev/nvidia0
crw-rw-rw- 1 root root 195, 255 Oct 10 04:13 /dev/nvidiactl
env
NVIDIA_VISIBLE_DEVICES=GPU-101ac189-c246-adb9-6e96-f7ab66dee379
```

Remove the demo pod:
```
kubectl delete pod -n demo demo-pod
```

## DefaultChannelStrategy=enabled deviceListStrategy=envvar

Apply the required config:
```shell
kubectl label node --all --overwrite nvidia.com/device-plugin.config=imex-envvar-config
```

Confirm that the IMEX config is applied:
```
$ kubectl logs -n nvidia-device-plugin     $(kubectl get pods -n nvidia-device-plugin -o custom-columns=NAME:.metadata.name | tail -1)     -c nvidia-device-plugin-ctr | grep -A3 imex
  "imex": {}
}
I1011 12:53:22.710244      43 main.go:334] Retrieving plugins.
I1011 12:53:22.757053      43 server.go:218] Starting GRPC server for 'nvidia.com/gpu'
--
  "imex": {
    "channelIDs": [
      "0"
    ]
```
(note that since the plugin main loop is restarted, the original output is also captured)

### Deploy a workload

Run a no-op sample requesting a GPU:
```
cat <<EOF | kubectl apply -f -
---
apiVersion: v1
kind: Pod
metadata:
  namespace: demo
  name: demo-pod
spec:
  runtimeClassName: nvidia
  restartPolicy: OnFailure
  containers:
  - name: demo-ctr
    image: ubuntu
    command: ["sleep", "9999"]
    resources:
      limits:
        nvidia.com/gpu: 1
EOF
```

Check the container environment and which NVIDIA devices are injected:
```
kubectl exec -ti -n demo demo-pod -- bash -c "echo 'devices'; ls -al /dev/nv*; echo 'env'; env | grep NVIDIA_"
```
```
devices
crw-rw-rw- 1 root root 502,   0 Oct 10 04:14 /dev/nvidia-uvm
crw-rw-rw- 1 root root 502,   1 Oct 10 04:14 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   1 Oct 10 04:13 /dev/nvidia1
crw-rw-rw- 1 root root 195, 255 Oct 10 04:13 /dev/nvidiactl

/dev/nvidia-caps-imex-channels:
total 0
drwxr-xr-x 2 root root     60 Oct 10 14:21 .
drwxr-xr-x 6 root root    460 Oct 10 14:21 ..
crw-rw-rw- 1 root root 508, 0 Oct 10 04:15 channel0
env
NVIDIA_VISIBLE_DEVICES=GPU-970c1f22-f1ca-482d-ab14-aea1d83ba922
NVIDIA_IMEX_CHANNELS=0
```

Remove the demo pod:
```
kubectl delete pod -n demo demo-pod
```

## DefaultChannelStrategy=enabled deviceListStrategy=cdi-annotations

Apply the required config:
```shell
kubectl label node --all --overwrite nvidia.com/device-plugin.config=imex-cdi-annotations-config
```

Confirm that the IMEX config is applied:
```
 kubectl logs -n nvidia-device-plugin     $(kubectl get pods -n nvidia-device-plugin -o custom-columns=NAME:.metadata.name | tail -1)     -c nvidia-device-plugin-ctr | grep -i -A3 imex
  "imex": {}
}
I1011 12:53:22.710244      43 main.go:334] Retrieving plugins.
I1011 12:53:22.757053      43 server.go:218] Starting GRPC server for 'nvidia.com/gpu'
--
  "imex": {
    "channelIDs": [
      "0"
    ]
--
  "imex": {
    "channelIDs": [
      "0"
    ]
--
time="2024-10-11T12:57:26Z" level=info msg="Generating CDI spec for resource: k8s.device-plugin.nvidia.com/imex-channel"
I1011 12:57:26.876356      43 server.go:218] Starting GRPC server for 'nvidia.com/gpu'
I1011 12:57:26.876780      43 server.go:149] Starting to serve 'nvidia.com/gpu' on /var/lib/kubelet/device-plugins/nvidia-gpu.sock
I1011 12:57:26.878187      43 server.go:156] Registered device plugin for 'nvidia.com/gpu' with Kubelet
```
Note the additional logging for generating the IMEX channel CDI spec.

When this mode is configured we can check the following on one of the nodes:
```
$ nvidia-ctk cdi list 2>/dev/null
k8s.device-plugin.nvidia.com/gpu=GPU-30dcdd9d-2489-f021-8297-5ff9406e5cc9
k8s.device-plugin.nvidia.com/gpu=GPU-43f40e92-a51c-a02d-cb07-ac8c8b6b8965
k8s.device-plugin.nvidia.com/imex-channel=0
nvidia.com/gpu=0
nvidia.com/gpu=1
nvidia.com/gpu=GPU-30dcdd9d-2489-f021-8297-5ff9406e5cc9
nvidia.com/gpu=GPU-43f40e92-a51c-a02d-cb07-ac8c8b6b8965
nvidia.com/gpu=all
```
with:
```
cat /var/run/cdi/k8s.device-plugin.nvidia.com-imex-channel.json | jq
{
  "cdiVersion": "0.5.0",
  "kind": "k8s.device-plugin.nvidia.com/imex-channel",
  "devices": [
    {
      "name": "0",
      "containerEdits": {
        "deviceNodes": [
          {
            "path": "/dev/nvidia-caps-imex-channels/channel0",
            "hostPath": "/dev/nvidia-caps-imex-channels/channel0"
          }
        ]
      }
    }
  ],
  "containerEdits": {}
}
```

### Deploy a workload

Run a no-op sample requesting a GPU:
```
cat <<EOF | kubectl apply -f -
---
apiVersion: v1
kind: Pod
metadata:
  namespace: demo
  name: demo-pod
spec:
  runtimeClassName: nvidia
  restartPolicy: OnFailure
  containers:
  - name: demo-ctr
    image: ubuntu
    command: ["sleep", "9999"]
    resources:
      limits:
        nvidia.com/gpu: 1
EOF
```

Check the container environment and which NVIDIA devices are injected:
```
kubectl exec -ti -n demo demo-pod -- bash -c "echo 'devices'; ls -al /dev/nv*; echo 'env'; env | grep NVIDIA_"
```
```
devices
crw-rw-rw- 1 root root 195, 254 Oct 10 14:28 /dev/nvidia-modeset
crw-rw-rw- 1 root root 502,   0 Oct 10 14:28 /dev/nvidia-uvm
crw-rw-rw- 1 root root 502,   1 Oct 10 14:28 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   1 Oct 10 14:28 /dev/nvidia1
crw-rw-rw- 1 root root 195, 255 Oct 10 14:28 /dev/nvidiactl

/dev/nvidia-caps-imex-channels:
total 0
drwxr-xr-x 2 root root     60 Oct 10 14:28 .
drwxr-xr-x 7 root root    500 Oct 10 14:28 ..
crw-rw-rw- 1 root root 510, 0 Oct 10 14:28 channel0
env
NVIDIA_VISIBLE_DEVICES=void
```
Note that no `NVIDIA_IMEX_CHANNELS` envvar is present.

Remove the demo pod:
```
kubectl delete pod -n demo demo-pod
```

### DIFF
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	Flags     Flags     `json:"flags,omitempty"     yaml:"flags,omitempty"`
 	Resources Resources `json:"resources,omitempty" yaml:"resources,omitempty"`
 	Sharing   Sharing   `json:"sharing,omitempty"   yaml:"sharing,omitempty"`
+	Imex      Imex      `json:"imex,omitempty"      yaml:"imex,omitempty"`
 }
 
 // NewConfig builds out a Config struct from a config file (or command line flags).
@@ -52,6 +53,14 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
 	}
 
 	config.Flags.UpdateFromCLIFlags(c, flags)
+	// TODO: This is currently not at the flags level?
+	// Does this mean that we should move UpdateFromCLIFlags to function off Config?
+	if c.IsSet("imex-channel-ids") {
+		config.Imex.ChannelIDs = c.IntSlice("imex-channel-ids")
+	}
+	if c.IsSet("imex-required") {
+		config.Imex.Required = c.Bool("imex-required")
+	}
 
 	// If nvidiaDevRoot (the path to the device nodes on the host) is not set,
 	// we default to using the driver root on the host.
--- a/api/config/v1/imex.go
+++ b/api/config/v1/imex.go
@@ -1,0 +1,53 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package v1
+
+import (
+	"errors"
+	"fmt"
+)
+
+const (
+	ImexChannelEnvVar = "NVIDIA_IMEX_CHANNELS"
+)
+
+var errInvalidImexConfig = errors.New("invalid IMEX config")
+
+// Imex stores the configuration options for fabric-attached devices.
+type Imex struct {
+	// ChannelIDs defines a list of channel IDs to inject into containers that request NVIDIA devices.
+	// If a channel ID is specified and the associated channel device node exists, the corresponding
+	// channel will be added to the ContainerAllocateResponse for containers with access to NVIDIA
+	// devices.
+	ChannelIDs []int `json:"channelIDs,omitempty" yaml:"channelIDs,omitempty"`
+	// Required specifies whether the requested IMEX channel IDs are required or not.
+	// If a channel is required, it is expected to exist as the device plugin starts.
+	// If it is not required its injection is skipped if the device nodes do not exist or if its
+	// existence cannot be queried.
+	Required bool `json:"required,omitempty" yaml:"required,omitempty"`
+}
+
+// AssertChannelIDsIsValid checks whether the specified list of channel IDs is valid.
+func AssertChannelIDsValid(ids []int) error {
+	switch {
+	case len(ids) == 0:
+		return nil
+	case len(ids) == 1 && ids[0] == 0:
+		return nil
+	}
+	return fmt.Errorf("%w: channelIDs must be [] or [0]; found %v", errInvalidImexConfig, ids)
+}
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -149,6 +149,16 @@ func main() {
 			Usage:   "the strategy to use to discover devices: 'auto', 'nvml', or 'tegra'",
 			EnvVars: []string{"DEVICE_DISCOVERY_STRATEGY"},
 		},
+		&cli.IntSliceFlag{
+			Name:    "imex-channel-ids",
+			Usage:   "A list of IMEX channels to inject.",
+			EnvVars: []string{"IMEX_CHANNEL_IDS"},
+		},
+		&cli.BoolFlag{
+			Name:    "imex-required",
+			Usage:   "The specified IMEX channels are required",
+			EnvVars: []string{"IMEX_REQUIRED"},
+		},
 	}
 
 	err := c.Run(os.Args)
@@ -188,6 +198,10 @@ func validateFlags(infolib nvinfo.Interface, config *spec.Config) error {
 	case "tegra":
 	default:
 		return fmt.Errorf("invalid --device-discovery-strategy option %v", *config.Flags.DeviceDiscoveryStrategy)
+	}
+
+	if err := spec.AssertChannelIDsValid(config.Imex.ChannelIDs); err != nil {
+		return fmt.Errorf("invalid IMEX channel IDs: %w", err)
 	}
 
 	return nil
--- a/internal/cdi/cdi.go
+++ b/internal/cdi/cdi.go
@@ -33,6 +33,7 @@ import (
 	cdiparser "tags.cncf.io/container-device-interface/pkg/parser"
 
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
+	"github.com/NVIDIA/k8s-device-plugin/internal/imex"
 )
 
 const (
@@ -59,7 +60,9 @@ type cdiHandler struct {
 	gdsEnabled   bool
 	mofedEnabled bool
 
-	cdilibs map[string]nvcdi.Interface
+	imexChannels imex.Channels
+
+	cdilibs map[string]cdiSpecGenerator
 }
 
 var _ Interface = &cdiHandler{}
@@ -108,7 +111,7 @@ func New(infolib info.Interface, nvmllib nvml.Interface, devicelib device.Interf
 		return nil, err
 	}
 
-	c.cdilibs = make(map[string]nvcdi.Interface)
+	c.cdilibs = make(map[string]cdiSpecGenerator)
 
 	c.cdilibs["gpu"], err = nvcdi.New(
 		nvcdi.WithInfoLib(c.infolib),
@@ -124,6 +127,10 @@ func New(infolib info.Interface, nvmllib nvml.Interface, devicelib device.Interf
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create nvcdi library: %v", err)
+	}
+
+	if len(c.imexChannels) > 0 {
+		c.cdilibs["imex-channel"] = c.newImexChannelSpecGenerator()
 	}
 
 	var additionalModes []string
--- a/internal/imex/imex.go
+++ b/internal/imex/imex.go
@@ -1,0 +1,98 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package imex
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"k8s.io/klog/v2"
+
+	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
+)
+
+// Channels represents a set of IMEX channels.
+type Channels []*Channel
+
+// Channel represents an IMEX channel.
+type Channel struct {
+	ID       string
+	Path     string
+	HostPath string
+}
+
+// GetChannels returns the set of channels for the given config.
+// If the selection of the default IMEX channel is disabled no channels are returned.
+func GetChannels(config *spec.Config, devRoot string) (Channels, error) {
+	var channels Channels
+	for _, channelID := range config.Imex.ChannelIDs {
+		id := fmt.Sprintf("%d", channelID)
+		channelName := "channel" + id
+		path := filepath.Join("/dev/nvidia-caps-imex-channels", channelName)
+		channel := Channel{
+			ID:       id,
+			Path:     path,
+			HostPath: filepath.Join(devRoot, path),
+		}
+		if exists, err := channel.exists(); !exists {
+			if config.Imex.Required {
+				return nil, errors.Join(err, fmt.Errorf("requested IMEX channel %v does not exist", channelName))
+			}
+			klog.Warningf("Ignoring requested IMEX channel %v (%v)", channelName, err)
+			continue
+		}
+		klog.Infof("Selecting IMEX channel %v", channelName)
+		channels = append(channels, &channel)
+	}
+	return channels, nil
+}
+
+// exists checks whether the IMEX channel exists.
+// We check both the Path and HostPath since the location of the device node
+// associated with the channel in the container is dependent on how it is
+// injected.
+// For example, if the host driver root is mounted at /driver-root the channel
+// device node would be available at /driver-root/dev even if it was not
+// injected into the container through any other mechanism.
+// For the case of management containers using CDI to inject device nodes, these
+// device nodes would exist at /dev in the container instead.
+func (c Channel) exists() (bool, error) {
+	paths := []string{c.HostPath}
+	if c.HostPath != c.Path {
+		paths = append(paths, c.Path)
+	}
+	var errs error
+	for _, path := range paths {
+		info, err := os.Stat(path)
+		if os.IsNotExist(err) {
+			continue
+		}
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
+
+		if info.Mode()&os.ModeCharDevice == 0 {
+			errs = errors.Join(errs, fmt.Errorf("%v is not a character device", path))
+			continue
+		}
+		return true, nil
+	}
+	return false, errs
+}
--- a/internal/plugin/manager/nvml.go
+++ b/internal/plugin/manager/nvml.go
@@ -34,7 +34,7 @@ func (m *nvmlmanager) GetPlugins() ([]plugin.Interface, error) {
 
 	var plugins []plugin.Interface
 	for _, r := range rms {
-		plugin, err := plugin.NewNvidiaDevicePlugin(m.config, m.kubeletSocket, r, m.cdiHandler)
+		plugin, err := plugin.NewNvidiaDevicePlugin(m.config, m.kubeletSocket, r, m.cdiHandler, m.imexChannels)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create plugin: %w", err)
 		}
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -31,6 +31,7 @@ import (
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 	"github.com/NVIDIA/k8s-device-plugin/cmd/mps-control-daemon/mps"
 	"github.com/NVIDIA/k8s-device-plugin/internal/cdi"
+	"github.com/NVIDIA/k8s-device-plugin/internal/imex"
 	"github.com/NVIDIA/k8s-device-plugin/internal/rm"
 
 	"github.com/google/uuid"
@@ -67,10 +68,12 @@ type NvidiaDevicePlugin struct {
 
 	mpsDaemon   *mps.Daemon
 	mpsHostRoot mps.Root
+
+	imexChannels imex.Channels
 }
 
 // NewNvidiaDevicePlugin returns an initialized NvidiaDevicePlugin
-func NewNvidiaDevicePlugin(config *spec.Config, kubeletSocket string, resourceManager rm.ResourceManager, cdiHandler cdi.Interface) (*NvidiaDevicePlugin, error) {
+func NewNvidiaDevicePlugin(config *spec.Config, kubeletSocket string, resourceManager rm.ResourceManager, cdiHandler cdi.Interface, imexChannels imex.Channels) (*NvidiaDevicePlugin, error) {
 	_, name := resourceManager.Resource().Split()
 
 	deviceListStrategies, _ := spec.NewDeviceListStrategies(*config.Flags.Plugin.DeviceListStrategy)
@@ -99,6 +102,8 @@ func NewNvidiaDevicePlugin(config *spec.Config, kubeletSocket string, resourceMa
 		socket:               pluginPath + ".sock",
 		cdiHandler:           cdiHandler,
 		cdiAnnotationPrefix:  *config.Flags.Plugin.CDIAnnotationPrefix,
+
+		imexChannels: imexChannels,
 
 		mpsDaemon:   mpsDaemon,
 		mpsHostRoot: mpsHostRoot,
@@ -364,6 +369,7 @@ func (plugin *NvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyEnvVar) {
 		plugin.updateResponseForDeviceListEnvVar(response, deviceIDs...)
+		plugin.updateResponseForImexChannelsEnvVar(response)
 	}
 	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyVolumeMounts) {
 		plugin.updateResponseForDeviceMounts(response, deviceIDs...)
@@ -406,6 +412,9 @@ func (plugin *NvidiaDevicePlugin) updateResponseForCDI(response *pluginapi.Conta
 	var devices []string
 	for _, id := range deviceIDs {
 		devices = append(devices, plugin.cdiHandler.QualifiedName("gpu", id))
+	}
+	for _, channel := range plugin.imexChannels {
+		devices = append(devices, plugin.cdiHandler.QualifiedName("imex-channel", channel.ID))
 	}
 	if *plugin.config.Flags.GDSEnabled {
 		devices = append(devices, plugin.cdiHandler.QualifiedName("gds", "all"))
@@ -502,13 +511,32 @@ func (plugin *NvidiaDevicePlugin) updateResponseForDeviceListEnvVar(response *pl
 	response.Envs[plugin.deviceListEnvVar] = strings.Join(deviceIDs, ",")
 }
 
+// updateResponseForImexChannelsEnvVar sets the environment variable for the requested IMEX channels.
+func (plugin *NvidiaDevicePlugin) updateResponseForImexChannelsEnvVar(response *pluginapi.ContainerAllocateResponse) {
+	var channelIDs []string
+	for _, channel := range plugin.imexChannels {
+		channelIDs = append(channelIDs, channel.ID)
+	}
+	if len(channelIDs) > 0 {
+		response.Envs[spec.ImexChannelEnvVar] = strings.Join(channelIDs, ",")
+	}
+}
+
 // updateResponseForDeviceMounts sets the mounts required to request devices if volume mounts are used.
 func (plugin *NvidiaDevicePlugin) updateResponseForDeviceMounts(response *pluginapi.ContainerAllocateResponse, deviceIDs ...string) {
 	plugin.updateResponseForDeviceListEnvVar(response, deviceListAsVolumeMountsContainerPathRoot)
+
 	for _, id := range deviceIDs {
 		mount := &pluginapi.Mount{
 			HostPath:      deviceListAsVolumeMountsHostPath,
 			ContainerPath: filepath.Join(deviceListAsVolumeMountsContainerPathRoot, id),
+		}
+		response.Mounts = append(response.Mounts, mount)
+	}
+	for _, channel := range plugin.imexChannels {
+		mount := &pluginapi.Mount{
+			HostPath:      deviceListAsVolumeMountsHostPath,
+			ContainerPath: filepath.Join(deviceListAsVolumeMountsContainerPathRoot, "imex", channel.ID),
 		}
 		response.Mounts = append(response.Mounts, mount)
 	}
@@ -534,6 +562,15 @@ func (plugin *NvidiaDevicePlugin) apiDeviceSpecs(devRoot string, ids []string) [
 		spec := &pluginapi.DeviceSpec{
 			ContainerPath: p,
 			HostPath:      filepath.Join(devRoot, p),
+			Permissions:   "rw",
+		}
+		specs = append(specs, spec)
+	}
+
+	for _, channel := range plugin.imexChannels {
+		spec := &pluginapi.DeviceSpec{
+			ContainerPath: channel.Path,
+			HostPath:      channel.HostPath,
 			Permissions:   "rw",
 		}
 		specs = append(specs, spec)
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
   - [Shared Access to GPUs](#shared-access-to-gpus)
     - [With CUDA Time-Slicing](#with-cuda-time-slicing)
     - [With CUDA MPS](#with-cuda-mps)
+  - [IMEX Support](#imex-support)
 - [Deployment via `helm`](#deployment-via-helm)
   - [Configuring the device plugin's `helm` chart](#configuring-the-device-plugins-helm-chart)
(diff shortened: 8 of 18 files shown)


api/config/v1/config.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	Flags     Flags     `json:"flags,omitempty"     yaml:"flags,omitempty"`
 	Resources Resources `json:"resources,omitempty" yaml:"resources,omitempty"`
 	Sharing   Sharing   `json:"sharing,omitempty"   yaml:"sharing,omitempty"`
+	Imex      Imex      `json:"imex,omitempty"      yaml:"imex,omitempty"`
```
elezar, line 38: I added an `Imex` config section here. Does it make sense to also move the domain config file to this section?

api/config/v1/imex.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math"
+)
+
+const (
+	ImexChannelEnvvar = "NVIDIA_IMEX_CHANNELS"
```
ArangoGutierrez, line 27: nit: `EnvVar`

cmd/nvidia-device-plugin/main.go
```diff
@@ -142,6 +142,12 @@ func main() {
 			Usage:   "the strategy to use to discover devices: 'auto', 'nvml', or 'tegra'",
 			EnvVars: []string{"DEVICE_DISCOVERY_STRATEGY"},
 		},
+		&cli.StringFlag{
+			Name:    "imex-default-channel-strategy",
+			Value:   "disalbed",
```
ArangoGutierrez, line 147:
```suggestion
			Value:   "disabled",
```

api/config/v1/imex.go
elezar: The use of `Envvar` is consistent with other such variables in the code base. I would rather create a chore to replace all occurrences and update this then.

internal/imex/imex.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Name returns the channel name.
+func (c Channel) Name() string {
+	return "channel" + c.id
+}
+
+// Path returns the absolute path to an IMEX channel.
+func (c Channel) Path() string {
+	return filepath.Join("/dev/nvidia-caps-imex-channels", c.Name())
```
ArangoGutierrez, line 96: is this something we want to hard code, isn't here going to be a customer/use case with a custom path?

api/config/v1/config.go
ArangoGutierrez: As discussed in Slack, I agree with this approach

api/config/v1/imex.go
ArangoGutierrez: - https://github.com/NVIDIA/k8s-device-plugin/pull/988

internal/imex/imex.go
elezar: This will always be the path of an IMEX channel.

api/config/v1/config.go
```diff
@@ -52,6 +53,15 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
 	}
 
 	config.Flags.UpdateFromCLIFlags(c, flags)
+	// TODO: This is currently not at the flags level?
+	// Does this mean that we should move UpdateFromCLIFlags to function off Config?
+	if c.IsSet("imex-default-channel-strategy") {
```
klueska, line 58: Would it simplify things to have the config option be a list of the IMEX channels to inject (if they exist)?

internal/imex/imex.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
+)
+
+// Channels represents a set of IMEX channels.
+type Channels []*Channel
```
klueska, line 27: I'm not sure how useful it is to support more than one channel. I know we originally had NVIDIA_IMEX_CHANNELS take a list, but that was because we didn't really understand how RM treats the available channels it sees. It only picks up and uses the lowest numbered one it sees -- all others are ignored.

api/config/v1/imex.go
```diff
@@ -0,0 +1,98 @@
...
+
+	DefaultChannelStrategyAuto     = DefaultChannelStrategy("auto")
+	DefaultChannelStrategyEnabled  = DefaultChannelStrategy("enabled")
+	DefaultChannelStrategyDisabled = DefaultChannelStrategy("disabled")
+
+	defaultChannelStrategyInvalid = DefaultChannelStrategy("invalid")
```
klueska, line 33: Do we really need all of these options?

If we accept my suggestion above, the existence of a channel in the config implies "auto" and the lack of one implies "disabled" using the terminology defined here. I don't really see the need for a strict "enabled", but that could be added later via a single boolean option of `imex.must-inject` or something.

internal/plugin/server.go
```diff
@@ -500,6 +503,10 @@ func (plugin *NvidiaDevicePlugin) apiDevices() []*pluginapi.Device {
 // updateResponseForDeviceListEnvvar sets the environment variable for the requested devices.
 func (plugin *NvidiaDevicePlugin) updateResponseForDeviceListEnvvar(response *pluginapi.ContainerAllocateResponse, deviceIDs ...string) {
 	response.Envs[plugin.deviceListEnvvar] = strings.Join(deviceIDs, ",")
+
+	if ids := plugin.rm.GetImexChannelIDs(); len(ids) > 0 {
```
elezar, line 507: @klueska what is the expected behaviour if we are using `volume-mounts`?

internal/imex/imex.go
elezar: I don't think we need to support multiple channels, but dealing with this as a slice allows for callers to not have to check for empty values on return. In practice we always iterate over zero or 1 elements.

Let me take another pass at it to see whether using a scalar here would make sense.

klueska: We can keep the data structure as a slice if that's easier (and even accept the input as a slie with the only valid value being `[0]`).

api/config/v1/imex.go
elezar: The behaviour is as follows:
* `disabled`: no IMEX-specific modifications are made.
* `enabled`: IMEX-modifications are made regardless of whether the device plugin has access to the device nodes from within the container. This means that workload containers will fail if they are started and the device nodes are not present. (the exact implementation of this in libnvidia-container would have to be checked, but at least in CDI the OCI spec cannot be constructed since we can't query the relevant information from the host device node).
* `auto`: The device plugin checks for the presence of the (default) channel and behaves as `enabled` if it can be found, if not we run without `imex` support. This allows for the GPU Operator to set a single value to `auto` and expect the device plugin to be sane wherever it is installed, meaning that it makes the decision as to whether IMEX is enabled or not.

I don't really see the benefit of having two possibilities `auto`, `disabled` for one option and a second option (`must-inject`) which only has meaning in conjunction with `auto`.

I'm happy to talk through scenarios that we need to cover here to see whether I'm missing something though.

I think we **can** remove the `invalid` value which is internal-only.

internal/plugin/server.go
klueska: This is specific to the "legacy", i.e. non-CDI case.

I guess it makes sense to have a separate container path mounted from /dev/null on the host for the IMEX channels, similar to the GPUs. SOmething like `/var/run/nvidia-imex-channels/<channel-num>` and then set the envvar to this path like we do here: https://github.com/NVIDIA/k8s-device-plugin/blob/966d5b8fe63bfb09ac26b46404e990bf3b5808e1/internal/plugin/server.go#L507

api/config/v1/config.go
elezar: I don't think this is a strong enough signal if we want to have an "auto" mode. We could get around this by allowing users to specify another flag (`imex.must-inject` or `imex.required`) that controls this behaviour. 

Note that the `TODO` here is more about the config API than IMEX in general and the same questions / comment would be applicable even if we do use something like `imex.defaultChannelIndex`.

internal/plugin/server.go
elezar: I suppose that in this case we would probably want to ALSO set the `NVIDIA_IMEX_CHANNELS` envvar.

klueska: That's what I mean by "and then set the envvar to this path like we do here..."

api/config/v1/config.go
klueska: I guess I just don't like the "switch" of modes. I'm also not a huge fan of them in the toolkit with the "auto" vs. "legacy" vs. "cdi" modes, but that ship has sailed. I would like to avoid it here if we can, and introduce as simple an API as possible without adding too much complexity in the code.

internal/plugin/server.go
elezar: > I guess it makes sense to have a separate container path mounted from /dev/null on the host for the IMEX channels, similar to the GPUs. Something like /var/run/nvidia-imex-channels/<channel-num> and then set the envvar to this path like we do here:

OK. This would requires explicit support in the NVIDIA Container Runtime. As long as we're ok introducing that requirement. 

My suggestion was that we do **the same thing** as what we do for the `envvar` case. That is to say, when `volume-mounts` is selected, we would request devices via `/dev/null` mounts and use the standard legacy mode for IMEX channels. This is probably not the right way to go though.

api/config/v1/config.go
elezar: The reason for the complexity is that this is expected to work "everywhere" and we are shifting the responsibility for making a decision into the plugin instead of requiring that the GPU Operator (or the user) make a configuration decision. This was also the case in the toolkit. We didn't want to have to tell users on non-NVML platforms that they need to make changes and since the support matrix is complex, the implementation is more complex than what it needs to be.

Regardless, I'll spend some time thinking about this tomorrow. The injection semantics as covered in this PR are simple enough and as such the discussion is more about the API.

api/config/v1/flags.go
```diff
@@ -114,7 +114,7 @@ type GFDCommandLineFlags struct {
 	ImexNodesConfigFile *string `json:"imexNodesConfigFile" yaml:"imexNodesConfigFile"`
 }
 
-// UpdateFromCLIFlags updates Flags from settings in the cli Flags if they are set.
+// UpdateFromCLIFlags Flags from settings in the cli Flags if they are set.
```
klueska, line 117: accidental?

api/config/v1/imex.go
```diff
@@ -0,0 +1,102 @@
...
+				if err := json.Unmarshal(r, &intslice); err == nil {
+					var strslice []string
+					for _, i := range intslice {
+						strslice = append(strslice, fmt.Sprintf("%d", i))
+					}
+					return strslice, nil
+				}
+
+				return nil, fmt.Errorf("unexpected channelIDs: %v", r)
+			}
```
klueska, line 71: Would it be simpler to only accept an intslice?

internal/plugin/manager/nvml.go
```diff
@@ -27,7 +27,7 @@ type nvmlmanager manager
 
 // GetPlugins returns the plugins associated with the NVML resources available on the node
 func (m *nvmlmanager) GetPlugins() ([]plugin.Interface, error) {
-	rms, err := rm.NewNVMLResourceManagers(m.infolib, m.nvmllib, m.devicelib, m.config)
+	rms, err := rm.NewNVMLResourceManagers(m.infolib, m.nvmllib, m.devicelib, m.config, m.imexChannels)
```
klueska, line 30: Is this needed anymore since the list of imex channels is embedded in the config?

api/config/v1/imex.go
elezar: It would probably be simpler, yes. Can we think of a reason to support strings instead?

internal/plugin/manager/nvml.go
elezar: This is still needed. The config values are used to construct a set of `imex.Channels`. The existence of device nodes is only checked at the point of construction.

api/config/v1/imex.go
elezar: Alternatively should we allow:
```
imex:
  channelIDs: 0
```
in addition to:
```
imex:
  channelIDs: [0]
```

elezar: Added the ints-only config as a commit on top. I think that is sufficient and definitely simpler.

elezar: I could even do away with the custom marshaller and check for a valid config in `validateFlags`?

elezar:
```diff --git a/api/config/v1/imex.go b/api/config/v1/imex.go
index a3fd7b5e..ca3050fa 100644
--- a/api/config/v1/imex.go
+++ b/api/config/v1/imex.go
@@ -17,7 +17,6 @@
 package v1
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 )
@@ -42,37 +41,6 @@ type Imex struct {
 	Required bool `json:"required,omitempty" yaml:"required,omitempty"`
 }
 
-// UnmarshalJSON implements the custom unmarshaler for fields of the IMEX config that require validation.
-func (f *Imex) UnmarshalJSON(b []byte) error {
-	rr := make(map[string]json.RawMessage)
-	if err := json.Unmarshal(b, &rr); err != nil {
-		return err
-	}
-
-	result := Imex{}
-	for k, v := range rr {
-		switch k {
-		case "channelIDs":
-			var intslice []int
-			if err := json.Unmarshal(v, &intslice); err != nil {
-				return fmt.Errorf("%w: %v", errInvalidImexConfig, err)
-			}
-			if err := AssertChannelIDsValid(intslice); err != nil {
-				return fmt.Errorf("%w: %v", errInvalidImexConfig, err)
-			}
-			result.ChannelIDs = intslice
-		case "required":
-			var value bool
-			if err := json.Unmarshal(v, &value); err != nil {
-				return err
-			}
-			result.Required = value
-		}
-	}
-	*f = result
-	return nil
-}
-
 // AssertChannelIDsIsValid checks whether the specified list of channel IDs is valid.
 func AssertChannelIDsValid(ids []int) error {
 	switch {
diff --git a/cmd/nvidia-device-plugin/main.go b/cmd/nvidia-device-plugin/main.go
index 6099305b..f6cb1e4a 100644
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -153,9 +153,6 @@ func main() {
 			Name:    "imex-channel-id",
 			Usage:   "A list of IMEX channels to inject.",
 			EnvVars: []string{"IMEX_CHANNEL_IDS"},
-			Action: func(ctx *cli.Context, s []int) error {
-				return spec.AssertChannelIDsValid(s)
-			},
 		},
 	}
 
@@ -198,6 +195,10 @@ func validateFlags(infolib nvinfo.Interface, config *spec.Config) error {
 		return fmt.Errorf("invalid --device-discovery-strategy option %v", *config.Flags.DeviceDiscoveryStrategy)
 	}
 
+	if err := spec.AssertChannelIDsValid(config.Imex.ChannelIDs); err != nil {
+		return fmt.Errorf("invalid IMEX channel IDs: %w", err)
+	}
+
 	return nil
 }
 
```

cmd/nvidia-device-plugin/main.go
```diff
@@ -149,6 +149,11 @@ func main() {
 			Usage:   "the strategy to use to discover devices: 'auto', 'nvml', or 'tegra'",
 			EnvVars: []string{"DEVICE_DISCOVERY_STRATEGY"},
 		},
+		&cli.IntSliceFlag{
+			Name:    "imex-channel-id",
```
klueska, line 153:
```suggestion
			Name:    "imex-channel-ids",
```

internal/cdi/cdi.go
```diff
@@ -126,6 +129,10 @@ func New(infolib info.Interface, nvmllib nvml.Interface, devicelib device.Interf
 		return nil, fmt.Errorf("failed to create nvcdi library: %v", err)
 	}
 
+	if len(c.imexChannels) > 0 {
+		c.cdilibs["imex-channel"] = c.newImexChannelSpecGenerator()
+	}
+
```
klueska, line 135: If / when we integrate this with the nvidia-container-toolkit, will this just become another case of the loop below, i.e.
```

	for _, mode := range additionalModes {
		lib, err := nvcdi.New(
			nvcdi.WithInfoLib(c.infolib),
			nvcdi.WithLogger(c.logger),
			nvcdi.WithNVIDIACDIHookPath(c.nvidiaCTKPath),
			nvcdi.WithDriverRoot(c.driverRoot),
			nvcdi.WithDevRoot(c.devRoot),
			nvcdi.WithVendor(c.vendor),
			nvcdi.WithMode(mode),
		)
		if err != nil {
			return nil, fmt.Errorf("failed to create nvcdi library: %v", err)
		}
		c.cdilibs[mode] = lib
	}
```

internal/imex/imex.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// Channels represents a set of IMEX channels.
+type Channels []*channel
+
+// channel represents an IMEX channel.
+type channel struct {
+	id      string
+	devRoot string
+}
```
klueska, line 37: This abstraction "feels" a little strange to me (in that when i just see this at the top of the file its not clear to me why the abstraction of a channel would have a `devRoot` in it). I get why you are doing it, and I don't have a better suggestion -- just thought it was worth pointing out the awkwardness of it.

klueska: Could we alternatively just construct the name, path and hostpath once and store them here (rather than having helper functions for them)?

internal/imex/imex.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Name returns the channel name.
+func (c channel) Name() string {
+	return "channel" + c.id
+}
+
+// Path returns the absolute path to an IMEX channel.
+func (c channel) Path() string {
+	return filepath.Join("/dev/nvidia-caps-imex-channels", c.Name())
+}
```
klueska, line 86: As suggested above, can we just turn all of these into exported fields on the struct (and make it public) rather than have these helper functions?

internal/imex/imex.go
```diff
@@ -0,0 +1,108 @@
...
+
+// HostPath returns the path to the IMEX channel adjusted for the configured devRoot.
+func (c channel) HostPath() string {
+	return filepath.Join(c.devRoot, c.Path())
+}
```
klueska, line 91: As suggested above, can we just turn all of these into exported fields on the struct (and make it public) rather than have these helper functions?

internal/plugin/manager/nvml.go
klueska: Is this just to avoid having to reevaluate all of this in each resource manager? It just feels very out of place alongside the other arguments to this function. Maybe I'm missing something though.

cmd/nvidia-device-plugin/main.go
elezar: I always wonder about this. This does probably make it clearer that it's a repeatable field.

klueska: I see -- because when you pass it its not `--imex-channel-ids=0,1`, its `--imex-channel-ids=0 --imex-channel-ids=1`...

My argument was more for consistence with the envvar of `IMEX_CHANNEL_IDS`. It felt weird to me to have the flag called `imex-channel-id` but then the envvar `IMEX_CHANNEL_IDS`.

internal/imex/imex.go
```diff
@@ -0,0 +1,108 @@
...
+
+// exists checks whether the channel exists at the host path.
+func (c channel) exists() (bool, error) {
```
klueska, line 94:
```suggestion
func (c channel) hostPathExists() (bool, error) {
```

api/config/v1/config.go
```diff
@@ -52,6 +53,11 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
 	}
 
 	config.Flags.UpdateFromCLIFlags(c, flags)
+	// TODO: This is currently not at the flags level?
+	// Does this mean that we should move UpdateFromCLIFlags to function off Config?
+	if c.IsSet("imex-channel-ids") {
+		config.Imex.ChannelIDs = c.IntSlice("imex-channel-ids")
+	}
```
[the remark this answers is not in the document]
elezar: Yes, that was my proposal. I would do it as a follow-up / in a different PR though.

internal/plugin/manager/nvml.go
klueska: Maybe we create a new type to pass in here called `rm.Settings`? Of which `ImexChannels` is the only element currently? That way we could at least extend it over time rather than continually adding more arguments to this function.

internal/imex/imex.go
elezar: I think that makes sense. Let me rework this.

internal/cdi/cdi.go
elezar: Yes, basically. We may need to define how we're passing mode-specific options through the API -- such as which channel to select.

cmd/nvidia-device-plugin/main.go
elezar: To be clear, the following are equivalent:
```
--imex-channel-ids=0,1
```
and:
```
--imex-channel-ids=0 --imex-channel-ids=1
```

internal/imex/imex.go
```diff
@@ -0,0 +1,81 @@
...
+			continue
+		}
+		klog.Infof("Selecting IMEX channel %v", channelName)
+		channels = append(channels, &channel)
+	}
+	return channels, nil
+}
+
+// exists checks whether the channel exists at the host path.
+func (c Channel) exists() (bool, error) {
```
klueska, line 67:
```suggestion
func (c Channel) hostPathExists() (bool, error) {
```

elezar: I have updated the implementation and added a more extensive docstring.

klueska: Looks good.

internal/plugin/manager/nvml.go
```diff
@@ -34,7 +34,7 @@ func (m *nvmlmanager) GetPlugins() ([]plugin.Interface, error) {
 
 	var plugins []plugin.Interface
 	for _, r := range rms {
-		plugin, err := plugin.NewNvidiaDevicePlugin(m.config, m.kubeletSocket, r, m.cdiHandler)
+		plugin, err := plugin.NewNvidiaDevicePlugin(m.config, m.kubeletSocket, r, m.cdiHandler, m.imexChannels)
```
klueska, line 37: Same comment as before applies here -- this extra flag seems out of place, but let's iterate on a future PR to figure out how best to encapsulate this.

internal/plugin/server.go
```diff
@@ -497,6 +505,18 @@ func (plugin *NvidiaDevicePlugin) apiDevices() []*pluginapi.Device {
...
+	plugin.updateResponseForDeviceListEnvVar(response, deviceIDs...)
+	var channelIDs []string
+	for _, channel := range plugin.imexChannels {
+		channelIDs = append(channelIDs, channel.ID)
+	}
+	if len(channelIDs) > 0 {
+		response.Envs[spec.ImexChannelEnvVar] = strings.Join(channelIDs, ",")
+	}
+}
+
```
klueska, line 519: Instead of hiding this in a single function, can we just create a separate one for the IMEX injection and call them both above, i.e.:
```
	if plugin.deviceListStrategies.Includes(spec.DeviceListStrategyEnvVar) {
		plugin.updateResponseForDeviceListEnvVar(response, deviceIDs...)
		plugin.updateResponseForImexChannelEnvVar(response, plugin.imexChannels...)
	}
```

internal/plugin/server.go
```diff
@@ -505,13 +525,21 @@ func (plugin *NvidiaDevicePlugin) updateResponseForDeviceListEnvVar(response *pl
...
 		}
 		response.Mounts = append(response.Mounts, mount)
 	}
+	for _, channel := range plugin.imexChannels {
+		mount := &pluginapi.Mount{
+			HostPath:      deviceListAsVolumeMountsHostPath,
+			ContainerPath: filepath.Join(deviceListAsVolumeMountsContainerPathRoot, "imex", channel.ID),
+		}
+		response.Mounts = append(response.Mounts, mount)
+	}
```
klueska, line 542: This is contingent on adding support for this to the toolkit, but I guess there is vendoring needed here -- we just need to ensure we are running on a version of the toolkit that supports this.

What happens if someone is on a new plugin, but an old toolkit? I guess the IMEX channels just won't be injected? That seems fine., I think, so long as there is no error.

internal/plugin/manager/nvml.go
elezar: If you feel strongly about it, I can refactor things before this change, otherwise I think we can clean it up as a follow-up.

internal/plugin/manager/nvml.go
elezar: I spent some time putting together https://github.com/NVIDIA/k8s-device-plugin/pull/991 which refactors the way in which we construct the plugin instances for each resource. I originally had that on top of this PR, but I think it can be reviewed separately from this one and I will rebase as required.

internal/plugin/manager/nvml.go
elezar: See https://github.com/NVIDIA/k8s-device-plugin/pull/991

internal/plugin/server.go
klueska: OK. I think this may cause some churn, but I think its probably OK to say that if you want IMEX support with the plugin you need minimum version X of the toolkit.

klueska (approved):

klueska:
Looking good

PR #989, opened 2024-10-10
[no-relnote] Allow ARCH to be specified for single image

elezar:
This changes allows the image architecture to be specified when building a "native" image. This is usefull when testing accross architectures.


ArangoGutierrez (approved):
thank you, I can shorten my make  uild cmd when testing the plugin

PR #991, opened 2024-10-14
Refactor server contruction

elezar:
These changes simplify device plugin "server" construction. It removes the `manager` abstraction and instead returns a slice of plugins (one for each separate resource) directly.


cmd/nvidia-device-plugin/main.go
```diff
@@ -41,18 +41,24 @@ import (
...
 	c.Name = "NVIDIA Device Plugin"
 	c.Usage = "NVIDIA device plugin for Kubernetes"
 	c.Version = info.GetVersionString()
 	c.Action = func(ctx *cli.Context) error {
-		return start(ctx, c.Flags)
+		return start(ctx, o)
 	}
 
-	c.Flags = []cli.Flag{
+	o.flags = []cli.Flag{
```
klueska, line 61: I would reverse this and set c.flags here and then update o.flags afterwards.

cmd/nvidia-device-plugin/main.go
```diff
@@ -324,12 +338,7 @@ func startPlugins(c *cli.Context, flags []cli.Flag) ([]plugin.Interface, bool, e
 
 	// Get the set of plugins.
 	klog.Info("Retrieving plugins.")
-	kubeletSocket := c.String("kubelet-socket")
-	pluginManager, err := NewPluginManager(infolib, nvmllib, devicelib, kubeletSocket, config)
-	if err != nil {
-		return nil, false, fmt.Errorf("error creating plugin manager: %v", err)
-	}
-	plugins, err := pluginManager.GetPlugins()
+	plugins, err := GetPlugins(infolib, nvmllib, devicelib, config)
```
klueska, line 341: I guess `kubeletSocket` isn't needed anymore to create a plugin?

cmd/nvidia-device-plugin/plugin-manager.go
```diff
@@ -63,20 +54,19 @@ func NewPluginManager(infolib info.Interface, nvmllib nvml.Interface, devicelib
...
-		manager.WithConfig(config),
-		manager.WithFailOnInitError(*config.Flags.FailOnInitError),
-		manager.WithKubeletSocket(kubeletSocket),
-		manager.WithMigStrategy(*config.Flags.MigStrategy),
+	plugins, err := plugin.New(infolib, nvmllib, devicelib,
+		plugin.WithCDIHandler(cdiHandler),
+		plugin.WithConfig(config),
+		plugin.WithDeviceListStrategies(deviceListStrategies),
+		plugin.WithFailOnInitError(*config.Flags.FailOnInitError),
 	)
```
klueska, line 62: Related to the comment above -- `kubeletSocket` isn't needed here anymore?

internal/plugin/server.go
```diff
@@ -69,41 +66,25 @@ type NvidiaDevicePlugin struct {
...
 
-	plugin := NvidiaDevicePlugin{
+	plugin := nvidiaDevicePlugin{
 		rm:                   resourceManager,
-		config:               config,
-		deviceListEnvVar:     "NVIDIA_VISIBLE_DEVICES",
-		deviceListStrategies: deviceListStrategies,
-		socket:               pluginPath + ".sock",
-		cdiHandler:           cdiHandler,
-		cdiAnnotationPrefix:  *config.Flags.Plugin.CDIAnnotationPrefix,
```
klueska, line 101: We originally included the `deviceListEnvVar` here in case there was a *different* envvar that we needed to use to pass the device info back to the runtime. In practice this was never necessary, so it should be fine to remove this level of configurability.

internal/plugin/server.go
```diff
@@ -113,27 +94,53 @@ func NewNvidiaDevicePlugin(config *spec.Config, kubeletSocket string, resourceMa
...
 // Devices returns the full set of devices associated with the plugin.
-func (plugin *NvidiaDevicePlugin) Devices() rm.Devices {
+func (plugin *nvidiaDevicePlugin) Devices() rm.Devices {
 	return plugin.rm.Devices()
 }
 
 // Start starts the gRPC server, registers the device plugin with the Kubelet,
 // and starts the device healthchecks.
-func (plugin *NvidiaDevicePlugin) Start() error {
+func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
```
klueska, line 143: I see -- its' not stored, but rather passed in to `Start()` now...

internal/plugin/server.go
```diff
@@ -69,41 +66,25 @@ type NvidiaDevicePlugin struct {
...
-			}
-		}
-		mpsDaemon = mps.NewDaemon(resourceManager, mps.ContainerRoot)
-		mpsHostRoot = mps.Root(*config.Flags.CommandLineFlags.MpsRoot)
+// devicePluginForResource creates a device plugin for the specified resource.
+func (o *options) devicePluginForResource(resourceManager rm.ResourceManager) (Interface, error) {
+	mpsDaemon, mpsHostRoot, err := o.getMPSDaemon(resourceManager)
+	if err != nil {
+		return nil, err
 	}
```
klueska, line 74: Can we move the if check for `o.config.Sharing.SharingStrategy() != spec.SharingStrategyMPS ` out here to make it clear that we only actually set these variables to anything if MPS is configured. Alternatively, rename the function something like `getMPSDaemonIfConfigured()`? As it reads now it seems as if we are *always* getting an MPS server setup for the plugin.

internal/plugin/factory.go
```diff
@@ -26,84 +26,104 @@ import (
...
 
-	kubeletSocket string
+	deviceListStrategies spec.DeviceListStrategies
 }
 
-// New creates a new plugin manager with the supplied options.
-func New(infolib info.Interface, nvmllib nvml.Interface, devicelib device.Interface, opts ...Option) (Interface, error) {
-	m := &manager{
+// New a new set of plugins with the supplied options.
+func New(infolib info.Interface, nvmllib nvml.Interface, devicelib device.Interface, opts ...Option) ([]Interface, error) {
```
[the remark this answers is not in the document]
elezar: Yes, maybe using a plural in the package name would be clearer.

Note that the return type is `[]Interface` though, so not sure how that reads here.

internal/plugin/server.go
elezar: I would rather clean this up here and readd it if we see that we need it. This struct was getting quite big already and I wanted to trim it down.

We also don't allow this level of configurability for other envvars.

internal/plugin/server.go
elezar: Yes, this is how I had it for the original change to allow an empty string for testing basic functionality without a full cluster. After a round of reviews, I think @tariq1890 suggested we make it a property instead. After looking at this file again from the perspective of this cleanup, I am back to thinking that it makes more sense to pass it as an argument.

internal/plugin/server.go
elezar: I'm more partial to the latter. I've reworked it a bit by introducing an `mpsOptions` type and moved the MPS-specifics to a separate file.

We now set `enabled` on this type specifically instead of having to query the sharing strategy all the time.

cmd/nvidia-device-plugin/main.go
klueska: I don't feel too strongly, but my preference would be to keep it as a property of each plugin. I guess that is complicated by the fact that we don't instantiate each plugin from this package anymore (not that we would ever more than one path for this anyway).

klueska: Like I said, no strong opinion though -- I can see the argument both ways...

cmd/nvidia-device-plugin/main.go
klueska: I see, so flags defined before even defining c.App and then grouping all of the c.X assignments together. Yeah, that seems fine. The only thing we would "lose" is the niceness of having this near the top of the function:
```
	c.Name = "NVIDIA Device Plugin"
	c.Usage = "NVIDIA device plugin for Kubernetes"
```
but that is minor

internal/plugin/server.go
```diff
@@ -42,68 +41,47 @@ import (
...
-		for _, device := range resourceManager.Devices() {
-			if device.IsMigDevice() {
-				return nil, errors.New("sharing using MPS is not supported for MIG devices")
-			}
-		}
-		mpsDaemon = mps.NewDaemon(resourceManager, mps.ContainerRoot)
-		mpsHostRoot = mps.Root(*config.Flags.CommandLineFlags.MpsRoot)
+// devicePluginForResource creates a device plugin for the specified resource.
+func (o *options) devicePluginForResource(resourceManager rm.ResourceManager) (Interface, error) {
+	mpsOptions, err := o.getMPSOptionsIfConfigured(resourceManager)
```
klueska, line 69: actually -- with it refactored like this, I think we can drop the `IfConfigured` since now you are getting back a set of options, which will indicate if it is enabled or not inside of it.

internal/plugin/server.go
klueska: Your choice. I don't feel strongly.

cmd/nvidia-device-plugin/main.go
elezar: After looking at the changes required, I have decided to go with your suggestion.

internal/plugin/server.go
elezar: I prefer the way it is now. If we see that we need to connect to mulitple kubelet sockets, we can make that change later.

internal/plugin/mps.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type mpsOptions struct {
+	enabled      bool
+	resourceName spec.ResourceName
+	daemon       *mps.Daemon
+	hostRoot     mps.Root
+}
+
+// getMPSOptions returns the MPS options specified resource manager.
```
klueska, line 38:
```suggestion
// getMPSOptions returns the MPS options specified for the resource manager.
```

internal/plugin/factory.go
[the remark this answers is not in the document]
klueska: Its fine -- we can revisit later if we want

internal/plugin/mps.go
elezar: Updated.

klueska (approved):

PR #996, opened 2024-10-17
Fix IMEX injection when passing device nodes

elezar:
This change fixes an incorrect HostPath for IMEX channels when passing device specs as part of the ContainerAllocateResponse.

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -529,8 +529,14 @@ func (plugin *nvidiaDevicePlugin) apiDeviceSpecs(devRoot string, ids []string) [
 	for _, channel := range plugin.imexChannels {
 		spec := &pluginapi.DeviceSpec{
 			ContainerPath: channel.Path,
-			HostPath:      channel.HostPath,
-			Permissions:   "rw",
+			// TODO: The HostPath property for a channel is not the correct value to use here.
+			// The `devRoot` there represents the devRoot in the current container when discovering devices
+			// and is set to "{{ .*config.Flags.Plugin.ContainerDriverRoot }}/dev".
+			// The devRoot in this context is the {{ .config.Flags.NvidiaDevRoot }} and defines the
+			// root for device nodes on the host. This is usually / or /run/nvidia/driver when the
+			// driver container is used.
+			HostPath:    filepath.Join(devRoot, channel.Path),
+			Permissions: "rw",
 		}
 		specs = append(specs, spec)
 	}


internal/imex/imex.go
```diff
@@ -46,8 +46,10 @@ func GetChannels(config *spec.Config, devRoot string) (Channels, error) {
 		channelName := "channel" + id
 		path := filepath.Join("/dev/nvidia-caps-imex-channels", channelName)
 		channel := Channel{
-			ID:       id,
-			Path:     path,
+			ID:   id,
+			Path: path,
+			// TODO: HostPath is not the correct name for this property.
+			// We should deal with this consistently.
 			HostPath: filepath.Join(devRoot, path),
```
klueska, line 53: Is this a different `devRoot` or is this a different `path`? At a superficial glance I don't see how the value for the change in [internal/plugin/server.go](https://github.com/NVIDIA/k8s-device-plugin/pull/996/files#diff-1855ca77ecb9dcdd1c59676ab0ad69676b1c219d76484174bbec4187252de5f2) is different than this...

elezar: The only difference here is the comment. I can update the comment to be clearer.

elezar: expanded the comment and moved it to where we made the change.

klueska: I understand that the comment is the only difference here -- what I'm trying to understand is how this (already existing assignment):
```
HostPath: filepath.Join(devRoot, path)
```
is differnt from the updated assignement we are doing in [internal/plugin/server.go](https://github.com/NVIDIA/k8s-device-plugin/pull/996/files#diff-1855ca77ecb9dcdd1c59676ab0ad69676b1c219d76484174bbec4187252de5f2):
```
HostPath:    filepath.Join(devRoot, channel.Path),
```
Is it a different `devRoot` is the `path` in the first case somehow different than `channel.Path` in the second (even though `channel.Path` seems to just be assigned to the same `path` that is passed in.

klueska: Got it -- the new comment clarifies...

klueska (approved):

internal/imex/imex.go
elezar: As I called out offline and in the updated comment. The `devRoot` here and in the device node injection code in server.go are different paths. The `devRoot` in our updated assignment is set from the `NVIDIA_DRIVER_ROOT` (`/`) wherease the value here is `/host` which is the driver root in the CONTAINER.

PR #965, opened 2024-09-24
Enable labels for IMEX Domain and Clique

ArangoGutierrez:

PR #997, opened 2024-10-17
use uuid.NewSHA1 package to generate uuids

tariq1890:
~Here is a [post](https://go.dev/blog/randv2) that goes into detail on the security limitations of the `math/rand` package.~

~Since the IMEX changes haven't publicised yet, I thought this would be a safe starting point to implement this change~

### DIFF
--- a/internal/lm/fabric.go
+++ b/internal/lm/fabric.go
@@ -20,17 +20,16 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"math/rand" // nolint:gosec
 	"net"
 	"os"
 	"sort"
 	"strings"
 
-	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
-	"github.com/NVIDIA/k8s-device-plugin/internal/resource"
-
 	"github.com/google/uuid"
 	"k8s.io/klog/v2"
+
+	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
+	"github.com/NVIDIA/k8s-device-plugin/internal/resource"
 )
 
 func newImexLabeler(config *spec.Config, devices []resource.Device) (Labeler, error) {
@@ -142,18 +141,5 @@ func getImexDomainID(r io.Reader) (string, error) {
 }
 
 func generateContentUUID(seed string) string {
-	// nolint:gosec
-	rand := rand.New(rand.NewSource(hash(seed)))
-	charset := make([]byte, 16)
-	rand.Read(charset)
-	uuid, _ := uuid.FromBytes(charset)
-	return uuid.String()
-}
-
-func hash(s string) int64 {
-	h := int64(0)
-	for _, c := range s {
-		h = 31*h + int64(c)
-	}
-	return h
+	return uuid.NewSHA1(uuid.Nil, []byte(seed)).String()
 }
--- a/internal/lm/fabric_test.go
+++ b/internal/lm/fabric_test.go
@@ -32,7 +32,7 @@ func TestGerenerateDomainUUID(t *testing.T) {
 		{
 			description: "single IP",
 			ips:         []string{"10.130.3.24"},
-			expected:    "4dbd3d31-fbb3-8a40-33bb-bcc0dd7b68b8",
+			expected:    "60ad7226-0130-54d0-b762-2a5385a3a26f",
 		},
 		{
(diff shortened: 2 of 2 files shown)


tariq1890:
It looks like CodeQL runs on an earlier version of Go

ArangoGutierrez (approved):
lgtm
@elezar  ?

klueska:
Have you tested to ensure that the generated UUID is the same for a constant seed value?

tariq1890:
I haven't had the chance to test this on a running k8s cluster. I'll keep you posted on that.

However, I tried this sample go programme and I was able to confirm that UUID was generated deterministically for a given seed

```
func main() {

	ips := []string{
		"10.130.3.24",
		"10.130.3.53",
		"10.130.3.23",
		"10.130.3.31",
		"10.130.3.27",
		"10.130.3.25",
	}
	sort.Strings(ips)

	seed := strings.Join(ips, "\n")
	r := rand.NewChaCha8(sha256.Sum256([]byte(seed)))
	charset := make([]byte, 16)
	_, err := r.Read(charset)
	if err != nil {
		fmt.Printf("Error generating random number: %v", err)
	}
	u, _ := uuid.FromBytes(charset)

	fmt.Printf("Output: %s\n", u.String())
}
```

internal/lm/fabric.go
```diff
@@ -142,18 +143,16 @@ func getImexDomainID(r io.Reader) (string, error) {
 }
 
 func generateContentUUID(seed string) string {
```
elezar, line 145: Before changing this, we should add some tests for expected outputs so that we understand the implications of these changes.

elezar:
We are not using these random numbers to generate a cryptographically secure random number -- we just want a unique idenfiier for a set of input IPs. I'm happy to make this change, but let's add tests that capture the behaviour since we can't have IMEX domain IDs changing across device plugin updates.

internal/lm/fabric.go
elezar: https://github.com/NVIDIA/k8s-device-plugin/pull/999

elezar (changes requested):
If we're going to make a change such as this we should use the `uuid.NewSHA1` or `uuid.NewMD5` functions instead of rolling our own.

This was called out in the review of the original code in #965

internal/lm/fabric.go
```diff
@@ -142,18 +143,16 @@ func getImexDomainID(r io.Reader) (string, error) {
...
-	// nolint:gosec
-	rand := rand.New(rand.NewSource(hash(seed)))
+	r := rand.NewChaCha8(hash(seed))
```
elezar, line 146: Would it make sense to use something like:
```
uudi.NewSHA1("", []bytes(seed)).String()
```
to replace all this logic here? We can decide if we want to make the "namespace" configurable.

tariq1890: Sure, I am all for removing code and making it simpler

.github/workflows/code_scanning.yaml
```diff
@@ -38,6 +38,15 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
+    - name: Get Golang version
```
tariq1890, line 41: @elezar I am creating a separate PR for this change

internal/lm/fabric.go
tariq1890: Thanks

internal/lm/fabric_test.go
```diff
@@ -32,7 +32,7 @@ func TestGerenerateDomainUUID(t *testing.T) {
 		{
 			description: "single IP",
 			ips:         []string{"10.130.3.24"},
-			expected:    "4dbd3d31-fbb3-8a40-33bb-bcc0dd7b68b8",
+			expected:    "60ad7226-0130-54d0-b762-2a5385a3a26f",
```
elezar, line 35: @klueska @ArangoGutierrez are we OK with this "breaking" change given that we haven't released this yet.

I think the simplifications in this PR are worth it.

klueska: yes

tariq1890: I've rebased the PR. Requesting another round of review

elezar (approved):
Thanks @tariq1890

PR #1007, opened 2024-10-21
[no-relnote] Temporarly ignore NFD minor versions at Depdendabot

ArangoGutierrez:


ArangoGutierrez:
Added the ignore to the `release-*` section as well

PR #1009, opened 2024-10-22
Search driver root for IMEX node configs

elezar:


cmd/gpu-feature-discovery/main.go
```diff
@@ -86,11 +86,16 @@ func main() {
...
 		},
-		&cli.StringFlag{
+		&cli.StringSliceFlag{
 			Name:    "imex-nodes-config-file",
+			Aliases: []string{"imex-nodes-config-files"},
 			Usage:   "Path to the IMEX nodes config file. This file contains a list of IP addresses of the nodes in the IMEX domain.",
-			Value:   "/etc/nvidia-imex/nodes_config.cfg",
-			EnvVars: []string{"GFD_IMEX_NODES_CONFIG_FILE"},
+			Value: cli.NewStringSlice(
+				"/etc/nvidia-imex/nodes_config.cfg",
```
tariq1890, line 94: I would prefer constructing these paths dynamically based on the `DRIVER_ROOT` instead of hardcoding them

cdesiniotis: The container path will always be either `/driver-root` or `/host`

tariq1890: But that's an assumption we are injecting into the code. It could always change in the future. When we make this path dependent on a configurable value, it is most robust and future-proof IMO

elezar: Could you expand on what you mean? These paths are container local and as such they are not the same as the driverRoot or (hostRoot for that matter).

tariq1890: > The container path will always be either /driver-root or /host

This path exists because of the Kubernetes host mount, yes? It is external to the application, and it should only be passed in as an input to the application. Other than that, the application shouldn't have k8s details hardwired into it. It becomes a leaky abstraction IMO

api/config/v1/imex.go
```diff
@@ -39,6 +39,10 @@ type Imex struct {
 	// If it is not required its injection is skipped if the device nodes do not exist or if its
 	// existence cannot be queried.
 	Required bool `json:"required,omitempty" yaml:"required,omitempty"`
+	// NodesConfigFile defines the locations to IMEX node config files.
```
cdesiniotis, line 42:
```suggestion
	// NodesConfigFile defines the location to the IMEX nodes config file.
```

api/config/v1/imex.go
```diff
@@ -39,6 +39,10 @@ type Imex struct {
...
+	// Such a nodes config file contains the IP addresses of nodes that are part of the IMEX domain.
+	// Note that this is the absolute path to the file in the device plugin container.
+	NodesConfigFile *string `json:"nodesConfigFiles,omitempty" yaml:"nodesConfigFiles,omitempty"`
```
cdesiniotis, line 45:
```suggestion
	NodesConfigFile *string `json:"nodesConfigFile,omitempty" yaml:"nodesConfigFile,omitempty"`
```

cdesiniotis (approved):

tariq1890 (approved):

cdesiniotis:
Verified on a IMEX enabled cluster using the gpu-operator changes from https://github.com/NVIDIA/gpu-operator/pull/1055. 
```
cdesiniotis@cnt-imex-devel:~$ kubectl logs -n gpu-operator gpu-feature-discovery-9qfnt
Defaulted container "gpu-feature-discovery" out of: gpu-feature-discovery, toolkit-validation (init)
IS_HOST_DRIVER=true
NVIDIA_DRIVER_ROOT=/
DRIVER_ROOT_CTR_PATH=/host
NVIDIA_DEV_ROOT=/
DEV_ROOT_CTR_PATH=/host

. . . snip . . .

I1022 22:42:02.452695       1 fabric.go:56] Using labeler for IMEX config /host/etc/nvidia-imex/nodes_config.cfg
```

PR #1010, opened 2024-10-23
Add init container to handle imex nodes config mount

elezar:
This change adds an init container to GFD that checks whether the /driver-root/etc/nvidia-imex exits and mounts it to /etc/nvidia-imex in the /config volume. This allows the main GFD container access to this config without requiring that the whole driver root be mounted.


deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,27 @@ spec:
...
+        command:  |-
+          #!/bin/bash
+
+          if [[ ! -d /driver-root/etc/nvidia-imex ]]; then
+            echo "No IMEX Nodes config path detected; Skipping"
+            exit 0
+          fi
+
+          mkdir -p /config/etc/nvidia-imex
+          mount /driver-root/etc/nvidia-imex /config/etc/nvidia-imex
```
elezar, line 73: I first thought to rather mount `/driver-root/etc/nvidia-imex` to /config/imex` instead.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -201,10 +221,10 @@ spec:
         - name: host-sys
           hostPath:
             path: "/sys"
-        - name: nvidia-imex-dir
-          type: DirectoryOrCreate
+        - name: driver-root
           hostPath:
-            path: {{ clean ( join "/" ( list "/" .Values.nvidiaDriverRoot "/etc/nvidia-imex" ) ) | quote }}
+            path: {{ clean ( join "/" ( list "/" .Values.nvidiaDriverRoot ) ) | quote }}
```
elezar, line 226: This mounts either `/` or `/run/nvidia/driver` (for example).

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,31 @@ spec:
...
+          mount --bind /driver-root/etc/nvidia-imex /config/etc/nvidia-imex
+        volumeMounts:
+          - name: config
+            mountPath: /config
+            mountPropagation: Bidirectional
+          - name: driver-root
+            mountPath: /driver-root
+            readOnly: true
+        securityContext:
+          privileged: true
```
klueska, line 83: If you make this the second init container does it remove the need for two separate ` {{- if $options.hasConfigMap }}` checks?

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,31 @@ spec:
...
+            echo "No IMEX Nodes config path detected; Skipping"
+            exit 0
+          fi
+
+          mkdir -p /config/etc/nvidia-imex
+          mount --bind /driver-root/etc/nvidia-imex /config/etc/nvidia-imex
+        volumeMounts:
+          - name: config
+            mountPath: /config
+            mountPropagation: Bidirectional
```
klueska, line 78: Why does this need to be bidirectional? At worst I think we only need `HostToContainer `-- but I'm curious if we even care about future mounts in either direction?

elezar: The issue I was seeing was that the mount made in the init container was not visible to other containers using the empty volume. Maybe there is a better way to ensure that the mounts are shared than propagating this back to the empty volume on the host.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: I don't think so due to the `shareProcessNamespace: true` that is required for the config sidecare container.

klueska: I see -- yeah, it's because you need to unconditionally create the initContainers section between setting shareProcessNamespace: true and defining the config init container.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
klueska: OK, yeah, that makes sense. The init container creates the mount, but if this mount is not reflected back on the host, then when it exits, the mount will be removed and never visible to the app container when it eventually starts up. Let's leave it this way for now and think a bit more about this for future releases.

klueska (dismissed):

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: The one issue that I did see was that the mount was not cleaned up. What is the best practice for doing this? (we don't do this for MPS either ... )

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,31 @@ spec:
       {{- end }}
       {{- if $options.hasConfigMap }}
       shareProcessNamespace: true
+      {{- end }}
       initContainers:
+      - image: ubuntu
```
elezar, line 62: @cdesiniotis @tariq1890 @klueska we're using an `ubuntu` image here because we don't have a `mount` command in the reduced ubi9 image that we use.

klueska: Do we remove it? It seems to be in the standard ubi9 image:
```
$ docker run --rm -it registry.access.redhat.com/ubi8/ubi mount --help

Usage:
 mount [-lhV]
 mount -a [options]
 mount [options] [--source] <source> | [--target] <directory>
 mount [options] <source> <directory>
 mount <operation> <mountpoint> [<target>]

Mount a filesystem.

Options:
 -a, --all               mount all filesystems mentioned in fstab
 -c, --no-canonicalize   don't canonicalize paths
 -f, --fake              dry run; skip the mount(2) syscall
 -F, --fork              fork off for each device (use with -a)
 -T, --fstab <path>      alternative file to /etc/fstab
 -i, --internal-only     don't call the mount.<type> helpers
 -l, --show-labels       show also filesystem labels
 -n, --no-mtab           don't write to /etc/mtab
 -o, --options <list>    comma-separated list of mount options
 -O, --test-opts <list>  limit the set of filesystems (use with -a)
 -r, --read-only         mount the filesystem read-only (same as -o ro)
 -t, --types <list>      limit the set of filesystem types
     --source <src>      explicitly specifies source (path, label, uuid)
     --target <target>   explicitly specifies mountpoint
 -v, --verbose           say what is being done
 -w, --rw, --read-write  mount the filesystem read-write (default)

 -h, --help              display this help
 -V, --version           display version

Source:
 -L, --label <label>     synonym for LABEL=<label>
 -U, --uuid <uuid>       synonym for UUID=<uuid>
 LABEL=<label>           specifies device by filesystem label
 UUID=<uuid>             specifies device by filesystem UUID
 PARTLABEL=<label>       specifies device by partition label
 PARTUUID=<uuid>         specifies device by partition UUID
 <device>                specifies device by path
 <directory>             mountpoint for bind mounts (see --bind/rbind)
 <file>                  regular file for loopdev setup

Operations:
 -B, --bind              mount a subtree somewhere else (same as -o bind)
 -M, --move              move a subtree to some other place
 -R, --rbind             mount a subtree and all submounts somewhere else
 --make-shared           mark a subtree as shared
 --make-slave            mark a subtree as slave
 --make-private          mark a subtree as private
 --make-unbindable       mark a subtree as unbindable
 --make-rshared          recursively mark a whole subtree as shared
 --make-rslave           recursively mark a whole subtree as slave
 --make-rprivate         recursively mark a whole subtree as private
 --make-runbindable      recursively mark a whole subtree as unbindable

For more details see mount(8).
```

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
klueska: Where does it live after the pod has been shut down? I would assume the shutdown of the pod and the removal of the temporary space for the mounted volume would have cleaned it up.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
klueska: Got it -- the minimal image doesnt have it:
```
$ docker run --rm -it registry.access.redhat.com/ubi9-minimal mount --help
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "mount": executable file not found in $PATH: unknown.
```

klueska: Can we use `registry.access.redhat.com/ubi9/ubi` instead of ubuntu so that at least we are using (sort of) the same base image for everything?

deployments/container/cleanup/cleanup.sh
```diff
@@ -52,6 +52,10 @@ microdnf remove -y \
 microdnf remove -y \
     $(rpm -qa | sort | grep -v -f package-names.minimal -e gpg-pubkey)
 
+# We need mount for our init container and install it here.
+# TODO: We need to provide a better way to install this / skip the cleanup of this package.
+microdnf install --assumeyes util-linux-core
```
klueska, line 57: I don't think we want to be pulling anything dynamically from the internet here. People can (and do) try to run this component in an airgapped environment.

elezar: the following are already in the `minimal` image:
```
$ cat package-list.minimal | grep openssl
openssl-fips-provider-3.0.7-2.el9.x86_64
openssl-libs-3.0.7-28.el9_4.x86_64
```

elezar: This is the full diff between the packages in the `ubi9-minimal` image and the one we construct here:
```
$ diff package-list.minimal package-list.cleaned
19,21c19,21
< glibc-2.34-100.el9_4.3.x86_64
< glibc-common-2.34-100.el9_4.3.x86_64
< glibc-minimal-langpack-2.34-100.el9_4.3.x86_64
---
> glibc-2.34-100.el9_4.4.x86_64
> glibc-common-2.34-100.el9_4.4.x86_64
> glibc-minimal-langpack-2.34-100.el9_4.4.x86_64
46d45
< libdnf-0.69.0-8.el9_4.1.x86_64
57d55
< libpeas-1.30.0-4.el9.x86_64
70d67
< libusbx-1.0.26-1.el9.x86_64
79d75
< microdnf-3.9.1-3.el9.x86_64
104a101
> util-linux-core-2.37.4-18.el9.x86_64
```
so a slightly newer glibc packge version (would have to check the reason for this) and the addition of `util-linux-core`. Everything else is either removed or the same.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: Running `mount | grep config` in the kind worker node container:
```
 mount  | grep config
configfs on /sys/kernel/config type configfs (rw,nosuid,nodev,noexec,relatime)
overlay on /var/lib/kubelet/pods/406054da-7ef0-48f1-a4ad-1da282913dc7/volumes/kubernetes.io~empty-dir/config/etc/nvidia-imex type overlay (ro,relatime,lowerdir=/var/lib/docker/overlay2/l/YJ3LUQCUVC2MXFPQQU5LOFINP2:/var/lib/docker/overlay2/l/NJLKU2BR3I52CNQBK6SCGI74WJ:/var/lib/docker/overlay2/l/BXAROID4RF5JY7PFABBT42PHK4,upperdir=/var/lib/docker/overlay2/f94a9595453a2c4a09dfaf3720ef1beb376620eec59443e7cbb4743e1a9d59ef/diff,workdir=/var/lib/docker/overlay2/f94a9595453a2c4a09dfaf3720ef1beb376620eec59443e7cbb4743e1a9d59ef/work,xino=off)
```

immediately after deleting the pod we see:
```
nvidia-device-plugin   nvidia-device-plugin-gpu-feature-discovery-wfh2f                  0/1     Terminating   0               29m
```
and
```
mount  | grep config
configfs on /sys/kernel/config type configfs (rw,nosuid,nodev,noexec,relatime)
overlay on /var/lib/kubelet/pods/406054da-7ef0-48f1-a4ad-1da282913dc7/volumes/kubernetes.io~empty-dir/config/etc/nvidia-imex type overlay (ro,relatime,lowerdir=/var/lib/docker/overlay2/l/YJ3LUQCUVC2MXFPQQU5LOFINP2:/var/lib/docker/overlay2/l/NJLKU2BR3I52CNQBK6SCGI74WJ:/var/lib/docker/overlay2/l/BXAROID4RF5JY7PFABBT42PHK4,upperdir=/var/lib/docker/overlay2/f94a9595453a2c4a09dfaf3720ef1beb376620eec59443e7cbb4743e1a9d59ef/diff,workdir=/var/lib/docker/overlay2/f94a9595453a2c4a09dfaf3720ef1beb376620eec59443e7cbb4743e1a9d59ef/work,xino=off)
overlay on /var/lib/kubelet/pods/3a948d67-b23b-41d3-a172-18ec8b37e7ad/volumes/kubernetes.io~empty-dir/config/etc/nvidia-imex type overlay (ro,relatime,lowerdir=/var/lib/docker/overlay2/l/YJ3LUQCUVC2MXFPQQU5LOFINP2:/var/lib/docker/overlay2/l/NJLKU2BR3I52CNQBK6SCGI74WJ:/var/lib/docker/overlay2/l/BXAROID4RF5JY7PFABBT42PHK4,upperdir=/var/lib/docker/overlay2/f94a9595453a2c4a09dfaf3720ef1beb376620eec59443e7cbb4743e1a9d59ef/diff,workdir=/var/lib/docker/overlay2/f94a9595453a2c4a09dfaf3720ef1beb376620eec59443e7cbb4743e1a9d59ef/work,xino=off)
```
running `umount` on the worker node allows terminations to continue / succeed.

deployments/container/cleanup/cleanup.sh
klueska: OK, I'm fine with this change if you are then.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
klueska: Should we just go back to a copying approach instead of a mounting? It means we will require a restart of the plugin if the config file changes on the host, but this is probably OK for this first iteration.

deployments/container/cleanup/cleanup.sh
tariq1890: It took me way longer than I'd like to admit to understand what `assumeyes` meant. I kept reading it as "assume eyes". Why this flag wasn't named `--assume-yes` instead is beyond me

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,31 @@ spec:
...
       shareProcessNamespace: true
+      {{- end }}
       initContainers:
+      - image: {{ include "nvidia-device-plugin.fullimage" . }}
+        name: gpu-feature-discovery-imex-init
+        command:
+        - "/bin/bash"
+        - "-c"
+        - |
+          if [[ ! -d /driver-root/etc/nvidia-imex ]]; then
```
cdesiniotis, line 68: Question: How would this look like in the GPU Operator, where driver root could be `/driver-root` OR `/host`? Would we simply check both locations for the presence of `/etc/nvidia-imex`?

deployments/container/cleanup/cleanup.sh
tariq1890, line 57: Non-blocking. Only because `--assumeyes` is such an eyesore
```suggestion
microdnf install -y util-linux-core
```

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
tariq1890: Shouldn't we be using this?

```suggestion
          if [[ ! -d {{ .Values.nvidiaDriverRoot }}/etc/nvidia-imex ]]; then
```

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: It's longer than 30s. I have one that's in terminating for more than 6 hours.

It could be that using the mounting logic that we have for the mps daemon doesn't have this issue since there is more complex logic being applied?

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
cdesiniotis: No. The container path will always be `/driver-root`.

elezar: @tariq1890 no, we mount either `{{ .Values.nvidiaDriverRoot }}` or `/` to `/driver-root` in the container. This means that in the container we should always use `/driver-root`.

deployments/container/cleanup/cleanup.sh
elezar: Addressed in #1013

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: For what it's worth, I'm happy to resort to copying and require a restart, alternatively we can do something like:
```
diff --git a/deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml b/deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
index cb3f9d36..fd1ffb8a 100644
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
@@ -214,6 +214,9 @@ spec:
         {{- end }}
           - name: config
             mountPath: /config
+          - name: driver-root-etc
+            mountPath: /driver-root/etc
+            readOnly: true
         {{- with .Values.resources }}
         resources:
           {{- toYaml . | nindent 10 }}
@@ -225,9 +228,9 @@ spec:
         - name: host-sys
           hostPath:
             path: "/sys"
-        - name: driver-root
+        - name: driver-root-etc
           hostPath:
-            path: {{ clean ( join "/" ( list "/" .Values.nvidiaDriverRoot ) ) | quote }}
+            path: {{ clean ( join "/" ( list "/" .Values.nvidiaDriverRoot "etc" ) ) | quote }}
       {{- if $options.hasConfigMap }}
         - name: available-configs
           configMap:
```
but this doesn't address the operator case?

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
cdesiniotis: My point is that in the operator use case, we will mount both `/` --> `/host` and `/run/nvidia/driver` --> `/driver-root`. I am assuming we need to update this script in the operator to look for the `/etc/nvidia-imex` directory at both locations? Or just reference the `DRIVER_ROOT_CTR_PATH` variable...

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
cdesiniotis: It could work in the operator case if we updated both the host-root and drive-root mounts to `/etc` --> `/host/etc` and `/run/nvidia/driver/etc` --> `/driver-root/etc` respectively.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: Note that the code still checks both `${GFD_IMEX_NODES_CONFIG_FILE}` (with a default value of `/etc/nvidia-imex/nodes_config.cfg`) and `${DRIVER_ROOT_CTR_PATH}/${GFD_IMEX_NODES_CONFIG_FILE}`, so as long as the operator mounts the file and sets the `DRIVER_ROOT_CTR_PATH` envvar accordingly it should work.

elezar: @cdesiniotis since we're using `DRIVER_ROOT_CTR_PATH` things should work since the validator sets this to `/host` for a host-installed driver here: https://github.com/NVIDIA/gpu-operator/blob/040d1539f6a8e2e114a9f4a4141e6e35cc6ca298/validator/main.go#L841

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,35 @@ spec:
...
+          fi
+
+          if [[ ! -f /driver-root/etc/nvidia-imex/nodes_config.cfg ]]; then
+            echo "No IMEX nodes config path detected; Skipping"
+            exit 0
+          fi
+
+          echo "Copying IMEX nodes config"
+          mkdir -p /config/etc/nvidia-imex
+          cp /driver-root/etc/nvidia-imex/nodes_config.cfg /config/etc/nvidia-imex/nodes_config.cfg
```
klueska, line 80: Can we make `/etc/nvidia-imex/nodes_config.cfg` a local envvar to this script instead of repeating it everywhere.

internal/lm/imex.go
```diff
@@ -64,6 +61,19 @@ func newImexLabeler(config *spec.Config, devices []resource.Device) (Labeler, er
 	return empty{}, nil
 }
 
+// imexNodesConfigFilePathSearchRoots returns a list of roots to search for the IMEX nodes config file.
+func imexNodesConfigFilePathSearchRoots(config *spec.Config) []string {
+	// By default se search / and /config for config files.
```
tariq1890, line 66:
```suggestion
	// By default, search / and /config for config files.
```

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: Done in latest.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,35 @@ spec:
...
+            rm -f /config/${IMEX_NODES_CONFIG_FILE}
+          fi
+
+          if [[ ! -f /driver-root/${IMEX_NODES_CONFIG_FILE} ]]; then
+            echo "No IMEX nodes config path detected; Skipping"
+            exit 0
+          fi
+
+          echo "Copying IMEX nodes config"
+          mkdir -p $(dirname /config//config/${IMEX_NODES_CONFIG_FILE})
```
cdesiniotis, line 79: Is this a typo? `/config//config`?

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -57,7 +57,35 @@ spec:
...
+          cp /driver-root/${IMEX_NODES_CONFIG_FILE} /config/${IMEX_NODES_CONFIG_FILE}
+        volumeMounts:
+        - name: config
+          mountPath: /config
+        - name: driver-root
+          mountPath: /driver-root/etc
+          subPath: etc
+          readOnly: true
+      {{- if $options.hasConfigMap }}
```
cdesiniotis, line 88: Why was this line added?

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
cdesiniotis: Got it.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: Yes.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
elezar: It wasn't added, it was essentially moved so that the config init container is guarded by this conditional. Note that the sharedprocessNamespace is also guarded by the same option.

cdesiniotis: Okay, I see it now. Thanks.

cdesiniotis (approved):

tariq1890 (approved):

tariq1890:
Just a thought (not blocking) , it feels a bit much to have the path below:

```
config/etc/nvidia-imex/nodes_config.cfg
```

Since `config/` is already the designated config directory of the GFD application, it would just be enough to copy the file from `/driver-root/etc/nvidia-imex/nodes_config.cfg` over to a file with a path like `/config/nvidia-imex-nodes_config.cfg`

klueska:
> Just a thought (not blocking) , it feels a bit much to have a file with the path below:
> 
> ```
> config/etc/nvidia-imex/nodes_config.cfg
> ```
> 
> Since `config/` is already the designated config directory of the GFD application, it would just be enough to copy the file from `/driver-root/etc/nvidia-imex/nodes_config.cfg` over to a file with a path like `/config/nvidia-imex-nodes-config.cfg` . Or even `/config/nvidia-imex/nodes_config.cfg`
> 
> The `/etc` part seems a bit redundant.

It's done the way it is so that the filepath is the same for all search directores whether in `/host`, `/driver-root` or `/config`

klueska (approved):

PR #1018, opened 2024-10-27
bugfix: Allow for optional TimeSlicing configuration

danielkleinstein:
This fixes a vestigial bug from the introduction of MPS. Now that two timeslicing configurations are available, it's possible for MPS to be configured instead of timeslicing.

But since the TimeSlicing field was made non-optional - its existence is forced when unmarshalling, and then the parsing fails because no resources are specified under the empty TimeSlicing field.

### DIFF
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -96,9 +96,13 @@ func DisableResourceNamingInConfig(logger logger, config *Config) {
 	config.Resources.MIGs = nil
 
 	// Disable renaming / device selection in Sharing.TimeSlicing.Resources
-	config.Sharing.TimeSlicing.disableResoureRenaming(logger, "timeSlicing")
+	if (config.sharing.TimeSlicing != nil) {
+		config.Sharing.TimeSlicing.disableResoureRenaming(logger, "timeSlicing")
+	}
 	// Disable renaming / device selection in Sharing.MPS.Resources
-	config.Sharing.MPS.disableResoureRenaming(logger, "mps")
+	if (config.Sharing.MPS != nil) {
+		config.Sharing.MPS.disableResoureRenaming(logger, "mps")
+	}
 }
 
 // parseConfig parses a config file as either YAML of JSON and unmarshals it into a Config struct.
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -96,13 +96,9 @@ func DisableResourceNamingInConfig(logger logger, config *Config) {
 	config.Resources.MIGs = nil
 
 	// Disable renaming / device selection in Sharing.TimeSlicing.Resources
-	if (config.sharing.TimeSlicing != nil) {
-		config.Sharing.TimeSlicing.disableResoureRenaming(logger, "timeSlicing")
-	}
+	config.Sharing.TimeSlicing.disableResoureRenaming(logger, "timeSlicing")
 	// Disable renaming / device selection in Sharing.MPS.Resources
-	if (config.Sharing.MPS != nil) {
-		config.Sharing.MPS.disableResoureRenaming(logger, "mps")
-	}
+	config.Sharing.MPS.disableResoureRenaming(logger, "mps")
 }
 
 // parseConfig parses a config file as either YAML of JSON and unmarshals it into a Config struct.
--- a/api/config/v1/sharing.go
+++ b/api/config/v1/sharing.go
@@ -19,7 +19,7 @@ package v1
 // Sharing encapsulates the set of sharing strategies that are supported.
 type Sharing struct {
 	// TimeSlicing defines the set of replicas to be made for timeSlicing available resources.
-	TimeSlicing ReplicatedResources `json:"timeSlicing,omitempty" yaml:"timeSlicing,omitempty"`
+	TimeSlicing *ReplicatedResources `json:"timeSlicing,omitempty" yaml:"timeSlicing,omitempty"`
 	// MPS defines the set of replicas to be shared using MPS
 	MPS *ReplicatedResources `json:"mps,omitempty"         yaml:"mps,omitempty"`
 }
@@ -38,7 +38,7 @@ func (s *Sharing) SharingStrategy() SharingStrategy {
 		return SharingStrategyMPS
 	}
 
-	if s.TimeSlicing.isReplicated() {
+	if s.TimeSlicing != nil && s.TimeSlicing.isReplicated() {
 		return SharingStrategyTimeSlicing
 	}
 	return SharingStrategyNone
@@ -49,5 +49,5 @@ func (s *Sharing) ReplicatedResources() *ReplicatedResources {
 	if s.MPS != nil {
 		return s.MPS
 	}
-	return &s.TimeSlicing
+	return s.TimeSlicing
 }
--- a/api/config/v1/sharing.go
+++ b/api/config/v1/sharing.go
@@ -38,7 +38,7 @@ func (s *Sharing) SharingStrategy() SharingStrategy {
 		return SharingStrategyMPS
 	}
 
-	if s.TimeSlicing != nil && s.TimeSlicing.isReplicated() {
+	if s.TimeSlicing.isReplicated() {
 		return SharingStrategyTimeSlicing
 	}
 	return SharingStrategyNone
--- a/internal/lm/mig-strategy_test.go
+++ b/internal/lm/mig-strategy_test.go
@@ -183,7 +183,7 @@ func TestMigStrategyNoneLabels(t *testing.T) {
 					},
 				},
 				Sharing: spec.Sharing{
-					TimeSlicing: tc.timeSlicing,
+					TimeSlicing: &tc.timeSlicing,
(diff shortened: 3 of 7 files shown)


api/config/v1/config.go
```diff
@@ -97,9 +97,13 @@ func DisableResourceNamingInConfig(logger logger, config *Config) {
 	config.Resources.MIGs = nil
 
 	// Disable renaming / device selection in Sharing.TimeSlicing.Resources
-	config.Sharing.TimeSlicing.disableResoureRenaming(logger, "timeSlicing")
+	if (config.sharing.TimeSlicing != nil) {
```
elezar, line 100: Note: I think because we have:
```
func (rrs *ReplicatedResources) disableResoureRenaming(logger logger, id string) {
	if rrs == nil {
		return
	}
	...
```
we may not need the `nil` check.

api/config/v1/sharing.go
```diff
@@ -38,7 +38,7 @@ func (s *Sharing) SharingStrategy() SharingStrategy {
 		return SharingStrategyMPS
 	}
 
-	if s.TimeSlicing.isReplicated() {
+	if s.TimeSlicing != nil && s.TimeSlicing.isReplicated() {
```
elezar, line 41:
```suggestion
	if s.TimeSlicing.isReplicated() {
```
we already return `false` if the receiver is `nil` in the `isReplicated()` implementation.

elezar (commented):
Thanks @danielkleinstein.

I have made some suggestions for minor cleanups. I understand that the way this was done probably mirrors the same checks done for MPS, so I'm also ok to merge this as-is and then drop the redundant ` == nil` checks in a follow-up (or as a second commit on top of the one here).

api/config/v1/config.go
danielkleinstein: Thanks - removed in a separate commit

api/config/v1/sharing.go
danielkleinstein: Thanks - removed in a separate commit

danielkleinstein:
@elezar Fixed

internal/lm/mig-strategy_test.go
```diff
@@ -83,7 +83,7 @@ func TestMigStrategyNoneLabels(t *testing.T) {
 				rt.NewFullGPU(),
 				rt.NewFullGPU(),
 			},
-			timeSlicing: spec.ReplicatedResources{
+			timeSlicing: &spec.ReplicatedResources{
```
elezar, line 86: We need to change the type in the struct definition above too.

elezar (dismissed):
Please run `make check` and / or `make test` locally if possible:

```
diff --git a/internal/lm/mig-strategy_test.go b/internal/lm/mig-strategy_test.go
index 005fc961..5aa7b9f6 100644
--- a/internal/lm/mig-strategy_test.go
+++ b/internal/lm/mig-strategy_test.go
@@ -30,7 +30,7 @@ func TestMigStrategyNoneLabels(t *testing.T) {
 	testCases := []struct {
 		description    string
 		devices        []resource.Device
-		timeSlicing    spec.ReplicatedResources
+		timeSlicing    *spec.ReplicatedResources
 		expectedError  bool
 		expectedLabels Labels
 	}{
diff --git a/internal/lm/resource_test.go b/internal/lm/resource_test.go
index 3f180188..3f44763c 100644
--- a/internal/lm/resource_test.go
+++ b/internal/lm/resource_test.go
@@ -271,7 +271,7 @@ func TestMigResourceLabeler(t *testing.T) {
 		description    string
 		resourceName   spec.ResourceName
 		count          int
-		timeSlicing    spec.ReplicatedResources
+		timeSlicing    *spec.ReplicatedResources
 		expectedLabels Labels
 	}{
 		{
@@ -301,7 +301,7 @@ func TestMigResourceLabeler(t *testing.T) {
 			description:  "shared appends suffix and doubles count",
 			resourceName: "nvidia.com/gpu",
 			count:        1,
-			timeSlicing: spec.ReplicatedResources{
+			timeSlicing: &spec.ReplicatedResources{
 				Resources: []spec.ReplicatedResource{
 					{
 						Name:     "nvidia.com/gpu",
@@ -329,7 +329,7 @@ func TestMigResourceLabeler(t *testing.T) {
 			description:  "renamed does not append suffix and doubles count",
 			resourceName: "nvidia.com/gpu",
 			count:        1,
-			timeSlicing: spec.ReplicatedResources{
+			timeSlicing: &spec.ReplicatedResources{
 				Resources: []spec.ReplicatedResource{
 					{
 						Name:     "nvidia.com/gpu",
@@ -358,7 +358,7 @@ func TestMigResourceLabeler(t *testing.T) {
 			description:  "mig mixed appends shared",
 			resourceName: "nvidia.com/mig-1g.1gb",
 			count:        1,
-			timeSlicing: spec.ReplicatedResources{
+			timeSlicing: &spec.ReplicatedResources{
 				Resources: []spec.ReplicatedResource{
 					{
 						Name:     "nvidia.com/gpu",
@@ -391,7 +391,7 @@ func TestMigResourceLabeler(t *testing.T) {
 			description:  "mig mixed rename does not append",
 			resourceName: "nvidia.com/mig-1g.1gb",
 			count:        1,
-			timeSlicing: spec.ReplicatedResources{
+			timeSlicing: &spec.ReplicatedResources{
 				Resources: []spec.ReplicatedResource{
 					{
 						Name:     "nvidia.com/mig-1g.1gb",
```

internal/lm/mig-strategy_test.go
danielkleinstein: I did it a bit differently - I kept the struct definition as-is, and changed the way the test cases are built.

elezar: Sure. Let me re-review.

danielkleinstein:
@elezar Sorry, was kind of hoping to use the project's CI to run the tests 👻 I was working from a Github Codespaces and `make test` didn't run so well. I'll clone locally and make sure tests pass before I push again.

internal/lm/mig-strategy_test.go
danielkleinstein: Blah, there's still CI problems - will make sure it passes before I push again.

elezar:
> @elezar Sorry, was kind of hoping to use the project's CI to run the tests 👻 I was working from a temporary Github Codespaces and `make test` didn't run so well. I'll clone locally and make sure tests pass before I push again.

We require an opt-in to run tests. (Still working on improving the configurations here).

```
diff --git a/internal/lm/resource_test.go b/internal/lm/resource_test.go
index 3f180188..6a3d7ae7 100644
--- a/internal/lm/resource_test.go
+++ b/internal/lm/resource_test.go
@@ -422,7 +422,7 @@ func TestMigResourceLabeler(t *testing.T) {
 		t.Run(tc.description, func(t *testing.T) {
 			config := &spec.Config{
 				Sharing: spec.Sharing{
-					TimeSlicing: tc.timeSlicing,
+					TimeSlicing: &tc.timeSlicing,
 				},
 			}
 			l, err := NewMIGResourceLabeler(tc.resourceName, config, device, tc.count)
```

elezar:
In the existing implementation, the `sharing == nil` code path should only be triggered when calling `NewGPUResourceLabelerWithoutSharing`. This explicitly sets the `replicas` label to `0` and does not add other sharing labels. It has been on my TODO list for a while to rework this to be clearer, but we have not had time to focus on it.

What about removing the last commit that attempts to address this and making the following change:
```
diff --git a/api/config/v1/sharing.go b/api/config/v1/sharing.go
index ab768112..0c35cf30 100644
--- a/api/config/v1/sharing.go
+++ b/api/config/v1/sharing.go
@@ -34,10 +34,9 @@ const (
 
 // SharingStrategy returns the active sharing strategy.
 func (s *Sharing) SharingStrategy() SharingStrategy {
-	if s.MPS != nil && s.MPS.isReplicated() {
+	if s.MPS.isReplicated() {
 		return SharingStrategyMPS
 	}
-
 	if s.TimeSlicing.isReplicated() {
 		return SharingStrategyTimeSlicing
 	}
@@ -46,8 +45,16 @@ func (s *Sharing) SharingStrategy() SharingStrategy {
 
 // ReplicatedResources returns the resources associated with the active sharing strategy.
 func (s *Sharing) ReplicatedResources() *ReplicatedResources {
-	if s.MPS != nil {
+	none := &ReplicatedResources{}
+	if s == nil {
+		return none
+	}
+	switch s.SharingStrategy() {
+	case SharingStrategyMPS:
 		return s.MPS
+	case SharingStrategyTimeSlicing:
+		return s.TimeSlicing
+	default:
+		return none
 	}
-	return s.TimeSlicing
 }
``` 
This ensures that `Sharing.ReplicatedResources()` always returns a valid pointer and that an empty set of resources are returned when required.

rajatchopra:
Closing this PR.
Feel free to re-open, then rebase and address comments from https://github.com/NVIDIA/k8s-device-plugin/pull/1018#issuecomment-2488250650

PR #1025, opened 2024-10-30
[no-relnote] Bump NFD Helm chart to v0.16.6

ArangoGutierrez:


elezar:
Do we need to bump the package versions as part of this as well? How does NFD handle that kind of drift?

ArangoGutierrez:
> Do we need to bump the package versions as part of this as well? How does NFD handle that kind of drift?

I think we need to bump this deps manually to enable Dependabot. because #1007 just tells dependabot to track NFD again and it will fail

elezar:
> > Do we need to bump the package versions as part of this as well? How does NFD handle that kind of drift?
> 
> I think we need to bump this deps manually to enable Dependabot. because #1007 just tells dependabot to track NFD again and it will fail

That wasn't my question. My question was whether we NEED to also update the client, or can a v0.15.x client talk to a v0.16.x server?

ArangoGutierrez:
> > > Do we need to bump the package versions as part of this as well? How does NFD handle that kind of drift?
> > 
> > 
> > I think we need to bump this deps manually to enable Dependabot. because #1007 just tells dependabot to track NFD again and it will fail
> 
> That wasn't my question. My question was whether we NEED to also update the client, or can a v0.15.x client talk to a v0.16.x server?

Oh, sorry, yeah NFD API is stable, `NodeFeature` and `NodeFeatureRules` API's haven't changed during these releases

ArangoGutierrez:
This PR now is just about HELM

ArangoGutierrez:
@elezar PTAL

elezar (approved):

PR #1031, opened 2024-10-31
Honor fail-on-init-error when no resources are found

elezar:
As implemented GFD will not fail if no resources are detected -- even if fail-on-init-error is set. This change ensures that fail-on-init-error is honored if no resources are detected.

This backports #1033 to the release-0.15 branch

### DIFF
--- a/tests/e2e/framework/framework.go
+++ b/tests/e2e/framework/framework.go
@@ -190,7 +190,7 @@ func (f *Framework) AfterEach(ctx context.Context) {
 			for namespaceKey, namespaceErr := range nsDeletionErrors {
 				messages = append(messages, fmt.Sprintf("Couldn't delete ns: %q: %s (%#v)", namespaceKey, namespaceErr, namespaceErr))
 			}
-			e2elog.Failf(strings.Join(messages, ","))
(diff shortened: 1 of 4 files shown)


ArangoGutierrez (approved):

tests/e2e/framework/framework.go
```diff
@@ -190,7 +190,7 @@ func (f *Framework) AfterEach(ctx context.Context) {
 			for namespaceKey, namespaceErr := range nsDeletionErrors {
 				messages = append(messages, fmt.Sprintf("Couldn't delete ns: %q: %s (%#v)", namespaceKey, namespaceErr, namespaceErr))
 			}
-			e2elog.Failf(strings.Join(messages, ","))
+			e2elog.Failf("%s", strings.Join(messages, ","))
```
tariq1890, line 193: Or
```suggestion
			e2elog.Fail(strings.Join(messages, ","))
```

tariq1890 (dismissed):

tests/e2e/framework/framework.go
elezar: Done.

PR #1032, opened 2024-10-31
[no-relnote] Set git user name and email in subfolder repo

elezar:
This prevents errors such as https://github.com/NVIDIA/k8s-device-plugin/actions/runs/11615138763 in the publish-helm action.

### DIFF
--- a/.github/workflows/publish-helm.yaml
+++ b/.github/workflows/publish-helm.yaml
@@ -39,7 +39,7 @@ jobs:
 
     - name: Update helm index
       env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         git config user.name "Github Actions"
         git config user.email "no-reply@github.com"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-helm.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
 
     - name: Update helm index
       env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_TOKEN: ${{ secrets.token }}
```
ArangoGutierrez, line 42: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

ArangoGutierrez: Documentation says is `GITHUB_TOKEN` I agree with the identing bit of this PR

elezar: OK. I was getting confused by the error in:
* https://github.com/NVIDIA/k8s-device-plugin/actions/runs/10300849949

Which was already addressed by the file as shown. Meaning I have to now address:
* https://github.com/NVIDIA/k8s-device-plugin/actions/runs/11615138763
* https://github.com/NVIDIA/k8s-device-plugin/actions/runs/11615080421

ArangoGutierrez: https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/.github/workflows/gh-pages.yml#L36

ArangoGutierrez (approved):

PR #1033, opened 2024-10-31
Honor fail-on-init-error when no resources are found

elezar:
As implemented GFD will not fail if no resources are detected -- even if fail-on-init-error is set. This change ensures that fail-on-init-error is honored if no resources are detected.

### DIFF
--- a/cmd/gpu-feature-discovery/main.go
+++ b/cmd/gpu-feature-discovery/main.go
@@ -186,7 +186,11 @@ func start(c *cli.Context, cfg *Config) error {
 			nvinfo.WithDeviceLib(devicelib),
 		)
 
-		manager := resource.NewManager(infolib, nvmllib, devicelib, config)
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (approved):

cmd/gpu-feature-discovery/main.go
```diff
@@ -186,7 +186,10 @@ func start(c *cli.Context, cfg *Config) error {
 			nvinfo.WithDeviceLib(devicelib),
 		)
 
-		manager := resource.NewManager(infolib, nvmllib, devicelib, config)
+		manager, err := resource.NewManager(infolib, nvmllib, devicelib, config)
+		if err != nil {
+			return err
```
tariq1890, line 191:
```suggestion
			return fmt.Errorf("error instantiating resource manager: %w", err)
```

elezar: Updated.

PR #1042, opened 2024-11-08
Update NV-GHA IP Ranges

ArangoGutierrez:


elezar (commented):
Is this config file the correct place to define this? This seems to depend on infrastructure that isn't provided by us and requires that we update this in all our projects for such changes.

ArangoGutierrez:
> Is this config file the correct place to define this? This seems to depend on infrastructure that isn't provided by us and requires that we update this in all our projects for such changes.

Agree, I think we can use the `k8s-test-infra` repo to centralize this

elezar (approved):

PR #1045, opened 2024-11-09
enable hostPID in the mps-control-daemon pod

tariq1890:
See [here](https://github.com/NVIDIA/k8s-dra-driver/pull/115) for context


deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -57,6 +57,7 @@ spec:
       {{- end }}
       {{- if $options.hasConfigMap }}
       shareProcessNamespace: true
+      hostPID: {{ .Values.mps.hostPID }}
```
elezar, line 60: Should we not be making this change to the MPS control daemon daemonset instead?

tariq1890: Whoops :). Thanks for catching that! But what are your thoughts on applying this setting consistently across the device plugin and MPS daemon?

elezar: I think this goes against the work that we're doing to limit the permissions that we expose in our components. The device plugin should not require host IPC.

deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -150,6 +150,10 @@ mps:
   # directories.
   # Pipe directories will be created at {{ mps.root }}/{{ .ResourceName }}
   root: "/run/nvidia/mps"
+  # hostPID is needed for the MPS server to find its own PID via /proc/self
```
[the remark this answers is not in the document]
tariq1890: Done

elezar (approved):

PR #1060, opened 2024-11-28
Add ada architecture label for compute capability 8.9

elezar:


internal/lm/resource.go
```diff
@@ -298,7 +298,10 @@ func getArchFamily(computeMajor, computeMinor int) string {
 		}
 		return "turing"
 	case 8:
-		return "ampere"
+		if computeMinor < 9 {
+			return "ampere"
+		}
+		return "ada"
```
klueska, line 304: Should this be `ada-lovelace`? I saw your comment elsewhere on the CUDA samples using just `Ada` but I'm not sure if this is the correct source of truth to go by.

elezar: We also have: https://github.com/NVIDIA/go-nvlib/blob/bf0cbc7fef6c51b968da7abd1e8788f0647b2266/pkg/nvlib/device/device.go#L89

klueska: OK. I don't feel too strongly - it mostly just feels inconsistent. Especially since its shortened to "Ada" and not "Lovelace" since all others are the last name, not the first name.

elezar: Updated to `ada-lovelace` and created https://github.com/NVIDIA/go-nvlib/pull/49

klueska:
Looks like this needs a rebase

klueska:
Side question -- should `getArchFamily()` just be a transformation of `nvlib.GetArchitectureAsString()`? Is there any particular reason to reply directly on compute capabilities to generate this string?

elezar:
> Side question -- should `getArchFamily()` just be a transformation of `nvlib.GetArchitectureAsString()`? Is there any particular reason to reply directly on compute capabilities to generate this string?

In theory, yes. The issue is that we do support CUDA devices and we would need a bit of refactoring to use that directly. I have some thoughts on how to do that, but didn't want to address it in this PR.

elezar:
@klueska I have rebased the changes. PTAL.

klueska (approved):

PR #1096, opened 2024-12-11
fix: plugin grpc server retry method

googs1025:
To improve observability and facilitate debugging, this PR enhances logging after the GRPC server crashes


googs1025:
friendly ping @elezar  /PTAL thanks

elezar (approved):

googs1025:
@cdesiniotis can you help this ?

internal/plugin/server.go
```diff
@@ -183,11 +183,13 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
 	go func() {
 		lastCrashTime := time.Now()
 		restartCount := 0
+		maxRestarts := 5
+		crashTimeoutSeconds := float64(3600) // 1 hour in seconds
```
elezar, line 187:
```suggestion
		crashTimeout := 1 * time.Hour
```

internal/plugin/server.go
```diff
@@ -198,17 +200,29 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
 				break
 			}
 
-			klog.Infof("GRPC server for '%s' crashed with error: %v", plugin.rm.Resource(), err)
-
 			timeSinceLastCrash := time.Since(lastCrashTime).Seconds()
```
elezar, line 203:
```suggestion
			timeSinceLastCrash := time.Since(lastCrashTime)
```

internal/plugin/server.go
```diff
@@ -198,17 +200,29 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
...
+
+			klog.Errorf(
+				"GRPC server for '%s' crashed with error: %v. Retry attempt: %d, Time since last crash: %.2f seconds",
+				plugin.rm.Resource(),
+				err,
+				restartCount+1,
+				timeSinceLastCrash,
+			)
+
+			if timeSinceLastCrash > crashTimeoutSeconds {
```
elezar, line 214:
```suggestion
			if timeSinceLastCrash > crashTimeout {
```

internal/plugin/server.go
```diff
@@ -198,17 +200,29 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
...
 			timeSinceLastCrash := time.Since(lastCrashTime).Seconds()
 			lastCrashTime = time.Now()
-			if timeSinceLastCrash > 3600 {
+
+			klog.Errorf(
+				"GRPC server for '%s' crashed with error: %v. Retry attempt: %d, Time since last crash: %.2f seconds",
+				plugin.rm.Resource(),
+				err,
+				restartCount+1,
+				timeSinceLastCrash,
```
elezar, line 211: You probably need:
```suggestion
				timeSinceLastCrash.Seconds(),
```
or need to adjust the format string.

internal/plugin/server.go
```diff
@@ -198,17 +200,29 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
...
 				// to reflect on the frequency
 				restartCount = 0
 			} else {
 				restartCount++
 			}
+
+			// add a small delay before restarting to prevent tight loops
+			retryDelay := 5 * time.Second
+			klog.Infof("Waiting for %v before attempting to restart GRPC server for '%s'", retryDelay, plugin.rm.Resource())
+			time.Sleep(retryDelay) // Wait for 5 seconds before attempting to restart
```
elezar, line 225:
```suggestion
			time.Sleep(retryDelay)
```

internal/plugin/server.go
```diff
@@ -198,17 +200,29 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
...
+
+			if timeSinceLastCrash > crashTimeoutSeconds {
 				// it has been one hour since the last crash.. reset the count
 				// to reflect on the frequency
 				restartCount = 0
 			} else {
 				restartCount++
 			}
+
+			// add a small delay before restarting to prevent tight loops
```
elezar, line 222: Is there a reason that you chose 5 seconds specifically? This seems quite long.

internal/plugin/server.go
```diff
@@ -198,17 +200,29 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
...
+				"GRPC server for '%s' crashed with error: %v. Retry attempt: %d, Time since last crash: %.2f seconds",
+				plugin.rm.Resource(),
+				err,
+				restartCount+1,
+				timeSinceLastCrash,
+			)
+
+			if timeSinceLastCrash > crashTimeoutSeconds {
 				// it has been one hour since the last crash.. reset the count
 				// to reflect on the frequency
```
elezar, line 216: Let's rather just log the `crashTimeout` instead of a comment that is now further from the place where the one hour is specified.

internal/plugin/server.go
```diff
@@ -183,11 +183,13 @@ func (plugin *nvidiaDevicePlugin) Serve() error {
 	go func() {
 		lastCrashTime := time.Now()
 		restartCount := 0
+		maxRestarts := 5
+		crashTimeoutSeconds := float64(3600) // 1 hour in seconds
 
 		for {
 			// quite if it has been restarted too often
 			// i.e. if server has crashed more than 5 times and it didn't last more than one hour each time
```
elezar, line 191: This comment should be moved to the docstring or at the point where we declare the variables.

elezar (changes requested):
Since we're changing this anyway. Let's use the built-in types.

internal/plugin/server.go
googs1025: Indeed, 5 seconds is too long, I changed it to 500 Millisecond. wdyt 🤔

googs1025:
friendly ping  @elezar :) /PTAL

PR #1102, opened 2024-12-19
fix: use embedded configmap for helm config

adammw:
There is currently logic to check the embeddedConfigMap but the `hasConfigMap` is evaluating true first before it even gets to that branch. This restores that logic by only returning true if `hasConfigMap && !hasEmbeddedConfigMap`.


rajatchopra:
@adammw ConfigMap is meant to override anything directly embedded through the values. Which means hasEmbeddedConfigMap does not need to be false.

Please re-open the PR if you feel otherwise.

PR #1106, opened 2024-12-20
Raise a warning instead of an error if GPU mode labeler fails

cdesiniotis:
For environments like WSL2 where GPUs do not show up as PCI devices, creating the GPU mode labeler will fail. This change allows GFD to proceed even if it is unable to get the GPU mode label.

### DIFF
--- a/internal/lm/nvml.go
+++ b/internal/lm/nvml.go
@@ -215,7 +215,8 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
 func newGPUModeLabeler(devices []resource.Device) (Labeler, error) {
 	classes, err := getDeviceClasses(devices)
 	if err != nil {
-		return nil, err
+		klog.Warningf("Failed to create GPU mode labeler: failed to get device classes: %v", err)
+		return Labels{"nvidia.com/gpu.mode": "unknown"}, nil
 	}
 	gpuMode := getModeForClasses(classes)
 	labels := Labels{


tariq1890:
LGTM. I am assuming that GFD just becomes a no-op when run on a node where pci devices aren't detected, yes?

cdesiniotis:
> LGTM. I am assuming that GFD just becomes a no-op when run on a node where pci devices aren't detected, yes?
> 
> 
> 
> 

In WSL2, NVML is available so GFD behaves like normal, except for the gpu.mode label which requires information about the PCI device under /sys.

internal/lm/nvml.go
```diff
@@ -215,7 +215,8 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
 func newGPUModeLabeler(devices []resource.Device) (Labeler, error) {
 	classes, err := getDeviceClasses(devices)
 	if err != nil {
-		return nil, err
+		klog.Warningf("Failed to create GPU mode labeler: failed to get device classes: %v", err)
```
elezar, line 218: The one question is whether we should return an `nvidia.com/gpu.mode="unknown"` label here instead?

cdesiniotis: I am not opposed. Do we have any precedent for using `unknown` for other labels where we are unable to gather the required information?

elezar: We already return `unknown` when getting the mode below.

cdesiniotis: Thanks, I clearly did not look at the code below carefully enough. I have updated this to return `unknown` instead.

elezar (approved):

elezar:
@cdesiniotis do we need to backport this?

internal/lm/nvml.go
```diff
@@ -215,7 +215,8 @@ func isMPSCapable(manager resource.Manager) (bool, error) {
...
+		return Labels{"nvidia.com/gpu.mode": "unknown"}, nil
```
tariq1890, line 219: Is there a string constant for this?

cdesiniotis:
> @cdesiniotis do we need to backport this?

No, as long as this change makes it into the next device-plugin release (0.18.0) and our upcoming GPU Operator release picks it up we should be fine.

internal/lm/nvml.go
cdesiniotis: There is not.

tariq1890 (approved):

PR #1116, opened 2025-01-09
Expose device UUIDs to node label

### LINKED ISSUE #1015: Add gpu uuids to node labels
I know it is easy to get it with nvidia-smi. It would be nice that the gpu-feature-discovery exposes it as a label of nodes, so that one doesn't need to ssh into the node.

xiongzubiao:
Closes #1015

### DIFF
--- a/internal/lm/nvml.go
+++ b/internal/lm/nvml.go
@@ -85,6 +85,11 @@ func NewDeviceLabeler(manager resource.Manager, config *spec.Config) (Labeler, e
 		return nil, fmt.Errorf("error creating IMEX labeler: %v", err)
 	}
 
+	uuidLabler, err := newGPUUUIDLabeler(devices)
+	if err != nil {
+		return nil, fmt.Errorf("error creating UUID labeler: %v", err)
+	}
+
 	l := Merge(
 		machineTypeLabeler,
 		versionLabeler,
@@ -93,6 +98,7 @@ func NewDeviceLabeler(manager resource.Manager, config *spec.Config) (Labeler, e
 		resourceLabeler,
 		gpuModeLabeler,
 		imexLabeler,
+		uuidLabler,
 	)
 
 	return l, nil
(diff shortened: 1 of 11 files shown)


internal/lm/nvml.go
```diff
@@ -260,3 +266,20 @@ func getDeviceClasses(devices []resource.Device) ([]uint32, error) {
...
+	for _, d := range devices {
+		uuid, err := d.GetUUID()
+		if err != nil {
+			return nil, err
+		}
+		uuids = append(uuids, uuid)
+	}
+
+	labels := Labels{
+		"nvidia.com/gpu.uuid": strings.Join(uuids, ","),
```
shan100github, line 282: Its not a good idea to keep appending all uuids in a single label since "Valid label value: must be 63 characters or less"
 https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/.
 
 Also for a server with 8GPUs adding all uuid in single label might not end up useful too.

xiongzubiao: Good point. Annotation can handle it, but there seems no easy way to add annotation from the gpu-feature-discovery. I ended up adding a label per gpu device.

elezar:
@xiongzubiao could you please provide information on how these labels will be used?

xiongzubiao:
> @xiongzubiao could you please provide information on how these labels will be used?

@elezar, we want to provide some sort of visualization to user. User can click each GPU to check its properties, status, and metrics. The device UUID is the natural choice for indexing. There are other ways to get UUID, but it is most straightforward to get it from node labels, because it is a part of node properties.

There is another use mentioned #1015: scheduling pod to a specific GPU using node label matching.

ramazanyich:
would like to have such a label in a scope of our project too. Where we need to see the GPU UUISs linked to node. So we can use that info in our workflows.

internal/resource/testing/resource-testing.go
```diff
@@ -41,6 +41,7 @@ func NewMigEnabledDevice(migs ...*resource.DeviceMock) resource.Device {
 func NewDeviceMock(migEnabled bool) *DeviceMock {
 	d := DeviceMock{resource.DeviceMock{
 		GetNameFunc: func() (string, error) { return "MOCKMODEL", nil },
+		GetUUIDFunc: func() (string, error) { return "MOCKUUID", nil },
```
rajatchopra, line 44: We need a negative test case too maybe. e.g. when we couldn't get the uuid.

rajatchopra (approved):
/lgtm

rajatchopra:
While the PR is appropriate and the implementation is good, we are not sure if we want to add extra labels unless necessary.
The use case of metrics lookup becoming easier isn't very convincing because GFD running as one-shot in replaceable VMs (as happens with GKE node pool break-fix) may result in stale UUIDs.
The right source of truth should be dcgm-exporter.

Closing the PR. Please re-open if anyone disagrees.

PR #1133, opened 2025-01-23
Add resources to sidecar container

jchanam:
When deploying the chart with a `hasConfigMap`, a sidecar container is added to the pod. This sidecar has no way to set the resources right now.

This PR reuses the `resources` value to also set it to the sidecar.


jchanam:
@elezar Could you please review this or assign anyone with permissions? thanks!

rajatchopra:
@jchanam Why would the config-manager need resources? For cpu/memory limits?
Also, the `resources` are used by the main container as of now. Not meant for the sidecar.

jchanam:
@rajatchopra thanks for checking this.

Yes, I want all the containers deployed in my clusters to have resources assigned to them, so I can have full control of the requested/limited cpu and memory.

If you don't want to use the current container resources, a new entry map could be added only for the sidecar.

abrarshivani:
@jchanam Can we add per-component resource configuration?

PR #1152, opened 2025-02-12
Remove nvidia.com/gpu.imex-domain label

elezar:
This removes the logic for constructing an nvidia.com/gpu.imex-domain label. This also means that the logic to include the imex nodes_config.cfg file in the GFD container is not required.

This is a backport of #1147


internal/lm/imex.go
```diff
@@ -17,92 +17,26 @@
...
-		return nil, nil
-	}
-
-	imexDomainID, err := getImexDomainID(imexConfigFile)
-	if err != nil {
-		return nil, err
-	}
-	if imexDomainID == "" {
-		return nil, nil
+		return empty{}, nil
```
klueska, line 35: why `empty{}` instead of `nil`?

elezar: In order to avoid explicit `nil` checks at the callsite, we have an `empty{}` `Labeler` which returns no labels.

klueska: OK. Thanks.

cdesiniotis:
@tariq1890 we should look into rolling back the changes introduced in https://github.com/NVIDIA/gpu-operator/pull/1070 as they are no longer needed.

internal/lm/imex.go
ArangoGutierrez: could we do the same for line (89/32)

internal/lm/imex.go
```diff
@@ -17,92 +17,26 @@
...
-		// No imex config file, return empty labels
-		return nil, nil
-	} else if err != nil {
-		return nil, fmt.Errorf("failed to open imex config file: %v", err)
-	}
-	defer imexConfigFile.Close()
-
 	clusterUUID, cliqueID, err := getFabricIDs(devices)
 	if err != nil {
 		return nil, err
```
ArangoGutierrez, line 32:
```suggestion
		return empty{}, nil
```

ArangoGutierrez (changes requested):

internal/lm/imex.go
elezar: > could we do the same for line (89/32)

No. On that line we're returning an actual error and we want to return `nil` as the labeler.

internal/lm/imex.go
elezar: No. Here we are returning an actual error. Do you have a motivation as to why we want to ignore it?

jgehrcke (approved):
I looked at the review conversations. They look good. I approve ignorantly. Thank you, @elezar

PR #1154, opened 2025-02-12
[no-relnotes] Use reusable workflows for CI

elezar:


.github/workflows/e2e.yaml
```diff
@@ -15,70 +15,73 @@
...
-        if: ${{ failure() }}
-        uses: actions/upload-artifact@v4
-        with:
-          name: e2e-test-logs
-          path: ./e2e_logs/
-          retention-days: 15
+    - name: Send Slack alert notification
+      id: slack
+      if: false
+      uses: slackapi/slack-github-action@v2.0.0
```
elezar, line 78: @ArangoGutierrez the upgrade to v2.0.0 has broken this step. I believe the API has changed.

.github/workflows/code_scanning.yaml
```diff
@@ -15,21 +15,16 @@
...
+  workflow_call:
+    inputs:
+      golang_version:
+        required: true
+        type: string
 
 jobs:
   analyze:
     name: Analyze Go code with CodeQL
+    if: always()
```
elezar, line 27: I think this can be removed

.github/workflows/e2e.yaml
ArangoGutierrez: Let me check and take action for our repos

ArangoGutierrez: fixed

.github/workflows/code_scanning.yaml
ArangoGutierrez: yup

elezar:
I'm still not 100% about the naming of the steps, but I suppose we can iterate on that at a later point.

ArangoGutierrez:
> I'm still not 100% about the naming of the steps, but I suppose we can iterate on that at a later point.

same here, not 100% on "basic" but I can't think of a better name

tests/e2e/device-plugin_test.go
```diff
@@ -162,12 +162,16 @@ var _ = NVDescribe("GPU Device Plugin", func() {
 						"nvidia.com/gpu": "^[1-9]$",
 					}}
 				eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchCapacity(capacityChecker, nodes), "Node capacity does not match")
-			})
-			It("it should run GPU jobs", func(ctx context.Context) {
-				By("Creating a GPU job")
+
```
elezar, line 165: @ArangoGutierrez just a note. There is an issue with the tests uninstalling the plugin between stanzas. This "fixes" them, but we should follow up at some point.

PR #1165, opened 2025-02-14
Ignore XID error 109

elezar:
This change adds XID error 109 to the list of ignored errors.


tariq1890 (approved):

HtcOrange:
> This change adds XID error 109 to the list of ignored errors.

We recently encountered an issue where running the application on the internal cluster triggers many 109 errors. Our device plugin is modified based on an outdated version, and the GPU currently experiencing the issue is an A100. Could you please tell me the starting point of the repair logic for this PR? Did you discover any issues that caused the official team to ignore this xid?

PR #1178, opened 2025-02-25
Allow MPS control daemonset to be explicitly disabled

### LINKED ISSUE #1177: Helm Chart Unable to Prevent MPS DaemonSet Deployment
**Problem description:**
Currently there's no way for a helm chart user to disable deploying MPS DaemonSet deployment without having to disable `devicePlugin.enabled` which affects both MPS & device-plugin DaemonSet.

**Desired solution:**
Be able to disable MPS DaemonSet without affecting device-plugin DaemonSet.

marinoborges:
This change allows the deployment of the MPS daemon to be explicitly disabled by adding `--set mps.enabled=false` to a  Helm install / upgrade command.

The default behaviour of the plugin is to deploy the MPS control daemonset even if not MPS sharing is configured. However, the actual MPS daemon is only started if a GPU is replicated using MPS. The `mps.enabled` Helm value allows this to be explicitly disabled. 

Fixes #1177 with backward compatibility so this should be a minor change.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -135,8 +135,10 @@ spec:
         name: nvidia-device-plugin-ctr
         command: ["nvidia-device-plugin"]
         env:
+        {{- if .Values.mps.enabled }}
           - name: MPS_ROOT
             value: {{ .Values.mps.root }}
+        {{- end }}
         {{- if typeIs "string" .Values.migStrategy }}
           - name: MIG_STRATEGY
             value: {{ .Values.migStrategy }}
@@ -205,6 +207,7 @@ spec:
             mountPath: /driver-root
             readOnly: true
         {{- end }}
+        {{- if .Values.mps.enabled }}
           # The MPS /dev/shm is needed to allow for MPS daemon health-checking.
           - name: mps-shm
             mountPath: /dev/shm
@@ -212,6 +215,7 @@ spec:
             mountPath: /mps
           - name: cdi-root
             mountPath: /var/run/cdi
+        {{- end }}
         {{- if $options.hasConfigMap }}
           - name: available-configs
             mountPath: /available-configs
@@ -226,6 +230,7 @@ spec:
         - name: device-plugin
           hostPath:
             path: /var/lib/kubelet/device-plugins
+        {{- if .Values.mps.enabled }}
         - name: mps-root
           hostPath:
             path: {{ .Values.mps.root }}
@@ -233,15 +238,16 @@ spec:
         - name: mps-shm
           hostPath:
             path: {{ .Values.mps.root }}/shm
+        - name: cdi-root
+          hostPath:
+            path: /var/run/cdi
+            type: DirectoryOrCreate
+        {{- end }}
        {{- if typeIs "string" .Values.nvidiaDriverRoot }}
         - name: driver-root
           hostPath:
             path: {{ .Values.nvidiaDriverRoot }}
         {{- end }}
-        - name: cdi-root
-          hostPath:
-            path: /var/run/cdi
-            type: DirectoryOrCreate
       {{- if $options.hasConfigMap }}
         - name: available-configs
           configMap:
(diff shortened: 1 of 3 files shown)


ArangoGutierrez (commented):
@elezar / @tariq1890  Thoughts?

marinoborges:
Unsure if we need also to change [NVIDIA_DRIVER_CAPABILITIES](https://github.com/NVIDIA/k8s-device-plugin/pull/1178/files#diff-dbab68f06da27aa9f4750f072fdb4750e0c1ccd71db972b9ebc766cf7b65694cR197) env var value when running without MPS.

deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -145,6 +145,7 @@ nfd:
           - vendor
 
 mps:
+  enabled: true
```
chipzoller, line 148: As a user, I think at a minimum there needs to be comments placed above this line to explain its use. If I see this in a values file, I may interpret `mps.enabled: true` to mean I, as a user, want MPS enabled right out of the gate. But that's not what this does. When set to `false` MPS will never work even if I later activate an MPS configuration by labeling a node.

chipzoller (commented):
You may also consider renaming this field so its purpose is more clear. Something more explicit like `createDaemonset` would be an example.

deployments/helm/nvidia-device-plugin/values.yaml
marinoborges: I added a comment clarifying this

sidewinder12s:
@chipzoller can you re-review this?

chipzoller (commented):
The comment itself looks fine, but this will need to be reviewed by a maintainer (I am not one).

elezar (commented):
Sorry for the delay here @marinoborges. I think the opt-out makes sense. I did have one question though:

What if a user is upgrading the device plugin and as such this value is not defined in the set that has already been applied? Does the default value of `true` function as expected here, or should we check for non-`nil` values explicitly?

marinoborges:
@elezar if a user is upgrading device plugin via helm chart upgrade, the default value of `true` will be applied so no undesired changes are expected.
Also, if a user is upgrading device plugin via `image` [value](https://github.com/NVIDIA/k8s-device-plugin/blob/4eb8001984b148a339114ba86e9e20abdcbc92d3/deployments/helm/nvidia-device-plugin/values.yaml#L48), then the helm chart change i'm proposing isn't taken into consideration so again no undesired changes are expected.

elezar:
Sorry for the delay in getting to this.

Since we are slowly working on adding MPS support to the k8s DRA driver for GPUs, we may want to consider making the MPS support in the device plugin explicitly OPT-in. We have not had the time to further flesh out this feature, and making it opt-in in addition to requiring an explicit sharing config with MPS replicas would help to set expectations accordingly.

@klueska what are your thoughts on this?

rahulait:
@elezar @klueska Is there any conclusion on this topic based on the past discussions?

deployments/helm/nvidia-device-plugin/values.yaml
rajatchopra: Right. This is more like a feature gate.

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -205,13 +207,15 @@ spec:
             mountPath: /driver-root
             readOnly: true
         {{- end }}
+        {{- if .Values.mps.enabled }}
           # The MPS /dev/shm is needed to allow for MPS daemon health-checking.
           - name: mps-shm
             mountPath: /dev/shm
           - name: mps-root
             mountPath: /mps
           - name: cdi-root
```
rajatchopra, line 217: cdi-root mount cannot be under the mps condition. Please fix.

rajatchopra (changes requested):

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
marinoborges: @rajatchopra , Addressed this in the latest commit

PR #1179, opened 2025-02-25
[FEATURE] [HELM] Add support to setup a custom serviceAccount

### LINKED ISSUE #1174: Helm Chart Unable to Specify Custom Service Account
**Problem description:**
Currently there's no way for a helm chart user to specify a custom `serviceAccount` without having `hasConfigMap` set or `gfd.enabled` plus `nfd.enableNodeFeatureApi` set [[1](https://github.com/NVIDIA/k8s-device-plugin/blob/186864bceed24ab269ba4f4813587c922273127b/deployments/helm/nvidia-device-plugin/templates/service-account.yml#L3C71-L3C91)].

**Desired solution:**
Be able to set a custom `serviceAccount` via helm values.

marinoborges:
Fixes #1174 with backward compatibility so this should be a minor change.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -52,8 +52,12 @@ spec:
       {{- end }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      {{- if $useServiceAccount }}
+      {{- if or .Values.serviceAccount.create $useServiceAccount }}
+      {{- if .Values.serviceAccount.name }}
+      serviceAccountName: {{ .Values.serviceAccount.name }}
+      {{- else }}
       serviceAccountName: {{ include "nvidia-device-plugin.fullname" . }}-service-account
+      {{- end }}
       {{- end }}
       {{- if $options.hasConfigMap }}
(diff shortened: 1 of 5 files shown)


deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -52,9 +52,13 @@ spec:
       {{- end }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      {{- if $useServiceAccount }}
```
ArangoGutierrez, line 55: Not sure about this, it could introduce a breaking change , we need to find a way that won't break existing users scripts that may rely on this existing HELM setup

ArangoGutierrez (changes requested):
Hey @marinoborges thanks for the PR! a nice addition for sure, happy to review

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
marinoborges: It's not a breaking change because the existing logic is respected. It replaces this `if` with another one but still considering `$useServiceAccount`.

dmpe:
Any chance this will be merged anytime soon. Just because of additing annotations to the service account object....

PR #1211, opened 2025-03-24
 Mark device as healthy if checkHealth function does not receive unhealthy event

Zeel-Patel:
Solves: [1014](https://github.com/NVIDIA/k8s-device-plugin/issues/1014) 
Currently, if device is marked as unhealthy, it's status remain unhealthy and kubelet does not show resource on node.

The ListAndWatch stream Receiver of kubelet device manager [checks health of device](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/devicemanager/manager.go#L297) and takes action to update node resources.

Making the change in checkHealth function to send device healthy signal (using healthy channel) to server ListAndWatch to update device health as healthy in case no event related to device unhealthy is received


jslouisyou:
Thanks @Zeel-Patel!
Could you please let me know when the next version of `k8s-device-plugin` will be released?
If possible then I can test it on my side.

Zeel-Patel:
cc @klueska

amitaekbote:
@klueska any eta for the component?

internal/plugin/server.go
```diff
@@ -57,10 +57,11 @@ type nvidiaDevicePlugin struct {
 	cdiHandler          cdi.Interface
 	cdiAnnotationPrefix string
 
-	socket string
-	server *grpc.Server
-	health chan *rm.Device
-	stop   chan interface{}
+	socket    string
+	server    *grpc.Server
+	healthy   chan *rm.Device
```
elezar, line 62: Instead of introducing a new channel, is there any benefit in having a single channel that accepts a device and the desired status?

elezar: We could even mark the device as healthy or unhealthy at the point where we send the device on the channel and then keep the health channel as is.

internal/rm/health.go
```diff
@@ -174,6 +164,18 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 			continue
 		}
 
+		if e.EventType != nvml.EventTypeXidCriticalError {
+			klog.Infof("Skipping non-nvmlEventTypeXidCriticalError event: %+v", e)
+			healthy <- d
```
elezar, line 169: Are xid events mutually exclusive? Why are we treating non-critical (or skipped events) as indicators of device health?

Zeel-Patel:
Hi @elezar , thanks for the review.
Do you feel this event can be exposed from NVIDIA to open source. It will help in the change made in PR
NVML_FI_DEV_GET_GPU_RECOVERY_ACTION

internal/rm/health.go
```diff
@@ -174,6 +164,18 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
...
+			continue
+		}
+
+		if skippedXids[e.EventData] {
```
HtcOrange, line 173: Assuming that the device is set as unhealthy due to event A, event B is detected at this time (i.e. satisfying the condition of (e.EventType! =Nvml. EventTypeXidCriticalError)) caused the device to be restored to health. Will this situation lead to the device being erroneously restored to health? I think a reasonable approach should be to restore the device from the same error state before setting it to healthy?

PR #1221, opened 2025-04-03
use MiB instead of MB for gpu memory

natasharomm:
gpu-feature-discovery -
Fix documentation for `nvidia.com/gpu.memory` label - the value represents the memory in MiB instead of MB.


internal/resource/cuda-device.go
```diff
@@ -58,8 +58,8 @@ func (d *cudaDevice) GetDeviceHandleFromMigDeviceHandle() (Device, error) {
 	return nil, fmt.Errorf("GetDeviceHandleFromMigDeviceHandle is unsupported for CUDA devices")
 }
 
-// GetTotalMemoryMB returns the total memory for a device
-func (d *cudaDevice) GetTotalMemoryMB() (uint64, error) {
+// GetTotalMemoryMiB returns the total memory for a device
```
elezar, line 61: QUestion: Do we want to add something that makes `MiB` clearer?

```suggestion
// GetTotalMemoryMiB returns the total memory for a device in mebibytes (2^20 bytes)
```

internal/resource/device_mock.go
elezar, line 1: Running:
```
make .build-image
make docker-generate
```
should regenerate these mocks -- updating manually is also ok.

internal/resource/nvml-device.go
```diff
@@ -81,8 +81,8 @@ func (d nvmlDevice) GetName() (string, error) {
 	return name, nil
 }
 
-// GetTotalMemoryMB returns the total memory on a device in MB
-func (d nvmlDevice) GetTotalMemoryMB() (uint64, error) {
+// GetTotalMemoryMiB returns the total memory on a device in MB
```
elezar, line 84:
```suggestion
// GetTotalMemoryMiB returns the total memory on a device in mebibytes (2^20 bytes).
```

elezar (commented):
Looks good. Thanks @natasharomm.

Could we squash the commits into a single comment and then I think we're good to merge.

elezar (approved):

PR #1233, opened 2025-04-15
chore: Update golang version to 1.24.1

dongjiang1989:


deployments/container/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG GOLANG_VERSION=1.22.6
+ARG GOLANG_VERSION=1.24.1
```
tariq1890, line 15:
```suggestion
ARG GOLANG_VERSION=1.24.2
```

deployments/devel/go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/NVIDIA/k8s-device-plugin/deployments/devel
 
-go 1.23
+go 1.24
```
tariq1890, line 3: Is this needed? Unless the minimum required go version to build k8s-device-plugin is truly 1.24, we shouldn't be changing this

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/NVIDIA/k8s-device-plugin
 
-go 1.23.0
+go 1.24
```
tariq1890, line 3: Same as above

deployments/devel/go.mod
dongjiang1989: Not required.
By upgrading to the latest version of golang, we can ensure that the `dependabot` gomod base library can be upgraded to the latest version

elezar:
Please note that Dependabot handles this automatically. See https://github.com/NVIDIA/k8s-device-plugin/pull/1319

PR #1249, opened 2025-04-28
Added an ngc-publish stage to trigger the NGC Automation scripts for publishing

JunAr7112:
This MR adds an ngc-publish stage to trigger the NGC Automation scripts as part of the publishing workflow. The NGC Automation scripts will raise an MR against the NGC configs repo for publishing per the new NGC requirements. The pipeline workflow was tested [here](https://gitlab-master.nvidia.com/dl/container-dev/kubernetes/device-plugin/-/pipelines/27471789).

### DIFF
--- a/.common-ci.yml
+++ b/.common-ci.yml
@@ -160,9 +160,9 @@ scan-ubi9-arm64:
   before_script:
     - !reference [.regctl-setup, before_script]
 
-    # We ensure that the OUT_IMAGE_VERSION is set
+    # We ensure that the OUT_IMAGE_VERSION and OUT_IMAGE_NAME are set
     - 'echo Version: ${OUT_IMAGE_VERSION} ; [[ -n "${OUT_IMAGE_VERSION}" ]] || exit 1'
-
+    - 'echo Version: ${OUT_IMAGE_NAME} ; [[ -n "${OUT_IMAGE_NAME}" ]] || exit 1'
     # In the case where we are deploying a different version to the CI_COMMIT_SHA, we
     # need to tag the image.
     # Note: a leading 'v' is stripped from the version if present
@@ -185,10 +185,10 @@ scan-ubi9-arm64:
   extends:
     - .release
   variables:
-    OUT_REGISTRY_USER: "${CI_REGISTRY_USER}"
-    OUT_REGISTRY_TOKEN: "${CI_REGISTRY_PASSWORD}"
-    OUT_REGISTRY: "${CI_REGISTRY}"
-    OUT_IMAGE_NAME: "${CI_REGISTRY_IMAGE}/staging/k8s-device-plugin"
+    OUT_REGISTRY_USER: "${NGC_REGISTRY_USER}"
+    OUT_REGISTRY_TOKEN: "${NGC_REGISTRY_TOKEN}"
+    OUT_REGISTRY: "${NGC_REGISTRY}"
+    OUT_IMAGE_NAME: "${NGC_REGISTRY_STAGING_IMAGE_NAME}"
 
 # Define an external release step that pushes an image to an external repository.
 # This includes a devlopment image off main.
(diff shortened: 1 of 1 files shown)


.common-ci.yml
```diff
@@ -160,8 +161,8 @@ scan-ubi9-arm64:
   before_script:
     - !reference [.regctl-setup, before_script]
 
-    # We ensure that the OUT_IMAGE_VERSION is set
-    - 'echo Version: ${OUT_IMAGE_VERSION} ; [[ -n "${OUT_IMAGE_VERSION}" ]] || exit 1'
+    # We ensure that the OUT_IMAGE_VERSION and OUT_IMAGE_NAME are set
+    - 'echo "Version: ${OUT_IMAGE_VERSION}, Image: ${OUT_IMAGE_NAME}" ; [[ -n "${OUT_IMAGE_VERSION}" && -n "${OUT_IMAGE_NAME}" ]] || exit 1'
```
elezar, line 165: nit: I think repeating the two checks across two separate lines is simpler and more robust.

.common-ci.yml
```diff
@@ -209,3 +210,18 @@ release:staging-ubi9:
...
+
+# Triggers the downstream container-dev/ngc-automation pipeline to publish the image to NGC
+# This stage is triggered specifically on tags   
+trigger-downstream-publishing-pipeline:
+  stage: ngc-publish
+  rules:
+    - if: $CI_COMMIT_TAG
+  variables:
+    PROJECT_NAME: "k8s-device-plugin"
+    SOURCE_VERSION: "${CI_COMMIT_TAG}"
```
elezar, line 222: Question: In discussions with @tariq1890 yesterday, I think we settled on using the short-SHA as the `SOURCE_VERSION` and the `CI_COMMIT_TAG` as teh output version. 

Also, instead of maintaining the mapping of `PROJECT_NAME` to full image name in `dl/container-dev/ngc-automation`, does it makes sense to also pass: `NGC_REGISTRY_STAGING_IMAGE_NAME` ?

.common-ci.yml
```diff
@@ -209,3 +210,18 @@ release:staging-ubi9:
     - .dist-ubi9
   needs:
     - image-ubi9
+
+# Triggers the downstream container-dev/ngc-automation pipeline to publish the image to NGC
+# This stage is triggered specifically on tags   
+trigger-downstream-publishing-pipeline:
```
elezar, line 216: Does this job remove the need for anything that depends on `.release:external:`? (we still have `release:ngc-ubi9`) in `.nvidia.yaml`.

.common-ci.yml
```diff
@@ -209,3 +210,18 @@ release:staging-ubi9:
...
+  stage: ngc-publish
```
elezar, line 217: should this have a `needs` entry?

.common-ci.yml
```diff
@@ -29,6 +29,7 @@ stages:
   - scan
   - release
   - sign
+  - ngc-publish
```
elezar, line 32: Do we need a new `stage`, or can we reuse the `release` stage?

.common-ci.yml
JunAr7112: Hi Evan, I updated the SOURCE_VERSION to CI_COMMIT_SHORT_SHA. 

On the PROJECT_NAME, in ngc-automation I don't directly map the PROJECT_NAME to the full image name (ie nvcr.io/nvstaging/cloud-native/k8s-device-plugin), instead I map it to a project specific yaml file k8s-device-plugin.yaml which is pushed to the NGC repo for publishing. So I don't think I will need to directly pass NGC_REGISTRY_STAGING_IMAGE_NAME.

.common-ci.yml
JunAr7112: Alright, I switched to two separate lines.

.common-ci.yml
JunAr7112: Hi Evan, initially Chris and I were planning for a separate ngc-publish stage to trigger the ngc-automation repo. But I could instead reuse the release stage and include a needs entry on release:staging-ubi9 to ensure the job executes after the image is staged in nvstaging. Would that format work better?

JunAr7112: Following up on this discussion, I added an explicit needs entry on release:staging-ubi9. I don't think we need an explicit ngc-publish stage, so I am fine running trigger-downstream-publishing-pipeline as part of the release stage.

.common-ci.yml
```diff
@@ -209,3 +209,24 @@ release:staging-ubi9:
...
+  needs:
+    - release:staging-ubi9
+  stage: release
+  rules:
+    - if: $CI_COMMIT_TAG
+  variables:
+    PROJECT_NAME: "k8s-device-plugin"
+    SOURCE_VERSION: "${CI_COMMIT_SHORT_SHA}"
+    TARGET_VERSION: "${CI_COMMIT_TAG}"
+    IMAGE_DIST_LIST: ",${DIST}"
```
elezar, line 227: Probably out of scope for this PR, but why is there a `comma` here?

elezar:
This was addressed in https://github.com/NVIDIA/k8s-device-plugin/pull/1230 and https://github.com/NVIDIA/k8s-device-plugin/pull/1330 (and follow-ups).

PR #1264, opened 2025-05-14
WIP: Add an NRI plugin to filter which NUMA nodes a container has access to

klueska:
This PR implements a Node Resource Interface (NRI) plugin to filter the set of NUMA nodes each container started by Kubernetes has access to. By default, each container is given access to either (1) the full set system NUMA nodes, or (2) the set of system nodes they have already been limited to before hitting the NRI plugin. The NUMA nodes associated with each GPU or MIG device they have access to are then added in one-by-one.

Currently, the implementation only works by parsing the incoming `NVIDIA_VISIBLE_DEVICES` envar set in the container to discover which GPUs / MIG devices it has access to. 

Before we can merge this PR we will need to add the following features (at a minimum):
1. Add a flag to disable the NRI plugin from running entirely
1. Add a helm-chart option to toggle the enable/disable the running of the NRI plugin
1. Support CDI injected GPU/MIG devices
1. Support volume mounted GPU/MIG devices
1. Support NVIDIA_VISIBLE_DEVICES with indices instead of UUIDs


internal/nri/plugin.go
```diff
@@ -0,0 +1,385 @@
...
+	systemNodes := make([]int, 0, len(nodes))
+	for _, node := range nodes {
+		nodeID := filepath.Base(node)[4:] // Remove "node" prefix
+		// Check if the node has any CPUs assigned
+		cpulist, err := os.ReadFile(filepath.Join(node, "cpulist"))
+		if err != nil {
+			continue
+		}
+
+		// Only include nodes with non-empty CPU lists
```
jgehrcke, line 322: So, we define a "system node" to be any NUMA node that has a CPU assigned. Is that right? Is that the one, definite criterion or is there any other criterion? Should we call them `cpuNodes`? Is there a better way to detect system nodes?

internal/nri/plugin.go
```diff
@@ -0,0 +1,385 @@
...
+				return nil, nil
+			case "all":
+				uuids, err := p.getAllDeviceUUIDs()
+				if err != nil {
+					return nil, fmt.Errorf("failed to get all device UUIDs: %w", err)
+				}
+				return uuids, nil
+			default:
+				// Handle comma-separated list of UUIDs
+				return strings.Split(value, ","), nil
```
jgehrcke, line 273: Can `value` at times be a list of integers instead of a list of UUIDs?

internal/nri/plugin.go
```diff
@@ -0,0 +1,385 @@
...
+			},
+		}
+	}
+
+	return adjust, nil, nil
+}
+
+// UpdateContainer is called when a container's resources are being updated. It ensures the
+// container remains on the correct NUMA nodes based on its assigned GPUs and MIG devices.
+func (p *NUMAFilterPlugin) UpdateContainer(ctx context.Context, pod *api.PodSandbox, container *api.Container, resources *api.LinuxResources) ([]*api.ContainerUpdate, error) {
```
jgehrcke, line 169: From were is this called? Ah... I assume this is the NRI lib that is calling into this.

internal/nri/plugin.go
```diff
@@ -0,0 +1,385 @@
...
+
+	// Cache the result
+	p.cache.deviceNodes[uuid] = node
+	return node, nil
+}
+
+// getAllSystemNUMANodes gets all NUMA nodes in the system that have CPUs assigned to them
+func (p *NUMAFilterPlugin) getAllSystemNUMANodes() ([]int, error) {
+	// Read /sys/devices/system/node/node* to get NUMA nodes
+	nodes, err := filepath.Glob("/sys/devices/system/node/node*")
```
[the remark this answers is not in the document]
klueska: Yeah, we should probably return an error. I'd rather error out and have to deal with a bug report if we ever encounter a system that doesn't have NUMA enabled, than to a silently skip this without setting anything. One can always disable deployment of the plugin on such a system if they do encounter the error (once we add this option).

internal/nri/plugin.go
klueska: That is the only criteria. We usually refer to this kind of memory as "system memory", which is why I picked "system nodes", but I'm open to suggestions.

internal/nri/plugin.go
klueska: Yes, that is why I have the todo at the top of this saying we need to handel all `list-strategies`. I didn't include it in the list in the PR description though -- adding now.

internal/nri/plugin.go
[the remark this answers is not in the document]
klueska: Piotr suggested just maintaining a list of GPU nodes vs. non-GPU nodes instead of an explicit list of system nodes. I think the rest of the logic would be robust to doing this since it operates as follows:

* When a container is being started, I first inspect what NUMA nodes it already has set
* If none are set, I set it to the full set of system nodes
* if something _is_ set, I leave it in place but filter it against the system nodes I know about (essentially filtering out the GPU nodes)
* I then go back through 1-by1 adding the GPU nodes for the GPUs the container has access to

I think the same logic could be applied if all I tracked was GPU nodes and non-GPU nodes instead of trying to explicitly discover the system nodes.

internal/nri/.plugin.go.swp
elezar, line 1: Should probably not be included?

PR #1277, opened 2025-06-03
remove internal logger in favour of klog

tariq1890:
The internal logger is an indirect call to the `klog` logger. It's used quite minimally, so I figured we'd just call `klog` directly instead


elezar (approved):
Thanks @tariq1890.

Looks good. Is this a change that you think should be backported?

tariq1890:
Thanks for the review. I don't think this change needs to be backported

PR #1284, opened 2025-06-12
pass a single context throughout the device-plugin method call stack

tariq1890:
This change follows the Go best practices. With a single ctx reference, we allow for the proper propagation of cancellations and graceful terminations across all goroutines of the device-plugin application.

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -354,7 +354,7 @@ func startPlugins(c *cli.Context, o *options) ([]plugin.Interface, bool, error) 
 
 	// Get the set of plugins.
 	klog.Info("Retrieving plugins.")
-	plugins, err := GetPlugins(infolib, nvmllib, devicelib, config)
+	plugins, err := GetPlugins(c.Context, infolib, nvmllib, devicelib, config)
 	if err != nil {
 		return nil, false, fmt.Errorf("error getting plugins: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


cmd/nvidia-device-plugin/main.go
```diff
@@ -354,7 +354,7 @@ func startPlugins(c *cli.Context, o *options) ([]plugin.Interface, bool, error)
 
 	// Get the set of plugins.
 	klog.Info("Retrieving plugins.")
-	plugins, err := GetPlugins(infolib, nvmllib, devicelib, config)
+	plugins, err := GetPlugins(c.Context, infolib, nvmllib, devicelib, config)
```
elezar, line 357: As a matter of interest, what is the CLI context?

elezar: Update: It seems as if it's using `context.Background()` in any case.

internal/plugin/server.go
```diff
@@ -75,6 +76,7 @@ func (o *options) devicePluginForResource(resourceManager rm.ResourceManager) (I
 	}
 
 	plugin := nvidiaDevicePlugin{
+		ctx:                  o.ctx,
```
elezar: Are there cases where this MAY be `nil`?

tariq1890: Great question. No, I don't think so. I guess we don't need the WithContext method then?

cmd/nvidia-device-plugin/main.go
tariq1890: Yes. Since this is a CLI "action", we should use the CLI object's context and only that (or a context derived from it) throughout the application.

internal/plugin/server.go
tariq1890: I've updated PR to remove the Option method to set the context

elezar (approved):
Thanks @tariq1890. LGTM.

PR #1335, opened 2025-07-28
Add env var DP_ENABLE_HEALTHCHECKS

robertdavidsmith:
Make it possible to enable health checks for xids which are disabled by default via an env var `DP_ENABLE_HEALTHCHECKS`.

This is occasionally useful, e.g. we have seen examples where xid 45 is very correlated with genuine GPU issues.

### DIFF
--- a/internal/rm/health.go
+++ b/internal/rm/health.go
@@ -27,26 +27,26 @@ import (
 )
 
 const (
+	allHealthChecks = "xids"
 	// envDisableHealthChecks defines the environment variable that is checked to determine whether healthchecks
 	// should be disabled. If this envvar is set to "all" or contains the string "xids", healthchecks are
 	// disabled entirely. If set, the envvar is treated as a comma-separated list of Xids to ignore. Note that
 	// this is in addition to the Application errors that are already ignored.
 	envDisableHealthChecks = "DP_DISABLE_HEALTHCHECKS"
-	envEnableHealthChecks  = "DP_ENABLE_HEALTHCHECKS"
-	allHealthChecks        = "xids"
+	// envEnableHealthChecks defines the environment variable that is checked to
+	// determine which XIDs should be explicitly enabled. XIDs specified here
+	// override the ones specified in the `DP_DISABLE_HEALTHCHECKS`.
+	// Note that this also allows individual XIDs to be selected when ALL XIDs
+	// are disabled.
+	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 
 // CheckHealth performs health checks on a set of devices, writing to the 'unhealthy' channel with any unhealthy devices
 func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devices, unhealthy chan<- *Device) error {
-	disableHealthChecks := strings.ToLower(os.Getenv(envDisableHealthChecks))
-	if disableHealthChecks == "all" {
-		disableHealthChecks = allHealthChecks
-	}
-	if strings.Contains(disableHealthChecks, "xids") {
+	skippedXids := getHealthCheckXids()
+	if skippedXids.IsAllDisabled() {
 		return nil
 	}
-
-	enableHealthChecks := strings.ToLower(os.Getenv(envEnableHealthChecks))
 
 	ret := r.nvml.Init()
 	if ret != nvml.SUCCESS {
@@ -61,31 +61,6 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 			klog.Infof("Error shutting down NVML: %v", ret)
 		}
 	}()
-
-	// FIXME: formalize the full list and document it.
-	// http://docs.nvidia.com/deploy/xid-errors/index.html#topic_4
-	// Application errors: the GPU should still be healthy
-	ignoredXids := []uint64{
-		13,  // Graphics Engine Exception
-		31,  // GPU memory page fault
-		43,  // GPU stopped processing
-		45,  // Preemptive cleanup, due to previous errors
-		68,  // Video processor exception
-		109, // Context Switch Timeout Error
-	}
-
-	skippedXids := make(map[uint64]bool)
-	for _, id := range ignoredXids {
-		skippedXids[id] = true
-	}
-
-	for _, additionalXid := range getAdditionalXids(disableHealthChecks) {
-		skippedXids[additionalXid] = true
-	}
-
-	for _, additionalXid := range getAdditionalXids(enableHealthChecks) {
-		delete(skippedXids, additionalXid)
-	}
 
 	klog.Infof("Health checks are disabled for xids: %v", skippedXids)
 
@@ -161,7 +136,7 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 			continue
 		}
 
-		if skippedXids[e.EventData] {
+		if skippedXids.IsNonFatal(e.EventData) {
 			klog.Infof("Skipping event %+v", e)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


elezar:
Thanks @robertdavidsmith. We have had some internal discussions around more flexibility here and I had the idea for https://github.com/NVIDIA/k8s-device-plugin/issues/1340 created against an internal issue tracker. I think your proposal here aligns with this to some extent and as such your input there would be appreciated.

internal/rm/health.go
```diff
@@ -32,6 +32,7 @@ const (
 	// disabled entirely. If set, the envvar is treated as a comma-separated list of Xids to ignore. Note that
 	// this is in addition to the Application errors that are already ignored.
 	envDisableHealthChecks = "DP_DISABLE_HEALTHCHECKS"
+	envEnableHealthChecks  = "DP_ENABLE_HEALTHCHECKS"
```
elezar: Can we rename this `DP_FORCE_HEALTHCHECKS` or `CRITICAL_XIDS` or `FATAL_XIDS` instead? (I would prefer the latter, but I could hear arguments for alternatives.

internal/rm/health.go
```diff
@@ -45,6 +46,8 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		return nil
 	}
 
+	enableHealthChecks := strings.ToLower(os.Getenv(envEnableHealthChecks))
```
elezar: Let's move this to wehre we actually use the value.

internal/rm/health.go
```diff
@@ -80,6 +83,12 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		skippedXids[additionalXid] = true
 	}
 
+	for _, additionalXid := range getAdditionalXids(enableHealthChecks) {
```
elezar: Does it make sense to allow something like `all` to be specified?

robertdavidsmith:
> Thanks @robertdavidsmith. We have had some internal discussions around more flexibility here and I had the idea for #1340 created against an internal issue tracker. I think your proposal here aligns with this to some extent and as such your input there would be appreciated.

Noted, I've commented on #1340

Do you see that issue as a replacement for this PR (in which case I'll close it), or in addition to it (in which case I'll update the PR to incorporate your feedback)?

internal/rm/health.go
robertdavidsmith: `DP_DISABLE_HEALTHCHECKS` is the exact opposite of `DP_ENABLE_HEALTHCHECKS` and IMO the naming should reflect this. So I'd favour renaming `DP_ENABLE_HEALTHCHECKS` to `DP_FORCE_ENABLE_HEALTHCHECKS`, that way you get the "force" but it's still very clear it's the exact opposite of `DP_DISABLE_HEALTHCHECKS`.

internal/rm/health.go
robertdavidsmith: Fair comment, though if we add support for "all" this all changes anyway.

internal/rm/health.go
robertdavidsmith: That's reasonable, I'm happy to do this

internal/rm/health.go
elezar: One question: What is the order of precedence of these? If `DP_DISABLE_HEALTHCHECKS=all|xids` does adding a specific XID to `DP_FORCE_ENABLE_HEALTHCHECKS` take precedence over this?

Then a comment on the naming. I don't think this envvar is the oposite of the existing one. While the existing one removes the predefined XIDs that we treat as fatal, this one adds ADDITIONAL ones.

elezar:
> > Thanks @robertdavidsmith. We have had some internal discussions around more flexibility here and I had the idea for #1340 created against an internal issue tracker. I think your proposal here aligns with this to some extent and as such your input there would be appreciated.
> 
> Noted, I've commented on #1340
> 
> Do you see that issue as a replacement for this PR (in which case I'll close it), or in addition to it (in which case I'll update the PR to incorporate your feedback)?

I think the core logic of what we're trying to achieve here is covered by #1340, but this PR calls out an omission in that issue which is that the set of "Critical" / "Forced" XIDs also need to be settable by an environment variable and not only in a config file. With that in mind, I think we can continue with your PR and then all that remains for #1340 would be to define the actual config format and map these to the relevant envvars using CLI flags like we do for other options.

Using our CLI handling to specify envvars also give us more flexibility to rename these envvars while falling back to the original. For the longest time, the `DP_ENABLE_HEALTHCHECKS` envvar has been considered an internal-only / undocumented features and if we pull this up to a more public configuration option (with support in helm etc.) a clearer name would be useful.

internal/rm/health.go
robertdavidsmith: On balance I'd make `DP_ADDITIONAL_HEALTHCHECKS` /`DP_FORCE_ENABLE_HEALTHCHECKS` support `all`, and have it override `DP_DISABLE_HEALTHCHECKS`. This matters more if we choose a name with `FORCE` in it, but is probably worth doing regardless of name. This would complicate the code a little, but would make the externally visible behavior to be what a reasonable user would expect.

robertdavidsmith:
Ok shall I make a PR to add `DP_ADDITIONAL_HEALTHCHECKS` that
- Will support `DP_ADDITIONAL_HEALTHCHECKS=all`
- `DP_ADDITIONAL_HEALTHCHECKS` always beats `DP_DISABLE_HEALTHCHECKS`
- Given increased complexity, I'll separate the enable/disable logic from the rest of `checkHealth()` so it can be properly unit-tested

You can then do #1340

Thoughts?

robertdavidsmith:
Just seen https://github.com/NVIDIA/k8s-device-plugin/pull/1360/files, that PR is largely complete and should be merged first.

elezar:
> Just seen https://github.com/NVIDIA/k8s-device-plugin/pull/1360/files, that PR is largely complete and should be merged first.

While #1360 does have some overlap with what you're doing here, I think that it could be seen as a follow up. I have created a branch https://github.com/elezar/k8s-device-plugin/tree/robertdavidsmith/main with a minor refactoring on top of your changes which should be compatible with what we want to do in 1360. What do you think?

robertdavidsmith:
> > Just seen https://github.com/NVIDIA/k8s-device-plugin/pull/1360/files, that PR is largely complete and should be merged first.
> 
> While #1360 does have some overlap with what you're doing here, I think that it could be seen as a follow up. I have created a branch https://github.com/elezar/k8s-device-plugin/tree/robertdavidsmith/main with a minor refactoring on top of your changes which should be compatible with what we want to do in 1360. What do you think?

Thank you @elezar 

I've commented on your commit [here](https://github.com/NVIDIA/k8s-device-plugin/commit/41e08a05e22b4c0d05b239bf26adc7af51ec1087), basically I think this makes a lot of sense, but see comments.

robertdavidsmith:
Hi @elezar,

How do you want to proceed from here?

Shall I copy your commit [here](https://github.com/NVIDIA/k8s-device-plugin/commit/41e08a05e22b4c0d05b239bf26adc7af51ec1087) onto my branch, make the changes I suggested (see comments on that commit), and then send you this PR for final review?

Thanks,

Rob

elezar:
Sorry for the delay here. If you think the changes proposed in the commit are reasonable, feel free to pull them in here so that we can continue reviewing your PR. It would be good for you to get the credit.

robertdavidsmith:
> Sorry for the delay here. If you think the changes proposed in the commit are reasonable, feel free to pull them in here so that we can continue reviewing your PR. It would be good for you to get the credit.

Ok, I'm happy to do so

elezar:
In order to close this out I have created https://github.com/NVIDIA/k8s-device-plugin/pull/1443. Please take a look.

robertdavidsmith:
@elezar PR #1443 LGTM, thank you. I will close this PR

elezar:
> @elezar PR #1443 LGTM, thank you. I will close this PR

Thanks again for your work here!

PR #1339, opened 2025-08-01
Add support for GDRCOPY

elezar:
This change adds support for GDRCOPY through the same mechansim as for GDS and MOFED. This requires new functionality in the nvidia-container-toolkit to generated CDI specs for gdrcopy devices when using CDI mode.

This depends on https://github.com/NVIDIA/nvidia-container-toolkit/pull/1230


cmd/nvidia-device-plugin/main.go
```diff
@@ -100,6 +100,11 @@ func main() {
 			Usage:   "the desired strategy for passing device IDs to the underlying runtime:\n\t\t[uuid | index]",
 			EnvVars: []string{"DEVICE_ID_STRATEGY"},
 		},
+		&cli.BoolFlag{
+			Name:    "gdrcopy-enabled",
+			Usage:   "ensure that containers are started with NVIDIA_GDRCOPY=enabled",
```
cdesiniotis: Question -- since the `NVIDIA_GDRCOPY=enabled` envvar only gets set when a non-CDI device list strategy is configured, would it make sense to omit this implementation detail from the usage string? What about a more generic usage string, like "ensure that containers get access to the GDRCopy device nodes". This comment also applies to the `gds-enabled` and `mofed-enabled` flags.

internal/cdi/options.go
```diff
@@ -80,6 +80,13 @@ func WithVendor(vendor string) Option {
 	}
 }
 
+// WithGdrcopyEnabled provides and option to set whether a GDS CDI spec should be generated
```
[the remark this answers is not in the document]
cdesiniotis: +1

cdesiniotis (dismissed):
Some minor comments. LGTM otherwise.

cmd/nvidia-device-plugin/main.go
elezar: Updated all usage strings.

cdesiniotis (approved):
Thanks @elezar LGTM.

PR #1356, opened 2025-08-12
Ignore errors getting optional device attributes

elezar:
On certain systems (e.g. iGPU-based systems), the NVML nvmlDeviceGetMemoryInformation API
is not supported and returns an error. In these cases we ignore
these errors and log a warning instead. This means that:

* For the GPU Device Plugin, memory limits will be enforced for MPS partioning.
* For GFD, no nvidia.com/gpu.memory label will be generated.

### DIFF
--- a/internal/rm/devices.go
+++ b/internal/rm/devices.go
@@ -21,6 +21,7 @@ import (
 	"strconv"
 	"strings"
 
+	"k8s.io/klog/v2"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
@@ -73,7 +74,7 @@ func BuildDevice(index string, d deviceInfo) (*Device, error) {
 
 	totalMemory, err := d.GetTotalMemory()
 	if err != nil {
-		return nil, fmt.Errorf("error getting device memory: %w", err)
+		klog.Warningf("Ignoring error getting device memory: %v", err)
 	}
 
 	computeCapability, err := d.GetComputeCapability()
--- a/internal/lm/resource.go
+++ b/internal/lm/resource.go
@@ -21,6 +21,8 @@ import (
 	"regexp"
 	"strings"
 
+	"k8s.io/klog/v2"
+
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 	"github.com/NVIDIA/k8s-device-plugin/internal/resource"
 )
@@ -46,7 +48,7 @@ func NewGPUResourceLabeler(config *spec.Config, device resource.Device, count in
 
 	totalMemoryMiB, err := device.GetTotalMemoryMiB()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get memory info for device: %v", err)
+		klog.Warningf("Ignoring error getting memory info for device: %v", err)
 	}
 
 	resourceLabeler := newResourceLabeler(fullGPUResourceName, config)


internal/rm/devices.go
```diff
@@ -73,12 +74,12 @@ func BuildDevice(index string, d deviceInfo) (*Device, error) {
...
 	totalMemory, err := d.GetTotalMemory()
 	if err != nil {
-		return nil, fmt.Errorf("error getting device memory: %w", err)
+		klog.Warningf("Ignoring error getting device memory: %w", err)
 	}
 
 	computeCapability, err := d.GetComputeCapability()
 	if err != nil {
-		return nil, fmt.Errorf("error getting device compute capability: %w", err)
+		klog.Warningf("Ignoring error getting device compute capability: %w", err)
```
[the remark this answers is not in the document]
ArangoGutierrez: This is not correct, klog is a wrapper for fmt.Printf , see 

https://github.com/kubernetes/klog/blob/e7125f792ea66a85818cfb45261c9e1acc585344/klog.go#L740C1-L761C2

ArangoGutierrez (approved):
lgtm

internal/resource/nvml-device.go
```diff
@@ -84,6 +85,17 @@ func (d nvmlDevice) GetName() (string, error) {
...
+	if ret == nvml.ERROR_NOT_SUPPORTED {
+		fs, err := procfs.NewDefaultFS()
+		if err != nil {
+			return 0, err
+		}
+		memInfo, err := fs.Meminfo()
+		if err != nil {
+			return 0, err
+		}
+		return *memInfo.MemTotal / (1024 * 1024), nil
```
ArangoGutierrez: Could we add a comment detailing why `1042 * 1024`

cmd/mps-control-daemon/mps/device.go
```diff
@@ -50,6 +50,9 @@ func (d *mpsDevice) maxClients() int {
 
 // isAtLeastVolta checks whether the specified device is a volta device or newer.
 func (d *mpsDevice) isAtLeastVolta() bool {
+	if d.ComputeCapability == "" {
+		return true
```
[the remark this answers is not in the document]
tariq1890: I agree with this review comment, let's add a comment explaining why we return true here

internal/resource/nvml-device.go
```diff
@@ -84,6 +85,17 @@ func (d nvmlDevice) GetName() (string, error) {
 // GetTotalMemoryMiB returns the total memory on a device in mebibytes (2^20 bytes)
 func (d nvmlDevice) GetTotalMemoryMiB() (uint64, error) {
 	info, ret := d.GetMemoryInfo()
+	if ret == nvml.ERROR_NOT_SUPPORTED {
```
tariq1890: Can we factor this out into a new method?

internal/rm/devices.go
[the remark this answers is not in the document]
cdesiniotis: While using `%w` is syntactically valid, I would argue wrapping errors is no longer required / necessary as we are not returning the error.

cmd/mps-control-daemon/mps/device.go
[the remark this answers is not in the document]
cdesiniotis: A comment would help. It is not clear to me that `true` is the "right" default here.

internal/resource/nvml-device.go
cdesiniotis: Question -- is there an additional check we can perform before falling back to reading `/proc/meminfo`? Specifically, is there any way to identify we are on this type of SOC system where the total system memory is representative of the GPU memory?

internal/rm/devices.go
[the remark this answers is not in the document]
cdesiniotis: I see this was changed to `%v` in a separate commit.

internal/resource/nvml-device.go
tariq1890: > The magic numbers 1024 * 1024 should be extracted to a named constant like bytesToMebibytes 

I actually like this review comment. Moving that to a constant will make it self-explanatory

internal/resource/nvml-device.go
elezar: I asked whether there is a specific API and got the following response:

> Maybe we could use the return code from nvmlDeviceGetMemoryInfo to determine if we need to query /proc/meminfo instead since that would return NOT SUPPORTED for iGPU platforms?

I'm happy to pull this change out of this PR and handle that as a follow up. I think correct labels and MPS memory partitioning is a lower priority than not failing.

elezar: Yes, I will. I have removed this from this changeset though and will created a follow up to readd this properly.

elezar: I have updated this PR to only skip errors and not update the mechansim used to extract memory information.

cmd/mps-control-daemon/mps/device.go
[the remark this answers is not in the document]
elezar: Looking at the behaviour of the target system, I noted that the NVML API to get CUDA Compute Capability is functional. I have reverted this change.

PR #1362, opened 2025-08-13
Futher cleanup of distroless

elezar:
These changes further clean up the container image following the move to distroless images.

This includes:
* Removing an unused `minimal` image stage
* Switching to a golang base image for builds
* Defining the `devel` image as a separate stage in the application Dockerfile to reuse the single base image definition.


deployments/container/Dockerfile
```diff
@@ -12,29 +12,22 @@
...
+FROM base AS devel
+
+WORKDIR /work
+COPY * .
+
+RUN make install-tools
+
+# We need to set the /work directory as a safe directory.
+# This allows git commands to run in the container.
+RUN git config --file=/.gitconfig --add safe.directory /work
```
tariq1890, line 27: Why do we need to run git commands within the container?

cdesiniotis: To fetch the commit sha: https://github.com/NVIDIA/k8s-device-plugin/blob/c87ad4047491d18999d666a9543bd0d134b86122/versions.mk#L30

deployments/container/Dockerfile
```diff
@@ -14,6 +14,18 @@
 
 FROM golang:1.24.6 AS base
 
+# The devel image is used for local development.
+FROM base AS devel
+
+WORKDIR /work
+COPY * .
+
+RUN make install-tools
```
cdesiniotis, line 23: Is this a target of the top-level Makefile? I believe this make target is only defined in `deployments/devel`

cdesiniotis: Okay I believe I see how this works now. The `DOCKERFILE_CONTEXT` is set to `deployments/devel` so this will invoke the Makefile in that directory...

elezar: Yes, the context defines where this is run.

deployments/container/Dockerfile
elezar: This is the `devel` image. Without this step go builds fail when running a `.shell` target.

PR #1330, opened 2025-07-23
Fix internal CI pipelines

elezar:
These changes include minor cleanups to the internal CI pipelines. Using variables consistent with other projects and removing unneeded stages.

PR #1242, opened 2025-04-24
[no-relnote] Update E2E test suite

ArangoGutierrez:
This pull request introduces significant updates to the end-to-end (E2E) testing framework for the NVIDIA Kubernetes Device Plugin. The changes include enhancements to the CI workflow, a migration to the Ginkgo v2 testing framework, restructuring of test-related files, and updates to licensing information. Below is a categorized summary of the most important changes:

### Workflow and CI Updates:
* Added a new `HELM_CHART` environment variable to the E2E CI workflow (`.github/workflows/e2e.yaml`) and updated the `make` command to use a specific Makefile for E2E tests. This improves configurability and aligns the workflow with the new test structure.
* Replaced the Slack alert notification step with a new Ginkgo log archiving step and updated the Slack notification configuration to use a JSON payload for better integration.

### Migration to Ginkgo v2:
* Updated the `Makefile` to include a `ginkgo` target for installing the Ginkgo v2 CLI and a new `test-e2e` target for running tests with Ginkgo. This replaces the previous custom test logic.
* Renamed and restructured test files to align with Ginkgo's conventions, such as renaming `common/kubernetes.go` to `cleanup_test.go` and removing redundant utility functions. [[1]](diffhunk://#diff-ff8edf4a90c068e14f3d05c34333264d6182cba862745973d3d76f95ffc432b3L2-R3) [[2]](diffhunk://#diff-ff8edf4a90c068e14f3d05c34333264d6182cba862745973d3d76f95ffc432b3L16-R17)

### Documentation Enhancements:
* Added a comprehensive `README.md` for the E2E test suite, detailing prerequisites, environment variables, execution flow, and troubleshooting steps. This improves developer onboarding and test maintainability.

### Licensing and Copyright:
* Updated all test-related files to use the SPDX license headers, reflecting the transition to a 2025 copyright year and the Apache 2.0 license. [[1]](diffhunk://#diff-16221b5a73a49f3b097836531eab77e6659f94ae31aa6eaabb20a13295da7be7L1-R2) [[2]](diffhunk://#diff-ff8edf4a90c068e14f3d05c34333264d6182cba862745973d3d76f95ffc432b3L2-R3) [[3]](diffhunk://#diff-6b3f18281211686d71e2a67b603b989feb390729ad71df9304ba1b197fcea26cL2-R3)

### Codebase Simplification:
[...]

PR #1363, opened 2025-08-13
Switch to single ci definition

elezar:
This follows on from https://github.com/NVIDIA/k8s-device-plugin/pull/1230 and https://github.com/NVIDIA/k8s-device-plugin/pull/1330 and applies the same changes that were made for the nvidia-container-toolkit and k8s-dra-driver-gpu to switch to a SINGLE file to serve as the internal CI definition.


jgehrcke (approved):
nice! :)

elezar:
The CI is red and should be addressed by #1242

PR #1365, opened 2025-08-13
[Helm] Supports HostNetwork Mode for each components

ssup2:


ssup2:
@elezar Hello. Please review this pr?

ssup2:
@elezar Hi! I'd appreciate it if you could take a look at this PR when you get a moment.

rahulait:
Thanks @ssup2 . Can you please also add default value to values.yaml so that its easier to find what one can set. So something like:

```
# In the devicePlugin section (~line 103):
devicePlugin:
  enabled: true
  # enableHostNetwork when set to true runs the pod with hostNetwork: true.
  # This can be useful in non-overlay network environments (e.g., AWS, Azure, GCP)
  # to avoid unnecessary IP allocation for daemon pods.
  enableHostNetwork: false

```

Similar change for gfd and mps as well.

rahulait:
Gentle reminder @ssup2 . Please update the values.yaml with default values so that its easier to find how it can be configured.

karthikvetrivel (approved):

PR #1372, opened 2025-08-18
Always use dedicated service account in device-plugin helm chart

cdesiniotis:
Before this change, installing the helm chart with default values would lead to the `default` service account being used. Using the `default` service account is generally discouraged as it is a shared service account that gets used by all pods in the same namespace that don't specify one. If the `default` service account was compromised in some way, e.g. given more privileges, then our device-plugin would also gain those privileges and we would not be adhering to the principle of least privilege.

This PR updates our helm chart so that a dedicated service account is always created and used.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
@@ -54,11 +54,9 @@ spec:
       {{- end }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      {{- if $options.hasConfigMap }}
       serviceAccountName: {{ include "nvidia-device-plugin.fullname" . }}-service-account
       {{- if not .Values.mps.enableHostPID }}
       shareProcessNamespace: true
-      {{- end }}
       {{- end }}
       {{- if .Values.mps.enableHostPID }}
       hostPID: true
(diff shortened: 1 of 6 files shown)


deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
```diff
@@ -50,12 +50,10 @@ spec:
       {{- end }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      {{- if $options.hasConfigMap }}
       serviceAccountName: {{ include "nvidia-device-plugin.fullname" . }}-service-account
       {{- if not .Values.mps.enableHostPID }}
       shareProcessNamespace: true
       {{- end }}
-      {{- end }}
```
elezar, line 62: Question: Was this a bug? Were we only adding `shareProcessNamespace` if we ALSO had a config map?

elezar (approved):
Just a question regarding the HOST PID setting for MPS. (Fixing that may justify a separate commit for the changelogs).

ArangoGutierrez:
/ok to test [d30d159](https://github.com/NVIDIA/k8s-device-plugin/pull/1372/commits/d30d15900ba150076ad3c5ad0f8de8a1fbfc8147)

deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml
rajatchopra: I think shareProcessName was a slip anyways. Looks good now.

rajatchopra (approved):

rajatchopra:
@cdesiniotis This is ready to go. Rebase and merge?

tariq1890 (approved):

PR #1374, opened 2025-08-19
Ignore errors getting device memory using NVML

elezar:
On certain systems, the NVML nvmlDeviceGetMemoryInformation API is not supported and returns an error. In these cases we ignore these errors and log a warning instead. This means that:

* For the GPU Device Plugin, memory limits will be enforced for MPS partioning.
* For GFD, no nvidia.com/gpu.memory label will be generated.

Backport of #1356

### DIFF
--- a/internal/lm/resource.go
+++ b/internal/lm/resource.go
@@ -21,6 +21,8 @@ import (
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (approved):
LGTM-Backport

PR #1413, opened 2025-09-09
Bump NFD Helm to v0.17.3

ArangoGutierrez:


elezar (commented):
As a general question: In the context of the GPU Operator we have discussed listing NFD as an EXPLICIT prerequisite instead of including the subcharts. Is this something that makes sense for the device plugin too?

deployments/static/nfd.yaml
elezar, line 1: We don't actually mention this file in any of our documentation. Can we rather remove it?

elezar (changes requested):
Let's remove the static deployment for NFD to further reduce the maintenance burden for this repo.

PR #1418, opened 2025-09-15
Add pod webhook for set runtimeclass

luohua13:


elezar:
Hi @luohua13. Thanks for the contribution.

Although a webhook to set the runtime class is something that we have discussed, we are not ready to offer it as an option to customers. Althoug the implementation is relatively straightforward, the overhead of ensuring compatibility with our support matrix is not something that we can commit to at the moment.

PR #1467, opened 2025-10-21
enable device plugin specific node selector

omrishiv:
This PR enables device plugin specific node selectors. The idea is to retain the global functionality of `.Values.nodeSelector`, but allow merging device plugin specific node selectors. Bottlerocket does not need the device plugin, but benefits from the rest of daemonsets. Right now, deploying as-is results in the device plugin constantly crashing on Bottlerocket nodes.


deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -102,6 +102,7 @@ runtimeClassName: null
 
 devicePlugin:
   enabled: true
+  nodeSelector: {}
```
elezar, line 105: My one question would be whether we wanted to do this for other top-level placement selectors such as `affinity` and also do the same for `gfd`?

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -18,6 +18,7 @@
 {{- $useServiceAccount := $options.hasConfigMap }}
 {{- $configMapName := (include "nvidia-device-plugin.configMapName" .) | trim }}
 {{- $daemonsetName := printf "%s" (include "nvidia-device-plugin.fullname" .) | trunc 63 | trimSuffix "-" }}
+{{- $devicePluginNodeSelector := merge .Values.devicePlugin.nodeSelector .Values.nodeSelector }}
```
elezar, line 21: Is merging what we want here? Should we not use `devicePlugin.nodeSelector` if specified and fallback to the top-level `nodeSelector` if it is not.

omrishiv: let me give a better example of what we have:

| - | node 1 | node 2 | node 3 |
| - | - | - | - |
|accelerator| nvidia | nvidia | aws |
| ami | al2023 | bottlerocket | bottlerocket |

- If we don't have any nodeselectors, we get pods scheduled on all nodes, but we don't need it on nodes 2 and 3
- If we have nodeSelector for accelerator: nvidia, we get it on 1 and 2, but we don't need it on 2.

We do need it on 1, but we don't want it to schedule on any other node. This attempts to maintain backwards compatibility while filtering out nodes that don't need it. The idea is having a global selector (accelerator) and filtering using a device plugin specific one (ami)

deployments/helm/nvidia-device-plugin/values.yaml
omrishiv: gfd is useful for all nvidia nodes, so I didn't see a need to extend it, but I'm happy to add it so it's usable if you want it

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
tariq1890: I agree with Evan here. Wouldn't specifying the exact nodeSelector in `.Values.devicePlugin.nodeSelector` solve your use case? Relying on a `merge` tends to be a slippery slope.

omrishiv: I think we've managed to work around the issue. This can be closed, thank you

tariq1890:
Closing as @omrishiv has reported that he has found an alternate way to solve his problem.

PR #1481, opened 2025-10-29
Helm fix identical template selector

peterzandbergen:
The helm chart used the same labels for the selector in the daemon sets 

I have fixed this for all daemon sets, being the mps and gfd.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/Chart.yaml
+++ b/deployments/helm/nvidia-device-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: nvidia-device-plugin
 type: application
 description: A Helm chart for the nvidia-device-plugin on Kubernetes
-version: "0.18.0"
+version: "0.18.1"
 appVersion: "0.18.0"
 kubeVersion: ">= 1.10.0-0"
 home: https://github.com/NVIDIA/k8s-device-plugin
--- a/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
+++ b/deployments/helm/nvidia-device-plugin/templates/_helpers.tpl
@@ -66,6 +66,15 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 {{- end }}
 
+# PEZA: Added to fix identical template matchLabels
+{{- define "nvidia-device-plugin-mps-control-daemon.templateLabels" -}}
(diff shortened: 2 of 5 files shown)


uristernik:
@elezar can someone have a look here?

npdgm:
Per [Helm Standard Labels](https://helm.sh/fr/docs/chart_best_practices/labels/#standard-labels) I think we should keep common `app.kubernetes.io/name` and `app.kubernetes.io/instance` for all pod controllers, and instead introduce `app.kubernetes.io/component` to each DaemonSet. Also templates won't need repetition and can then call a common one for `name` and `component`.

rahulait:
Thanks @peterzandbergen. Can you make sure commits are signed and fix the merge conflicts.

deployments/helm/nvidia-device-plugin/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: nvidia-device-plugin
 type: application
 description: A Helm chart for the nvidia-device-plugin on Kubernetes
-version: "0.18.0"
+version: "0.18.1"
```
rahulait, line 5: Lets not bump the version. We bump it when we want to release.

rahulait:
> Per [Helm Standard Labels](https://helm.sh/fr/docs/chart_best_practices/labels/#standard-labels) I think we should keep common app.kubernetes.io/name and app.kubernetes.io/instance for all pod controllers, and instead introduce app.kubernetes.io/component to each DaemonSet. Also templates won't need repetition and can then call a common one for name and component.

Yeah, agree with this. We can just introduce additional `app.kubernetes.io/component: {{ .component }}` for each daemonset.

So each daemonsets / pods labels will be like:
```
# device-plugin
app.kubernetes.io/name: nvidia-device-plugin
app.kubernetes.io/instance: my-release
app.kubernetes.io/component: device-plugin

# gfd
app.kubernetes.io/name: nvidia-device-plugin
app.kubernetes.io/instance: my-release
app.kubernetes.io/component: gpu-feature-discovery

# mps
app.kubernetes.io/name: nvidia-device-plugin
app.kubernetes.io/instance: my-release
app.kubernetes.io/component: mps-control-daemon
```

rahulait:
Closing this PR in favor of https://github.com/NVIDIA/k8s-device-plugin/pull/1716

PR #1504, opened 2025-11-17
Enable NodeFeature API by default in GFD

rajathagasthya:
This change enables NodeFeature API in GFD. This results in GFD creating
in a new NodeFeature CR on startup.

This commit also removes the redundant `enableNodeFeatureApi` helm value
NodeFeature is now enabled by default in NFD since it's v0.17.0 release.
The k8s-device-plugin helm chart has been using NFD v0.17.3 subchart
since k8s-device-plugin's v0.18.0 release, so NodeFeature API has been
enabled by default since then and this helm value has had no effect.

Verification steps:

1. Deploy gpu-operator that uses GFD image built by this PR.
2. Verify new `NodeFeature` CR is created.
3. Verify no file is added by GFD in `/etc/kubernetes/node-feature-discovery/features.d/`.

### DIFF
--- a/cmd/gpu-feature-discovery/main.go
+++ b/cmd/gpu-feature-discovery/main.go
@@ -99,6 +99,7 @@ func main() {
 		},
 		&cli.BoolFlag{
 			Name:    "use-node-feature-api",
+			Value:   true,
 			Usage:   "Use NFD NodeFeature API to publish labels",
 			EnvVars: []string{"GFD_USE_NODE_FEATURE_API", "USE_NODE_FEATURE_API"},
 		},
--- a/deployments/helm/nvidia-device-plugin/values.yaml
+++ b/deployments/helm/nvidia-device-plugin/values.yaml
@@ -118,7 +118,6 @@ gfd:
 # Helm dependency
 nfd:
   nameOverride: node-feature-discovery
-  enableNodeFeatureApi: false
   master:
     serviceAccount:
       name: node-feature-discovery
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
@@ -15,7 +15,7 @@
 {{- if .Values.gfd.enabled }}
 ---
 {{- $options := (include "nvidia-device-plugin.options" . | fromJson) }}
-{{- $useServiceAccount := or ( $options.hasConfigMap ) ( and .Values.gfd.enabled .Values.nfd.enableNodeFeatureApi ) }}
+{{- $useServiceAccount := or $options.hasConfigMap .Values.gfd.enabled }}
 {{- $configMapName := (include "nvidia-device-plugin.configMapName" .) | trim }}
 {{- $daemonsetName := printf "%s-gpu-feature-discovery" (include "nvidia-device-plugin.fullname" .) | trunc 63 | trimSuffix "-" }}
 apiVersion: apps/v1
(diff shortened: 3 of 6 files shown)


deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -118,7 +118,7 @@ gfd:
 # Helm dependency
 nfd:
   nameOverride: node-feature-discovery
-  enableNodeFeatureApi: false
+  enableNodeFeatureApi: true
```
elezar: Does the behaviour differ between upgrades and new installations due to this toggle?

rajathagasthya: It should automatically upgrade to using NFD when a new chart and device plugin image is used, so I don't think the behavior differs. Is there a particular scenario you were thinking of?

cdesiniotis: I believe the answer is yes, but just to confirm -- if we upgrade the device-plugin helm chart, does the NFD subchart also get upgraded and the value of this field, `enableNodeFeatureApi`, gets toggled from `false` to `true`?

rajathagasthya: @elezar Do you have any thoughts on if we should keep this value enabled by default?

cdesiniotis: @rajathagasthya have you tried a helm chart upgrade? What behavior have you observed?

rajathagasthya: @elezar The `enableNodeFeatureApi` variable is actually redundant and can be removed. The reason is that `v0.18.0` of `k8s-device-plugin` upgraded NFD to `v0.17.3`, which enables `NodeFeature` API by default. So this behavior was already changed when older charts upgraded to `v0.18.0`.

I've verified this by doing the following:

1. Install `v0.17.4` of `k8s-device-plugin`. This installs NFD `v0.15.3` subchart.
2. No `NodeFeature` CRs are created since that version of NFD doesn't enable `NodeFeature` API by default (controlled by `nfd.enableNodeFeatureApi` in `values.yaml`).
3. Upgrade the chart to `v0.18.0` of `k8s-device-plugin`. This installs NFD `v0.17.3` subchart.
4. A `NodeFeature` CR is created by default since it's now enabled in NFD.

Essentially, `nfd.enableNodeFeatureApi` has been a no-op since the last release of `k8s-device-plugin` and can be removed.

Bonus: I've also verified that when I upgrade the chart to current changes in this, it creates two `NodeFeature` CRs — one created by NFD and another by GFD.

cmd/gpu-feature-discovery/main.go
```diff
@@ -99,6 +99,7 @@ func main() {
 		},
 		&cli.BoolFlag{
 			Name:    "use-node-feature-api",
```
rajathagasthya, line 101: Should we deprecate this flag too for future removal?

deployments/helm/nvidia-device-plugin/values.yaml
rajathagasthya: I've removed `nfd.enableNodeFeatureApi` value. PTAL.

cmd/gpu-feature-discovery/main.go
ArangoGutierrez: I think is sane to do, in a couple more releases, NFD will start deprecating the old method.

ArangoGutierrez: I would say a depreciation path will involve not fully removing `enableNodeFeatureApi` from the Helm charts but adding a logic to helm helper to print a warning mentioning that the new default and the flag will go away soon

rajathagasthya: I removed `enableNodeFeatureApi` from the helm chart because that value is not functional right now. And giving users the option to set `enableNodeFeatureApi=false` will actually leave the deployment broken because it doesn't create the necessary role/rolebinding/serviceaccount objects in the chart. So I'm not sure adding a warning there is of much help.

Let me know what you think!

ArangoGutierrez: I am ok with removing it, NFD has been with `NodeFeatureApi` enabled by default for a long time now, most users must be on it already.

PR #1508, opened 2025-11-18
Refactor checkHealth function

ArangoGutierrez:
This patch refactors the device health check system by extracting the logic into a dedicated HealthProvider interface with proper lifecycle management using WaitGroups and context. 
No behavior changes - this is a pure refactoring to improve code modularity and testability.

### DIFF
--- a/internal/rm/health.go
+++ b/internal/rm/health.go
@@ -32,18 +32,16 @@ import (
 )
 
 const (
-	// envDisableHealthChecks defines the environment variable that is
-	// checked to determine whether healthchecks should be disabled. If
-	// this envvar is set to "all" or contains the string "xids",
-	// healthchecks are disabled entirely. If set, the envvar is treated
-	// as a comma-separated list of Xids to ignore. Note that this is in
-	// addition to the Application errors that are already ignored.
+	// envDisableHealthChecks defines the environment variable that is checked to determine whether healthchecks
+	// should be disabled. If this envvar is set to "all" or contains the string "xids", healthchecks are
+	// disabled entirely. If set, the envvar is treated as a comma-separated list of Xids to ignore. Note that
+	// this is in addition to the Application errors that are already ignored.
 	envDisableHealthChecks = "DP_DISABLE_HEALTHCHECKS"
-	// envEnableHealthChecks defines the environment variable that is
-	// checked to determine which XIDs should be explicitly enabled. XIDs
-	// specified here override the ones specified in the
-	// `DP_DISABLE_HEALTHCHECKS`. Note that this also allows individual
-	// XIDs to be selected when ALL XIDs are disabled.
+	// envEnableHealthChecks defines the environment variable that is checked to
+	// determine which XIDs should be explicitly enabled. XIDs specified here
+	// override the ones specified in the `DP_DISABLE_HEALTHCHECKS`.
+	// Note that this also allows individual XIDs to be selected when ALL XIDs
+	// are disabled.
 	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 


internal/plugin/server.go
```diff
@@ -155,6 +172,9 @@ func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
 		}
 	}()
 
+	// Start recovery worker to detect when unhealthy devices become healthy
+	go plugin.runRecoveryWorker()
```
elezar: Can we split the refactoring (that doesn't add any new behaviour) into a different PR from the one that adds devices becoming healthy again?

ArangoGutierrez: sounds like a good idea, and even more based on your other comment https://github.com/NVIDIA/k8s-device-plugin/pull/1508#pullrequestreview-3488039792 
I wanted a re-factor, but that interface is a diff conversation. Going to work on splitting this PR

internal/plugin/server.go
```diff
@@ -269,13 +275,19 @@ func (plugin *nvidiaDevicePlugin) ListAndWatch(e *pluginapi.Empty, s pluginapi.D
 		return err
 	}
 
+	// If health provider not available, wait for context cancellation
+	if plugin.healthProvider == nil {
+		<-plugin.ctx.Done()
+		return nil
+	}
```
elezar: Under which conditions is the healthProvider `nil`? Could we not rather ALWAYS use at least a "no-op" healthProvider to ensure that we don't need to special case this here or at any point where we call `Start` or `Stop`?

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
-	// envDisableHealthChecks defines the environment variable that is checked to determine whether healthchecks
-	// should be disabled. If this envvar is set to "all" or contains the string "xids", healthchecks are
-	// disabled entirely. If set, the envvar is treated as a comma-separated list of Xids to ignore. Note that
-	// this is in addition to the Application errors that are already ignored.
+	// envDisableHealthChecks defines the environment variable that is
+	// checked to determine whether healthchecks should be disabled. If
+	// this envvar is set to "all" or contains the string "xids",
+	// healthchecks are disabled entirely. If set, the envvar is treated
+	// as a comma-separated list of Xids to ignore. Note that this is in
+	// addition to the Application errors that are already ignored.
```
elezar: This is a nit: For complex refactorings, keeping changes to a minimum is important as we are able to reduce the noise and focus on the changes. In cases like these, we should update these comments as a separate commit.

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
+func NewNVMLHealthProvider(
+	nvml nvml.Interface,
+	config *spec.Config,
+	devices Devices,
+) HealthProvider {
+	return &nvmlHealthProvider{
+		nvml:       nvml,
+		config:     config,
+		devices:    devices,
+		healthChan: make(chan *Device, 64),
```
elezar: Question: Why 64?

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
+// Start initializes NVML, registers event handlers, and starts the
+// monitoring goroutine. Blocks until initialization completes.
+func (p *nvmlHealthProvider) Start(ctx context.Context) error {
+	p.mu.Lock()
+	if p.started {
+		p.mu.Unlock()
+		return fmt.Errorf("health provider already started")
+	}
+	p.started = true
+	p.mu.Unlock()
```
elezar: Any reason to not `defer p.mu.Unlock()` instead?

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
+	// NVML resources
+	eventSet nvml.EventSet
+
+	// Lifecycle management
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+
+	// State guards
+	mu      sync.Mutex
```
elezar: We could use an `IsA` relationship to simplify taking and releasing the lock:

```suggestion
	sync.Mutex
```

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
+		klog.Info("Health checks disabled via DP_DISABLE_HEALTHCHECKS")
 		return nil
 	}
 
-	ret := r.nvml.Init()
+	// Initialize NVML
+	ret := p.nvml.Init()
 	if ret != nvml.SUCCESS {
-		if *r.config.Flags.FailOnInitError {
+		if *p.config.Flags.FailOnInitError {
```
elezar: nit: Let's not rename `r` to `p` in a single commit. (see comment on managing diffs).

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
+	}
+	p.started = true
+	p.mu.Unlock()
+
+	// Check if health checks are disabled
+	p.xidsDisabled = getDisabledHealthCheckXids()
+	if p.xidsDisabled.IsAllDisabled() {
+		klog.Info("Health checks disabled via DP_DISABLE_HEALTHCHECKS")
 		return nil
 	}
```
elezar: This should happen at construction and not as `Start` is called. If all healthChecks are disabled, we should return a `no-op` `HealthProvider`.

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
+	ret := p.nvml.Init()
 	if ret != nvml.SUCCESS {
-		if *r.config.Flags.FailOnInitError {
+		if *p.config.Flags.FailOnInitError {
 			return fmt.Errorf("failed to initialize NVML: %v", ret)
 		}
+		klog.Warningf("NVML init failed: %v; health checks disabled", ret)
 		return nil
 	}
-	defer func() {
```
elezar, line 57: Could you explain the move away from a deferred shutdown?

internal/rm/health.go
```diff
@@ -17,189 +17,460 @@
...
-	eventSet, ret := r.nvml.EventSetCreate()
+	// Create event set
+	eventSet, ret := p.nvml.EventSetCreate()
 	if ret != nvml.SUCCESS {
+		if shutdownRet := p.nvml.Shutdown(); shutdownRet != nvml.SUCCESS {
+			klog.Warningf("Failed to shutdown NVML: %v", shutdownRet)
+		}
 		return fmt.Errorf("failed to create event set: %v", ret)
 	}
-	defer func() {
```
elezar, line 70: is there a reason that we don't use the deferred cleanup here?

internal/rm/health.go
```diff
@@ -276,51 +548,8 @@ func newHealthCheckXIDs(xids ...string) disabledXIDs {
 	return output
 }
 
-// getDevicePlacement returns the placement of the specified device.
-// For a MIG device the placement is defined by the 3-tuple <parent UUID, GI, CI>
-// For a full device the returned 3-tuple is the device's uuid and 0xFFFFFFFF for the other two elements.
-func (r *nvmlResourceManager) getDevicePlacement(d *Device) (string, uint32, uint32, error) {
```
elezar, line 282: Why move the implementation?

internal/rm/tegra_manager.go
```diff
@@ -70,7 +71,29 @@ func (r *tegraResourceManager) GetDevicePaths(ids []string) []string {
 	return nil
 }
 
-// CheckHealth is disabled for the tegraResourceManager
-func (r *tegraResourceManager) CheckHealth(stop <-chan interface{}, unhealthy chan<- *Device) error {
+// HealthProvider returns a no-op HealthProvider for Tegra devices.
+// Tegra devices do not support health monitoring.
+func (r *tegraResourceManager) HealthProvider() HealthProvider {
+	return &noopHealthProvider{}
```
elezar, line 77: OK. You already have it. Why not use it?

internal/rm/tegra_manager.go
```diff
@@ -70,7 +71,29 @@ func (r *tegraResourceManager) GetDevicePaths(ids []string) []string {
...
+}
+
+// noopHealthProvider is a no-op implementation for platforms that don't
+// support health monitoring.
+type noopHealthProvider struct {
+	healthChan chan *Device
+}
+
+func (n *noopHealthProvider) Start(context.Context) error {
+	n.healthChan = make(chan *Device)
```
elezar: Why not just do this at construction?

elezar: Also, do we need to actually create a channel? Can we no leave it `nil`?

internal/rm/health.go
ArangoGutierrez: `defer` would be simpler but slower. Using `defer` would hold the mutex during {NVML initialization, Event set creation, Device registration}, blocking other operations (like Stop()).

internal/rm/health.go
ArangoGutierrez: size would be len(devices) × 4, but I thought 64 was a safe hard coded number as it covers all possible len(devices) sizes

internal/rm/health.go
ArangoGutierrez: All error paths after Init() have now individual clean up logic.

internal/rm/health.go
ArangoGutierrez: All error paths after Init() have now individual clean up logic.

internal/rm/health.go
ArangoGutierrez: now in a `[no-relnote]` commit

internal/rm/health.go
ArangoGutierrez: fixed, moved to it's original position

internal/rm/health.go
ArangoGutierrez: thanks for the suggestion, adopted

internal/rm/health.go
ArangoGutierrez: thanks for the suggestion, adopted , now an independent commit

internal/rm/health.go
ArangoGutierrez: thanks for the suggestion, adopted

internal/plugin/server.go
ArangoGutierrez: thanks for the suggestion, adopted

internal/rm/tegra_manager.go
ArangoGutierrez: now used as proposed in above comment

internal/rm/health.go
elezar: I don't agree that "slowness" is something we should optimize for. `Start()` and `Stop()` should not be running concurrently. As implemented, because we release the lock before `Start()` has completed (and similarly for `Stop()`) the remaining code may end up overlapping which is not what we want.

Also note that we set `started` to true before the health monitor is actually ready and this is not reset in the envent of an error.

internal/rm/health.go
```diff
@@ -40,166 +45,372 @@ const (
...
+	}
+}
+
+// Start initializes NVML, registers event handlers, and starts the
+// monitoring goroutine. Blocks until initialization completes.
+func (r *nvmlHealthProvider) Start(ctx context.Context) error {
+	r.Lock()
+	if r.started {
+		r.Unlock()
+		return fmt.Errorf("health provider already started")
```
elezar: Why is this an error? Assuming we only set `started` once the HealthProvider / `HealthMonitor` has been successfully started, is there any harm in calling it again?

internal/rm/health.go
```diff
@@ -40,166 +45,372 @@ const (
...
 	}
+	r.started = true
+	r.Unlock()
+
+	// Get XID filter configuration
+	r.xidsDisabled = getDisabledHealthCheckXids()
```
elezar: This should be moved to construction.

internal/rm/health.go
```diff
@@ -40,166 +45,372 @@ const (
...
-			return nil
+		case <-r.ctx.Done():
+			return
 		default:
 		}
 
-		e, ret := eventSet.Wait(5000)
+		// Wait for NVML event (5 second timeout)
+		event, ret := r.eventSet.Wait(5000)
+
```
elezar:
```suggestion
```

internal/rm/health.go
ArangoGutierrez: Agree, now is a `return nil`

internal/rm/health.go
ArangoGutierrez: Done

internal/rm/health.go
ArangoGutierrez: moved

internal/rm/health.go
ArangoGutierrez: suggestion adopted

internal/rm/tegra_manager.go
ArangoGutierrez: done

internal/rm/health.go
elezar: I see that you have pulled this up to the `func (r *nvmlResourceManager) HealthProvider() HealthProvider {` implementation. This means that we have to construct the list of disabled XIDs twice. Why not handle this (and other static config) in the `NewNVMLHealthProvider` constructor instead?

internal/rm/health.go
elezar: I think this is more error prone than it needs to be. Note that we may want to distinguish between acceptable errors and non acceptable ones. As was the case for the related k8s-dra-driver PR, we may want to call `Init()` in the constructor and handle these errors (with deferred `Shutdown()` separately from errors that are triggered during `Start`.

For example, what about updating the constructor to:
```
// NewNVMLHealthProvider creates a new health provider for NVML devices.
// Does not start monitoring - caller must call Start().
func newNVMLHealthProvider(nvmllib nvml.Interface, config *spec.Config, devices Devices) (HealthProvider, error) {
	xids := getDisabledHealthCheckXids()
	if xids.IsAllDisabled() {
		return &noopHealthProvider{}, nil
	}

	ret := nvmllib.Init()
	if ret != nvml.SUCCESS {
		if *config.Flags.FailOnInitError {
			return nil, fmt.Errorf("failed to initialize NVML: %v", ret)
		}
		klog.Warningf("NVML init failed: %v; health checks disabled", ret)
		return &noopHealthProvider{}, nil
	}
	defer func() {
		ret := nvmllib.Shutdown()
		if ret != nvml.SUCCESS {
			klog.Infof("Error shutting down NVML: %v", ret)
		}
	}()

	klog.Infof("Ignoring the following XIDs for health checks: %v", xids)

	p := &nvmlHealthProvider{
		nvml:         nvmllib,
		config:       config,
		devices:      devices,
		unhealthy:    make(chan *Device, 64),
		xidsDisabled: xids,
	}
	return p, nil
}
```

Note that we return a `noopHealthProvider` if we are tolerant of `Init` errors at this point. However, we update `Start()` to look something like:

```
// Start initializes NVML, registers event handlers, and starts the
// monitoring goroutine. Blocks until initialization completes.
func (r *nvmlHealthProvider) Start(ctx context.Context) (rerr error) {
	r.Lock()
	defer r.Unlock()
	if r.started {
		// TODO: Is this an error condition? Could we just return?
		return fmt.Errorf("health provider already started")
	}
	r.Unlock()

	// Initialize NVML
	ret := r.nvml.Init()
	if ret != nvml.SUCCESS {
		return fmt.Errorf("failed to initialize NVML: %v", ret)
	}
	defer func() {
		if rerr != nil {
			_ = r.nvml.Shutdown()
		}
	}()

	// Create event set
	eventSet, ret := r.nvml.EventSetCreate()
	if ret != nvml.SUCCESS {
		return fmt.Errorf("failed to create event set: %v", ret)
	}
	defer func() {
		if rerr != nil {
			_ = eventSet.Free()
		}
	}()

	// Register devices
	if err := r.registerDevices(eventSet); err != nil {
		return fmt.Errorf("failed to register devices: %w", err)
	}

	klog.Infof("Health monitoring started for %d devices", len(r.devices))

	// Create child context
	r.ctx, r.cancel = context.WithCancel(ctx)

	// Start monitoring goroutine
	r.wg.Add(1)
	go r.runEventMonitor(eventSet)

	r.started = true

	return nil
}
```

Where we take actions immediately in the event of failure so that we don't have to rely on `cleanup()` being called eventually for these resources.

internal/rm/health.go
elezar: Please see my comment above.

internal/rm/health.go
```diff
@@ -40,166 +45,375 @@ const (
...
+		r.xidsDisabled)
 
-		gpu, ret := r.nvml.DeviceGetHandleByUUID(uuid)
-		if ret != nvml.SUCCESS {
-			klog.Infof("unable to get device handle from UUID: %v; marking it as unhealthy", ret)
-			unhealthy <- d
-			continue
-		}
+	// Create child context
+	r.ctx, r.cancel = context.WithCancel(ctx)
```
elezar: Could you comment on whether we need to add `WithCancel` to the context if it already has this done for the plugin?

internal/plugin/server.go
```diff
@@ -148,13 +144,14 @@ func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
...
-	go func() {
-		// TODO: add MPS health check
-		err := plugin.rm.CheckHealth(plugin.stop, plugin.health)
-		if err != nil {
-			klog.Errorf("Failed to start health check: %v; continuing with health checks disabled", err)
-		}
-	}()
+	// Initialize and start health provider
+	plugin.ctx, plugin.cancel = context.WithCancel(context.Background())
+	plugin.healthProvider = plugin.rm.HealthProvider()
```
elezar: This should be moved to the constructor.

internal/plugin/server.go
```diff
@@ -90,11 +92,9 @@ func (o *options) devicePluginForResource(ctx context.Context, resourceManager r
 		mps: mpsOptions,
 
 		socket: getPluginSocketPath(resourceManager.Resource()),
-		// These will be reinitialized every
-		// time the plugin server is restarted.
+		// server and healthProvider will be reinitialized every time
```
elezar: Why can't we instantiate the healthProvider here? Why is it required to reinitilize it on every start?

Would the following be valid?
```suggestion

		healthProvider: resourceManager.HealthProvider(),
		// server and healthProvider will be reinitialized every time
```

internal/plugin/server.go
```diff
@@ -148,13 +144,14 @@ func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
...
 
-	go func() {
-		// TODO: add MPS health check
-		err := plugin.rm.CheckHealth(plugin.stop, plugin.health)
-		if err != nil {
-			klog.Errorf("Failed to start health check: %v; continuing with health checks disabled", err)
-		}
-	}()
+	// Initialize and start health provider
+	plugin.ctx, plugin.cancel = context.WithCancel(context.Background())
```
elezar: At which point do we rather pass in a `ctx` and handle the `Cancel()` call externally?

internal/plugin/server.go
ArangoGutierrez: Added

internal/rm/health.go
ArangoGutierrez: Thanks for the detailed suggestion, adopted

internal/rm/health.go
ArangoGutierrez: moved to `NewNVMLHealthProvider` now

internal/rm/health.go
ArangoGutierrez: ack

internal/rm/health.go
ArangoGutierrez: we now use the plugin context

internal/plugin/server.go
ArangoGutierrez: we now use the plugin context

internal/plugin/server.go
ArangoGutierrez: moved

ArangoGutierrez:
Closing in favor of #1601

PR #1510, opened 2025-11-18
[GFD] Prevent GFD from crashing when device is unhealthy

ArangoGutierrez:
Modified Merge() to log and skip failed labelers instead of aborting the entire pipeline. When a device fails (XID error, GPU lost, etc.), GFD continues with healthy devices and generates partial labels.

### DIFF
--- a/internal/lm/list.go
+++ b/internal/lm/list.go
@@ -16,7 +16,9 @@
 
 package lm
 
-import "fmt"
+import (
+	"k8s.io/klog/v2"
+)
 
 // list represents a list of labelers that iself implements the Labeler interface.
 type list []Labeler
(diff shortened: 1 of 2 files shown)


ArangoGutierrez:
> How is what you're doing here different to just ignoring errors during labelling?

PTAL

internal/lm/list.go
```diff
@@ -29,13 +31,17 @@ func Merge(labelers ...Labeler) Labeler {
...
+// partial results. This provides fault tolerance for unhealthy devices or
+// transient errors without crashing the entire labeling process.
 func (labelers list) Labels() (Labels, error) {
 	allLabels := make(Labels)
-	for _, labeler := range labelers {
+	for i, labeler := range labelers {
 		labels, err := labeler.Labels()
 		if err != nil {
-			return nil, fmt.Errorf("error generating labels: %v", err)
+			klog.Warningf("Labeler at index %d failed, skipping: %v", i, err)
```
elezar, line 43: Just a nit: is the index useful in this case? Should we add a `Name()` or `String()` function to the labeler?

elezar, line 43: Also, does it make sense to use:
```suggestion
			klog.ErrorS(err, "skipping failed labeller", "index", i)
```
insread?

internal/lm/list_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// mockLabeler is a test double for the Labeler interface
+type mockLabeler struct {
```
elezar, line 27: Question: Why not use `moq` to generatre an fully-featured mock?

ArangoGutierrez:
Closing in favor of https://github.com/NVIDIA/k8s-device-plugin/pull/1561

PR #1529, opened 2025-11-20
fix: handle merge commits in backport bot with -m 1

karthikvetrivel:
Fix handling of merge commits, see here: https://github.com/NVIDIA/k8s-device-plugin/pull/1505

### DIFF
--- a/.github/scripts/backport.js
+++ b/.github/scripts/backport.js
@@ -70,7 +70,7 @@ for (const targetBranch of branches) {
       const commitMessage = commit.commit.message.split('\n')[0];
       core.info(`Cherry-picking commit ${i + 1}/${commits.length}: ${commitSha.substring(0, 7)} - ${commitMessage}`);
       try {
-        execSync(`git cherry-pick -x ${commitSha}`, { 
+        execSync(`git cherry-pick -m 1 -x ${commitSha}`, { 
           encoding: 'utf-8',
           stdio: 'pipe'
         });


.github/scripts/backport.js
```diff
@@ -69,12 +69,31 @@ for (const targetBranch of branches) {
       const commitSha = commit.sha;
       const commitMessage = commit.commit.message.split('\n')[0];
       core.info(`Cherry-picking commit ${i + 1}/${commits.length}: ${commitSha.substring(0, 7)} - ${commitMessage}`);
+
```
ArangoGutierrez: This differs from the GPU-operator implementation. Would you CP this change there as well?

karthikvetrivel: Yep, I would. I just opened this initial PR here because this is where the bug was.

.github/scripts/backport.js
```diff
@@ -69,12 +69,31 @@ for (const targetBranch of branches) {
...
+      let isMergeCommit = false;
+      try {
+        const parents = execSync(`git rev-list --parents -n 1 ${commitSha}`, { encoding: 'utf-8' }).trim().split(/\s+/);
+        isMergeCommit = parents.length > 2;
+      } catch (e) {
+        // If detection fails, proceed without -m and retry if git errors
+      }
+      
       try {
-        execSync(`git cherry-pick -x ${commitSha}`, {
```
elezar, line 73: I think it may be simpler to always add `-m 1` here instead of specific logic for merge commits? `-m 1` *should* be the default for regular commits anyway.

karthikvetrivel: Great idea. I wasn't sure if `-m 1` could be used for non-merge commits. Added.

ArangoGutierrez:
@karthikvetrivel  let's fix the commit history so we can move forward

karthikvetrivel:
> @karthikvetrivel let's fix the commit history so we can move forward

Fixed!

ArangoGutierrez (approved):

elezar (approved):
Thanks @karthikvetrivel.

PR #1537, opened 2025-11-25
api/config/v1: Abstract flag names to constants

aireet:
- Move hardcoded flag names from main.go files to consts.go
- Add FlagKubeletSocket and FlagCDIFeatureFlags constants
- Add Config Manager specific flag constants
- Update all components to use spec package constants
- Maintain backward compatibility

This improves code consistency and maintainability by centralizing flag name definitions in a single location.

### DIFF
--- a/api/config/v1/flags_test.go
+++ b/api/config/v1/flags_test.go
@@ -133,6 +133,22 @@ func TestUnmarshalFlags(t *testing.T) {
 				},
 			},
 		},
+		{
+			input: `{
+				"plugin": {
+					"kubeletSocket": "/var/lib/kubelet/device-plugins/kubelet.sock",
+					"cdiFeatureFlags": "nvidia.com/gpu"
+				}
+			}`,
+			output: Flags{
+				CommandLineFlags{
+					Plugin: &PluginCommandLineFlags{
+						KubeletSocket:   ptr("/var/lib/kubelet/device-plugins/kubelet.sock"),
+						CDIFeatureFlags: ptr("nvidia.com/gpu"),
+					},
+				},
+			},
+		},
 	}
 
 	for i, tc := range testCases {
@@ -214,6 +230,35 @@ func TestMarshalFlags(t *testing.T) {
 					"outputFile": null,
 					"sleepInterval": "5ns",
 					"machineTypeFile": null
+				}
+			}`,
+		},
+		{
+			input: Flags{
+				CommandLineFlags{
+					Plugin: &PluginCommandLineFlags{
+						KubeletSocket:   ptr("/var/lib/kubelet/device-plugins/kubelet.sock"),
+						CDIFeatureFlags: ptr("nvidia.com/gpu"),
+					},
+				},
+			},
+			output: `{
+				"migStrategy": null,
+				"failOnInitError": null,
+				"gdrcopyEnabled": null,
+				"gdsEnabled": null,
+				"mofedEnabled": null,
+				"useNodeFeatureAPI": null,
+				"deviceDiscoveryStrategy": null,
+				"plugin": {
+					"passDeviceSpecs": null,
+					"deviceListStrategy": null,
+					"deviceIDStrategy": null,
+					"cdiAnnotationPrefix": null,
+					"nvidiaCTKPath": null,
+					"containerDriverRoot": null,
+					"kubeletSocket": "/var/lib/kubelet/device-plugins/kubelet.sock",
+					"cdiFeatureFlags": "nvidia.com/gpu"
 				}
 			}`,
 		},


ArangoGutierrez:
hey @aireet thanks for this contribution

could you split thhis into multiple `[no-relnote] ....` commits? 

From your list 

- Move hardcoded flag names from main.go files to consts.go <--- this can be an individual commit
- Add FlagKubeletSocket and FlagCDIFeatureFlags constants <--- this can be an individual commit
- Add Config Manager specific flag constants <--- this can be an individual commit
- Update all components to use spec package constants <--- this can be an individual commit
- Maintain backward compatibility

aireet:
> hey @aireet thanks for this contribution
> 
> could you split thhis into multiple `[no-relnote] ....` commits?
> 
> From your list
> 
> * Move hardcoded flag names from main.go files to consts.go <--- this can be an individual commit
> * Add FlagKubeletSocket and FlagCDIFeatureFlags constants <--- this can be an individual commit
> * Add Config Manager specific flag constants <--- this can be an individual commit
> * Update all components to use spec package constants <--- this can be an individual commit
> * Maintain backward compatibility


hello @ArangoGutierrez 
I have done splited comments. The changes include:

1. Moved hardcoded flag names from main.go files to consts.go - Centralized all flag name constants in the spec package for better maintainability
2. Added Config Manager specific flag constants - Added proper constant definitions for config-manager specific flags
3. Updated all components to use spec package constants - Refactored all main.go files (nvidia-device-plugin, gpu-feature-discovery, mps-control-daemon, and config-manager) to use the new constants from the spec package


Best regards

elezar:
While I appreciate the intent behind this, using constants instead of actual strings in the CLI defintion makes the code more difficult to read / parse as a maintainer. I understand that we're duplicating the strings when updating these from the config file, but I would rather investigate addressing this as a follow-up to moving to urfave v3 -- where the altsrc package should provide aditional flexibility as required.

api/config/v1/consts.go
```diff
@@ -54,3 +54,67 @@ const (
 	DefaultNvidiaCTKPath       = "/usr/bin/nvidia-ctk"
 	DefaultContainerDriverRoot = "/driver-root"
 )
+
+// Command line flag names - Common flags
+const (
+	FlagMigStrategy             = "mig-strategy"
```
elezar, line 60: These are not actually part of the API in the strictest sense.

api/config/v1/consts.go
```diff
@@ -54,3 +54,67 @@ const (
...
+const (
+	FlagImexChannelIDs = "imex-channel-ids"
+	FlagImexRequired   = "imex-required"
+)
+
+// Command line flag names - Plugin additional flags
+const (
+	FlagKubeletSocket   = "kubelet-socket"
+	FlagCDIFeatureFlags = "cdi-feature-flags"
+)
```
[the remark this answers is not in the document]
elezar: These really should not be specified here since we don't read them from the config file.

api/config/v1/consts.go
```diff
@@ -54,3 +54,67 @@ const (
...
+
+// Command line flag names - Config manager specific flags
+const (
+	FlagKubeconfig         = "kubeconfig"
```
elezar, line 110: These are not specific to the config file format.

aireet:
> While I appreciate the intent behind this, using constants instead of actual strings in the CLI defintion makes the code more difficult to read / parse as a maintainer. I understand that we're duplicating the strings when updating these from the config file, but I would rather investigate addressing this as a follow-up to moving to urfave v3 -- where the altsrc package should provide aditional flexibility as required.


Thank you for the valuable feedback! After reconsidering, I completely agree with your perspective. Sometimes a bit of duplication is better than over-abstraction, especially when it comes at the cost of readability.

aireet:
> > While I appreciate the intent behind this, using constants instead of actual strings in the CLI defintion makes the code more difficult to read / parse as a maintainer. I understand that we're duplicating the strings when updating these from the config file, but I would rather investigate addressing this as a follow-up to moving to urfave v3 -- where the altsrc package should provide aditional flexibility as required.
> 
> Thank you for the valuable feedback! After reconsidering, I completely agree with your perspective. Sometimes a bit of duplication is better than over-abstraction, especially when it comes at the cost of readability.


Thank you again for the detailed feedback and guidance. I completely understand your perspective on maintaining readability vs. eliminating duplication.

Given your insights about waiting for urfave v3 and the altsrc package, I'll step back from this particular change for now. I appreciate you taking the time to explain the reasoning - it's been a valuable learning experience about the project's architecture decisions.

I'll look for other opportunities to contribute that align better with the current codebase patterns. Thanks for the constructive discussion!

Best regards

aireet:
Closing this PR based on the valuable feedback from @elezar. I appreciate the insights about maintaining readability vs. eliminating duplication, and agree that waiting for urfave v3 is a better approach.

PR #1538, opened 2025-11-26
WIP: Refactor health checks

elezar:
Before starting a more serious refactor as in #1508 it may be useful to:

1. Add some basic unit testing for health checking behaviour
2. Perform minor code reorganisation so as to move from what we currently have to what we want.

This is a demonstrator, and would still need to have additional work done to get this to the same point where #1508 is. The idea being that we get there more gradually, with fewer unrelated changes.

### DIFF
--- a/internal/rm/health.go
+++ b/internal/rm/health.go
@@ -40,8 +40,20 @@ const (
 	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 
+type nvmlHealthProvider struct {
+	nvmllib           nvml.Interface
+	devices           Devices
+	parentToDeviceMap map[string]*Device
+	deviceIDToGiMap   map[string]uint32
+	deviceIDToCiMap   map[string]uint32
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
Nice draft, LGTM, look forward for when it's reayd for review.

uristernik:
Combining healthchecks with startup probe could help with the deadlock I described in #1540 and fixed in #1541

ArangoGutierrez:
I put together https://github.com/NVIDIA/k8s-device-plugin/pull/1554 a combination of this PR and #1508 - I have it as a Draft for now while we discuss over it

rajatchopra:
@elezar The PR needs a rebase.

@ArangoGutierrez This has a slight overlap with your PR #1601

And the go-nvml/mock is imported in other PRs like this one too #1554

Evan, Carlos, please help me clean up this bunch, given we want to integrate device-api based checks soon.

PR #1541, opened 2025-11-30
Fix race condition in config-manager when label is unset

### LINKED ISSUE #1540: Init container hangs waiting for device-plugin.config label
We’ve encountered a recurring issue where the init container in the nvidia-device-plugin (and similarly in the mps-control-daemon) becomes stuck in the Running state and never completes. This prevents the main container from starting.

The last log line observed before the hang is (for the device-plugin):
```
nvidia-device-plugin-init W1127 22:37:28.837842       7 client_config.go:659] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
nvidia-device-plugin-init I1127 22:37:28.838748       7 main.go:246] Waiting for change to 'nvidia.com/device-plugin.config' label
```
The last log line observed before the hang is (for the mps-control-daemon):
```
mps-control-daemon-mounts I1127 18:28:06.816180       7 main.go:81] "NVIDIA MPS Control Daemon" version=<
mps-control-daemon-mounts     e0a461e1
mps-control-daemon-mounts     commit: e0a461e1e7ad1d239d4708c954f08c3038e2654a
mps-control-daemon-mounts  >
mps-control-daemon-mounts W1127 18:28:06.818457       7 mount_helper_common.go:34] Warning: mount cleanup skipped because path does not exist: /mps/shm
mps-control-daemon-init W1127 18:28:14.816030      20 client_config.go:659] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
mps-control-daemon-init I1127 18:28:14.816535      20 main.go:246] Waiting for change to 'nvidia.com/device-plugin.config' label
```

Restarting the affected pod immediately resolves the issue.

Observed Behavior
* The init container remains stuck indefinitely.
* The main container does not start.
* After restarting the pod, it proceeds normally.

Environment Details
* Component: nvidia-device-plugin (and mps-control-daemon)
* Kubernetes version: v1.32.9 EKS
* Driver / Toolkit version: v0.17.3
* Config override: Not overriding config per node. 
* Deployed via helm chart
* General config:
```
config:
  map:
    default: |-
      version: v1
      flags:
        migStrategy: none
      sharing:
        mps:
          renameByDefault: false
          resources:
          - name: nvidia.com/gpu
            replicas: 5
```

Frequency
* Occurs on approximately 1–3% of nodes.
* Appears to be non-deterministic, potentially a race condition.

Code Reference
Code reference for where the hang occurs:
https://github.com/NVIDIA/k8s-device-plugin/blob/624b771e02e9c2f14bfafe1d3433ebb4250aa9cb/cmd/config-manager/main.go#L229-L255
From initial review, the issue seems to occur while waiting to acquire the lock.
We considered setting ONESHOT=false, but since it appears to hang during lock acquisition, it may not resolve the issue.

uristernik:
## Summary
Fixes a race condition in the config-manager that causes it to hang indefinitely when the `nvidia.com/device-plugin.config` label is not set on the node.

## Problem
When the node label is not configured, there's a timing-dependent race condition:
1. If the Kubernetes informer's `AddFunc` fires before the first `Get()` call, it sets `current=""` and broadcasts
2. When `Get()` is subsequently called, it finds `lastRead == current` (both empty strings) and waits on the condition variable
3. No future events wake it up since the label remains unset, causing a permanent hang

This manifests as the init container hanging after printing:
```
Waiting for change to 'nvidia.com/device-plugin.config' label
```

## Solution
Added an `initialized` boolean flag to `SyncableConfig` to track whether `Get()` has been called at least once. The first `Get()` call now returns immediately with the current value, avoiding the deadlock. Subsequent `Get()` calls continue to wait properly when the value hasn't changed.

Fixes #1540

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -82,7 +82,7 @@ type SyncableConfig struct {
 	cond     *sync.Cond
 	mutex    sync.Mutex
 	current  string
-	lastRead string
+	lastRead *string
 }
 
 // NewSyncableConfig creates a new SyncableConfig
@@ -106,11 +106,12 @@ func (m *SyncableConfig) Set(value string) {
 func (m *SyncableConfig) Get() string {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
-	if m.lastRead == m.current {
+	if m.lastRead != nil && *m.lastRead == m.current {
 		m.cond.Wait()
 	}
-	m.lastRead = m.current
-	return m.lastRead
+	val := m.current
+	m.lastRead = &val
+	return *m.lastRead
 }
 
 func main() {


elezar:
@jgehrcke do we do something similar in the k8s-dra-driver? If so, how do we handle the initial synchronization there?

cmd/config-manager/main.go
elezar, line 101: In the `mig-manager` we have a conditional broadcast (https://github.com/NVIDIA/mig-parted/blob/a56cd122e899778996f7488012058056100d091f/cmd/nvidia-mig-manager/main.go#L101-L103). Would that address the race that you are trying to fix here, or does the `mig-manager` suffer from the same possible deadlock?

uristernik: I think it will still suffer the same possible deadlock. If I understand the scenario correctly, the deadlock happens when the "first" broadcast fires before the other go routine waits and listens for the broadcast. The condition here would skip the broadcast but in the described scenario no one listens to that broadcast anyways.

When it happens the only ways to release the waiting go routine is by:
1. labeling the node with some dummy config (and then deleting the label)
2. killing the pod

uristernik:
Hey @elezar, did you get a chance to have a look?

uristernik:
Gentle ping here @elezar this is happening quite a lot and requires manual intervention each time

elezar:
@uristernik we are reviewing. Thanks for your patience.

cmd/config-manager/main.go
```diff
@@ -79,10 +79,11 @@ type Flags struct {
...
 // *before* a call to Set() will be notified.
 type SyncableConfig struct {
-	cond     *sync.Cond
-	mutex    sync.Mutex
-	current  string
-	lastRead string
+	cond        *sync.Cond
+	mutex       sync.Mutex
+	current     string
+	lastRead    string
```
jojimt: I would suggest just changing lastRead to a pointer. This way, it effectively becomes a tertiary allowing distinction between uninitialized and ''

uristernik: Thanks I implemented your suggestion!

uristernik:
Quick update here @elezar @jgehrcke, I am running a forked version with this commit and we are not seeing the issue reproduce. We are running at any given time anywhere between 200-550 GPU nodes. 
CleanShot 2025-12-14 at 08 37 08@2x

uristernik:
@elezar any chance to get it merged? fix it some other way? 
We are running the forked version but we don't want to manage the fork forever 🙏

uristernik:
@elezar ping 🙏

uristernik:
@klueska @ArangoGutierrez @cdesiniotis @RenaudWasTaken can someone please have a look on this?

uristernik:
Can anyone please review this?

uristernik:
Thank you! @cdesiniotis done!

uristernik:
Thank you @cdesiniotis.
Following the link that @elezar sent https://github.com/NVIDIA/k8s-device-plugin/pull/1541#discussion_r2576920269 I think this fix is needed also in mig-manager, do you want me to open a pull request for that?
And can you please have a look at https://github.com/NVIDIA/k8s-device-plugin/pull/1481? Today I have to patch the helm chart manually to properly deploy MPS daemonset.

Please let me know if I can help

PR #1550, opened 2025-12-04
Enable gds, gdrcopy and mofed flags by default

rahulait:
## Description

With this change, we now always try to dynamically detect if the drivers are present or not.
Current code takes care of failure cases to fail-safe and we are leveraging that
for dynamic driver detection.

## Testing
- [x] Manual cluster testing (describe below)

Manually built an image with these changes and then deployed gpu-operator (once with CDI enabled and once with CDI disabled) with the image and gdrcopy driver enabled.

Verified gdrcopy specific spec is generated in /var/run/cdi when cdi and gdrcopy are enabled. Verified sample workload works fine.
Verified no spec generated when cdi is false and sample workload failing as driver is missing.

### DIFF
--- a/internal/plugin/server.go
+++ b/internal/plugin/server.go
@@ -336,6 +336,16 @@ func (plugin *nvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 		plugin.updateResponseForMPS(response)
 	}
 
+	if plugin.config.Flags.GDRCopyEnabled != nil && *plugin.config.Flags.GDRCopyEnabled {
+		response.Envs["NVIDIA_GDRCOPY"] = "enabled"
+	}
+	if plugin.config.Flags.GDSEnabled != nil && *plugin.config.Flags.GDSEnabled {
+		response.Envs["NVIDIA_GDS"] = "enabled"
+	}
+	if plugin.config.Flags.MOFEDEnabled != nil && *plugin.config.Flags.MOFEDEnabled {
+		response.Envs["NVIDIA_MOFED"] = "enabled"
+	}
+
 	// The following modifications are only made if at least one non-CDI device
 	// list strategy is selected.
 	if plugin.deviceListStrategies.AllCDIEnabled() {
@@ -351,15 +361,6 @@ func (plugin *nvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
 	}
 	if plugin.config.Flags.Plugin.PassDeviceSpecs != nil && *plugin.config.Flags.Plugin.PassDeviceSpecs {
 		response.Devices = append(response.Devices, plugin.apiDeviceSpecs(*plugin.config.Flags.NvidiaDevRoot, requestIds)...)
-	}
-	if plugin.config.Flags.GDRCopyEnabled != nil && *plugin.config.Flags.GDRCopyEnabled {
-		response.Envs["NVIDIA_GDRCOPY"] = "enabled"
-	}
-	if plugin.config.Flags.GDSEnabled != nil && *plugin.config.Flags.GDSEnabled {
-		response.Envs["NVIDIA_GDS"] = "enabled"
-	}
-	if plugin.config.Flags.MOFEDEnabled != nil && *plugin.config.Flags.MOFEDEnabled {
-		response.Envs["NVIDIA_MOFED"] = "enabled"
 	}
 	return response, nil
 }
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -103,16 +103,19 @@ func main() {
 		},
 		&cli.BoolFlag{
 			Name:    "gdrcopy-enabled",
+			Value:   true,
 			Usage:   "ensure that containers that request NVIDIA GPU resources are started with GDRCopy support",
 			EnvVars: []string{"GDRCOPY_ENABLED"},
 		},
 		&cli.BoolFlag{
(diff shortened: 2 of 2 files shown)


internal/utils/utils.go
```diff
@@ -0,0 +1,61 @@
...
+	GDRCOPY_DRIVER_MODULE = "gdrdrv"
+)
+
+// IsGdrdrvLoaded checks if the "gdrdrv" driver module is loaded.
+func IsGdrdrvLoaded() bool {
+	return isDriverLoaded(GDRCOPY_DRIVER_MODULE)
+}
+
+// isDriverLoaded checks if the specified driver module is loaded by reading the modules file.
+// It first checks "/host/proc/modules" (for containerized environments) and falls back to "/proc/modules" if not found.
```
ArangoGutierrez: Nit: let's keep a maximum column length for the docu-comments around 80.

internal/utils/utils.go
```diff
@@ -0,0 +1,61 @@
...
+func isDriverLoaded(module string) bool {
+	paths := []string{"/host/proc/modules", "/proc/modules"}
+	return slices.ContainsFunc(paths, func(path string) bool {
+		return isDriverLoadedWithPath(module, path)
+	})
+}
+
+// isDriverLoadedWithPath checks if the specified driver module is loaded by reading the specified modules file.
```
ArangoGutierrez: Nit: let's keep a maximum column length for the docu-comments around 80.

cmd/nvidia-device-plugin/plugin-manager.go
```diff
@@ -54,7 +55,7 @@ func GetPlugins(ctx context.Context, infolib info.Interface, nvmllib nvml.Interf
 		cdi.WithNvidiaCTKPath(*config.Flags.Plugin.NvidiaCTKPath),
 		cdi.WithDeviceIDStrategy(*config.Flags.Plugin.DeviceIDStrategy),
 		cdi.WithVendor("k8s.device-plugin.nvidia.com"),
-		cdi.WithGdrcopyEnabled(*config.Flags.GDRCopyEnabled),
+		cdi.WithGdrcopyEnabled(*config.Flags.GDRCopyEnabled || utils.IsGdrdrvLoaded()),
```
elezar: Question: Since these are effectively no-ops if the drivers are not available, is it a simpler change to change the default for these settings instead of adding logic to check whether the modules are loaded?

internal/plugin/server.go
```diff
@@ -352,13 +353,13 @@ func (plugin *nvidiaDevicePlugin) getAllocateResponse(requestIds []string) (*plu
...
 	}
-	if plugin.config.Flags.GDRCopyEnabled != nil && *plugin.config.Flags.GDRCopyEnabled {
+	if (plugin.config.Flags.GDRCopyEnabled != nil && *plugin.config.Flags.GDRCopyEnabled) || utils.IsGdrdrvLoaded() {
 		response.Envs["NVIDIA_GDRCOPY"] = "enabled"
 	}
-	if plugin.config.Flags.GDSEnabled != nil && *plugin.config.Flags.GDSEnabled {
+	if (plugin.config.Flags.GDSEnabled != nil && *plugin.config.Flags.GDSEnabled) || utils.IsGdsDriverLoaded() {
 		response.Envs["NVIDIA_GDS"] = "enabled"
 	}
-	if plugin.config.Flags.MOFEDEnabled != nil && *plugin.config.Flags.MOFEDEnabled {
```
rahulait, line 361: @elezar do you think we can simplify this as well for non-CDI case?

I tested non-cdi with the following change and didn't see any issue (ran a gdrcopy workload in non-cdi mode with below change in device-plugin and everything seems to work just fine).
```
-       if (plugin.config.Flags.GDRCopyEnabled != nil && *plugin.config.Flags.GDRCopyEnabled) || utils.IsGdrdrvLoaded() {
-               response.Envs["NVIDIA_GDRCOPY"] = "enabled"
-       }
-       if (plugin.config.Flags.GDSEnabled != nil && *plugin.config.Flags.GDSEnabled) || utils.IsGdsDriverLoaded() {
-               response.Envs["NVIDIA_GDS"] = "enabled"
-       }
-       if (plugin.config.Flags.MOFEDEnabled != nil && *plugin.config.Flags.MOFEDEnabled) || utils.IsMofedDriverLoaded() {
-               response.Envs["NVIDIA_MOFED"] = "enabled"
-       }
+       response.Envs["NVIDIA_GDRCOPY"] = "enabled"
+       response.Envs["NVIDIA_GDS"] = "enabled"
+       response.Envs["NVIDIA_MOFED"] = "enabled"
```
Looking for your feedback on whether this sounds right to you or not even for non-CDI case.

internal/utils/utils.go
rahulait: Updated. Ended up not requiring this file change.

internal/utils/utils.go
rahulait: Updated. Ended up not requiring this file change.

cmd/nvidia-device-plugin/plugin-manager.go
rahulait: Thanks, updated.

elezar (approved):
Thanks @rahulait. This looks good. Sorry for the delay.

I have also updated the PR title and description to represent the actual state.

elezar:
Removing the cherry-pick label. Let's readd it if we decide that we need this in the maintenance branch.

PR #1554, opened 2025-12-04
Enhance health check robustness and observability 

ArangoGutierrez:
## Enhance health check robustness and observability

This PR improves the device health check system with the following changes:

### Changes

1. **Extract `nvmlHealthProvider` struct** - Modularize health monitoring logic for better
   testability and separation of concerns

2. **Add buffered health channel** - Prevent health check goroutine from blocking when
   ListAndWatch is slow to consume events. Uses a 64-entry buffer with non-blocking send
   and fallback error logging

3. **Add device state tracking** - Track `LastUnhealthyTime` and `UnhealthyReason` on
   Device struct for better observability

4. **Add `withDevicePlacements` wrapper** - Enable unit testing of device placement logic
   independently of the full resource manager

5. **Add `TestCheckHealth` test** - Unit test for health check flow using `dgxa100` mock

6. **Add `healthCheckStats`** - Track events processed, devices marked unhealthy, errors,
   and XID distribution for operational visibility

### Commits

- `refactor: extract nvmlHealthProvider for modular health monitoring`
- `feat: add buffered health channel and device state tracking`
- `refactor: add withDevicePlacements wrapper and TestCheckHealth test`
- `fix: improve test type assertion and buffer size documentation`

### Testing

- All existing tests pass
- New `TestCheckHealth` test validates health check event processing
- Verified with `golangci-lint`, `go vet`, and `go build`

### Related

- Aligned with patterns from #1538


internal/rm/nvml_manager.go
```diff
@@ -95,6 +95,48 @@ func (r *nvmlResourceManager) CheckHealth(stop <-chan interface{}, unhealthy cha
 	return r.checkHealth(stop, r.devices, unhealthy)
 }
 
+// CheckDeviceHealth performs a simple health check on a single device by
+// verifying it can be accessed via NVML and responds to basic queries.
+// This is used for recovery detection - if a previously unhealthy device
+// passes this check, it's considered recovered. We intentionally keep this
+// simple and don't try to classify XIDs as recoverable vs permanent - that's
+// controlled via DP_DISABLE_HEALTHCHECKS / DP_ENABLE_HEALTHCHECKS env vars.
+func (r *nvmlResourceManager) CheckDeviceHealth(d *Device) (bool, error) {
```
elezar: I don't agree with this mechanism for transitioning the device back to healthy. This is an oversimplification and will lead to unhealthy devices being considered healthy. 

For example, if a device becomes unhealth due to repteated ECC memory errors, it is LIKELY that query functions such as the device name will continue to succeed and result in the device being marked as healthy when it needs a RESET.

Before we add this logic to the device plugin let us properly define and agree upon how we are detecting health.

Futhermore, although the XID-based health checking is something that is a means to an end, our ideal state is that some other component decides whether a device is health and the device plugin responds to these signals. Defining the unhealthy -> healthy transition here goes against this premise.

internal/plugin/server_test.go
```diff
@@ -254,3 +256,96 @@ func TestCDIAllocateResponse(t *testing.T) {
 func ptr[T any](x T) *T {
 	return &x
 }
+
+func TestTriggerDeviceListUpdate_Phase2(t *testing.T) {
```
elezar: As a matter of interest, what is `Phase2`? (Were these tests generated?)

internal/rm/health.go
```diff
@@ -123,6 +123,29 @@ func (s *healthCheckStats) report() {
 	}
 }
 
+// nvmlHealthProvider encapsulates the state and logic for NVML-based GPU
+// health monitoring. This struct groups related data and provides focused
+// methods for device registration and event monitoring.
+type nvmlHealthProvider struct {
```
elezar, line 165: Question: Why is the refactoring done AFTER the functional changes in this PR?

internal/rm/health.go
```diff
@@ -146,6 +146,38 @@ type nvmlHealthProvider struct {
 	stats *healthCheckStats
 }
 
+// registerDeviceEvents registers NVML event handlers for all devices in the
```
elezar, line 310: How is this actually different from the changes proposed in https://github.com/NVIDIA/k8s-device-plugin/commit/a6a9f186b106f42701b6c7391b853989b304ac24?

internal/rm/health.go
```diff
@@ -146,6 +146,126 @@ type nvmlHealthProvider struct {
...
+		case result, ok := <-eventChan:
+			if !ok {
+				// Event channel closed, exit
+				return nil
+			}
+
+			// Handle timeout - just continue
+			if result.ret == nvml.ERROR_TIMEOUT {
+				continue
+			}
```
elezar, line 241: Why do we even send the event in the case of a timeout?

internal/rm/health.go
```diff
@@ -146,6 +146,126 @@ type nvmlHealthProvider struct {
...
+
+			// Wait for NVML event with timeout
+			e, ret := eventSet.Wait(5000)
+
+			// Try to send event result, but respect context cancellation
+			select {
+			case <-ctx.Done():
+				return
+			case eventChan <- eventResult{event: e, ret: ret}:
+			}
```
elezar, line 221: This seems like the wrong way to try and ensure that the context has not been closed before sending to the event channel. What are we concerned about here? Is there a better way to ensure that this go routine terminates on the context being cancelled and doesn't block permenantly on the send?

internal/rm/health_test.go
elezar, line 1: The commit message mentions adding tests, but I only see code being removed here.

internal/rm/nvml_manager.go
ArangoGutierrez: I've removed the healthy transition logic from this PR.

The current implementation:
- Only transitions devices from healthy → unhealthy (never the reverse)
- Devices stay unhealthy until the pod/node is restarted
- Tracks `LastUnhealthyTime` and `UnhealthyReason` for observability only

I agree that:
1. Automatic healthy transitions based on query success is an oversimplification
2. The ideal state is an external component (like DCGM or node-health-checker) deciding health
3. This PR should focus on robust unhealthy detection, not recovery

The `MarkUnhealthy()` method added here is one-way by design. 
Any future healthy transitions should be driven by external signals as you suggest.

internal/plugin/server_test.go
ArangoGutierrez: Yes it _was_ generated, now removed

rajatchopra:
@ArangoGutierrez This PR needs a rebase. Is this going to be overlapping with the device-api prep PR?

rajatchopra:
@ArangoGutierrez closing this PR in favor of #1601 

Please re-open if I have gotten this wrong. Meanwhile, please help with merging #1601.

PR #1597, opened 2026-01-23
Add ownerReference to NodeFeature CRs for garbage collection

rajathagasthya:
Adds the Pod and Daemonset (controller reference) as owners, so that the NodeFeature CR is deleted when the GFD daemonset is deleted. This mirrors the behavior of how NFD adds ownerReferences to NodeFeature CRs it creates.


internal/lm/output.go
```diff
@@ -54,9 +55,17 @@ func NewOutputer(config *spec.Config, nodeConfig flags.NodeConfig, clientSets fl
 	if nodeConfig.Namespace == "" {
 		return nil, fmt.Errorf("required flag namespace not set")
 	}
+
+	ownerRefs, err := getOwnerReferences(context.TODO(), clientSets.Core, nodeConfig.Namespace, nodeConfig.PodName)
```
karthikvetrivel, line 59: `context.TODO()` vs `context.Background()` returns the same empty, non-nil context but `context.Background() `might be a better fit here since you delibrately want no parent context (i.e. you don't plan to eventually wire one in).

karthikvetrivel: (same applies to other usages of `context.TODO() `in diff)

internal/lm/output.go
```diff
@@ -119,37 +129,95 @@ func (n *nodeFeatureObject) Output(labels Labels) error {
...
-		} else {
-			klog.Infof("no changes in NodeFeature object, not updating")
-		}
+	nfrUpdated := nfr.DeepCopy()
+	nfrUpdated.Labels = map[string]string{nfdv1alpha1.NodeFeatureObjNodeNameLabel: nodename}
+	nfrUpdated.Spec = nfdv1alpha1.NodeFeatureSpec{Features: *nfdv1alpha1.NewFeatures(), Labels: labels}
+	nfrUpdated.OwnerReferences = n.ownerRefs
+
+	if apiequality.Semantic.DeepEqual(nfr, nfrUpdated) {
+		klog.V(2).Infof("no changes in NodeFeature object %s", nodeFeatureName)
```
karthikvetrivel: Just curious--why set this at a higher verbosity? Why 2?

internal/lm/output.go
```diff
@@ -119,37 +129,95 @@ func (n *nodeFeatureObject) Output(labels Labels) error {
...
+func getOwnerReferences(ctx context.Context, client kubernetes.Interface, namespace, podName string) ([]metav1.OwnerReference, error) {
+	if podName == "" {
+		klog.Info("Pod name not provided, skipping owner reference resolution")
+		return nil, nil
+	}
+
+	if client == nil {
+		klog.Info("Kubernetes client not available, skipping owner reference resolution")
+		return nil, nil
+	}
```
karthikvetrivel: nit: This check seems unnecessary. By the time we reach `getOwnerReferences`, the invariant `client != nil `should already be established — client initialization happens at startup, and if it fails, the application exits before reaching this code path. A `nil` client here would indicate a bug in the caller. What do you think?

internal/lm/output.go
rajathagasthya: `context.Background()` is typically used as top-level context in a main function, while `context.TODO()` indicates the surrounding function doesn't accept a context (yet). So I think `context.TODO()` is more appropriate here and what was in the code before as well.

internal/lm/output.go
rajathagasthya: Good catch. I'll remove this.

internal/lm/output.go
rajathagasthya: Makes sense. This check is not strictly necessary.

internal/lm/output.go
karthikvetrivel: Yes, that's fair. When I read the rest of the file, I also found it used TODO. Will resolve.

deployments/helm/nvidia-device-plugin/templates/daemonset-gfd.yml
```diff
@@ -145,6 +145,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: POD_NAME
```
cdesiniotis, line 148: Just a note -- we will need to make a similar change in the gpu-operator repo.

rajathagasthya: Thanks. Changes are here: https://github.com/NVIDIA/gpu-operator/pull/2071

PR #1598, opened 2026-01-23
Bump github.com/NVIDIA/nvidia-container-toolkit to v1.18.2

elezar:
The most notable change to include is https://github.com/NVIDIA/nvidia-container-toolkit/pull/1594 which fixes an inconsistency in generated CDI specs where IPC sockets are mounted readonly causing failures in certain nested container use cases (e.g. Slurm on K8s).


elezar:
Replaced by #1599

PR #1601, opened 2026-01-26
Refactor health checks

ArangoGutierrez:


internal/plugin/server.go
```diff
@@ -109,14 +114,18 @@ func getPluginSocketPath(resource spec.ResourceName) string {
...
+	if plugin.healthCancel != nil {
+		plugin.healthCancel()
+	}
+	plugin.healthWg.Wait()
 	plugin.server = nil
 	plugin.health = nil
-	plugin.stop = nil
+	plugin.healthCtx = nil
+	plugin.healthCancel = nil
 }
```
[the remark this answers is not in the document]
ArangoGutierrez: addressed

internal/plugin/server.go
```diff
@@ -265,15 +276,19 @@ func (plugin *nvidiaDevicePlugin) GetDevicePluginOptions(context.Context, *plugi
 
 // ListAndWatch lists devices and update that list according to the health status
 func (plugin *nvidiaDevicePlugin) ListAndWatch(e *pluginapi.Empty, s pluginapi.DevicePlugin_ListAndWatchServer) error {
+	// Capture references at start to avoid race with cleanup() which may nil these fields.
+	healthCtx := plugin.healthCtx
+	health := plugin.health
```
elezar: Do we need to add synchronisation to ensure that these are actually initialised by the time that `ListAndWatch` is called, or is it guaranteed that this will be called only after the context has been initialized. Looking at the cleanup code, it is not clear why we would set `plugin.healthCtx` to `nil` there.

Note that if we were to pass the context in to the server construction, then we would not have to do this since the lifetime would be controlled at the caller and we would not set the context to nil in plugin cleanup.

internal/plugin/server.go
```diff
@@ -109,14 +114,18 @@ func getPluginSocketPath(resource spec.ResourceName) string {
...
-	close(plugin.stop)
+	if plugin.healthCancel != nil {
+		plugin.healthCancel()
+	}
+	plugin.healthWg.Wait()
 	plugin.server = nil
 	plugin.health = nil
-	plugin.stop = nil
+	plugin.healthCtx = nil
+	plugin.healthCancel = nil
```
elezar: Why are we setting these to `nil` now?

internal/plugin/server.go
```diff
@@ -148,10 +157,12 @@ func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
...
+	plugin.healthWg.Add(1)
 	go func() {
+		defer plugin.healthWg.Done()
 		// TODO: add MPS health check
-		err := plugin.rm.CheckHealth(plugin.stop, plugin.health)
-		if err != nil {
+		err := plugin.rm.CheckHealth(plugin.healthCtx, plugin.health)
+		if err != nil && !errors.Is(err, context.Canceled) {
 			klog.Errorf("Failed to start health check: %v; continuing with health checks disabled", err)
 		}
```
elezar: What about logging the happy path case? (Also in chich cases will we return `contex.Cancelled`?
```suggestion
		switch {
		case err == nil:
				fallthrough
		case errors.Is(err, context.Cancelled):
				klog.Infof("Stopping health check")
		case err != nil:
				klog.Errorf("Failed to start health check: %v; continuing with health checks disabled", err)
		}
}
```

elezar (commented):
Thanks @ArangoGutierrez. I have a minor question regarding the cleaning up of the context.

internal/plugin/server.go
ArangoGutierrez: Addressed

internal/plugin/server.go
ArangoGutierrez: removed.
Fields preserved (not niled) for restart support

internal/plugin/server.go
ArangoGutierrez: Addressed, now `healthCtx` and `healthCancel` now created in `devicePluginForResource()`

internal/plugin/server.go
```diff
@@ -62,7 +63,12 @@ type nvidiaDevicePlugin struct {
 	socket string
 	server *grpc.Server
 	health chan *rm.Device
-	stop   chan interface{}
+
+	// healthCtx and healthCancel control the health check goroutine lifecycle.
+	// healthWg is used to wait for the health check goroutine to complete during cleanup.
+	healthCtx    context.Context
+	healthCancel context.CancelFunc
+	healthWg     sync.WaitGroup
```
elezar:
```suggestion
	// healthCtx and healthCancel control the health check goroutine lifecycle.
	healthCtx    context.Context
	healthCancel context.CancelFunc
	// healthWg is used to wait for the health check goroutine to complete during cleanup.
	healthWg     sync.WaitGroup
```

internal/plugin/server.go
elezar, line 55: Question: Why do we need a second context?

internal/plugin/server.go
```diff
@@ -109,14 +122,27 @@ func getPluginSocketPath(resource spec.ResourceName) string {
...
+	// healthCtx and healthCancel already initialized at construction time
 }
 
 func (plugin *nvidiaDevicePlugin) cleanup() {
-	close(plugin.stop)
+	if plugin.healthCancel != nil {
+		plugin.healthCancel()
+		// Recreate context for potential plugin restart. The same plugin instance
+		// may be restarted via Start() after Stop(), so we need a fresh context.
+		plugin.healthCtx, plugin.healthCancel = context.WithCancel(plugin.ctx)
```
elezar: When are we ever setting these to nil now? This should ONLY be done in the constructor.

internal/plugin/server.go
ArangoGutierrez: So we can cancel the healthchecks independently.

internal/plugin/server.go
ArangoGutierrez: Thanks will implement

internal/plugin/server.go
ArangoGutierrez: Addressed by [da507ef](https://github.com/NVIDIA/k8s-device-plugin/pull/1601/commits/da507ef1c71219fc8c1a2cd4a1c3d4a9625ae11a)

internal/rm/health.go
```diff
@@ -71,13 +83,14 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
...
+	// TODO: This should be factored out. The main issue is marking the devices
+	// unhealthy as part of this loop.
 	parentToDeviceMap := make(map[string]*Device)
 	deviceIDToGiMap := make(map[string]uint32)
 	deviceIDToCiMap := make(map[string]uint32)
-
-	eventMask := uint64(nvml.EventTypeXidCriticalError | nvml.EventTypeDoubleBitEccError | nvml.EventTypeSingleBitEccError)
 	for _, d := range devices {
-		uuid, gi, ci, err := r.getDevicePlacement(d)
+		uuid, gi, ci, err := (&withDevicePlacements{r.nvml}).getDevicePlacement(d)
```
cdesiniotis, line 118: Could you provide some context as to why we landed on this change / refactor?

internal/rm/health.go
```diff
@@ -107,14 +137,16 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 			klog.Warningf("Device %v is too old to support healthchecking.", d.ID)
 		case ret != nvml.SUCCESS:
 			klog.Infof("Marking device %v as unhealthy: %v", d.ID, ret)
```
cdesiniotis: nit: could we update the log message here to indicate why we are marking the device unhealthy? (we failed during NVML event registration)

AGENTS.md
cdesiniotis, line 1: Could you explain how you are using this file / how it was generated? If you used a coding agent, could you acknowledge that in the PR description and describe how you used it? 

This file needs to be removed before we can merge this PR.

internal/rm/health.go
guptaNswati: +1.  on dra, its `unable to register events for %s: %v; marking it as unhealthy", d.ID, ret`

internal/plugin/server.go
```diff
@@ -164,12 +164,18 @@ func (plugin *nvidiaDevicePlugin) Stop() error {
 	if plugin == nil || plugin.server == nil {
 		return nil
 	}
+	// We stop the health checks
+	plugin.healthCancel()
```
guptaNswati: nit: add a nil check
`if plugin.healthCancel != nil { plugin.healthCancel() }`

internal/plugin/server.go
```diff
@@ -148,10 +139,19 @@ func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
 	}
 	klog.Infof("Registered device plugin for '%s' with Kubelet", plugin.rm.Resource())
 
+	plugin.health = make(chan *rm.Device)
```
guptaNswati: should these be initialized before registering the plugin?

internal/plugin/server.go
```diff
@@ -164,12 +164,18 @@ func (plugin *nvidiaDevicePlugin) Stop() error {
 	if plugin == nil || plugin.server == nil {
 		return nil
 	}
+	// We stop the health checks
+	plugin.healthCancel()
+	plugin.healthWg.Wait()
+	// Close the health channel after waiting for the health check goroutine to terminate.
+	close(plugin.health)
```
guptaNswati: nil check here also. 
`if plugin.health != nil { close(plugin.health); plugin.health = nil }`

internal/plugin/server.go
```diff
@@ -271,9 +278,13 @@ func (plugin *nvidiaDevicePlugin) ListAndWatch(e *pluginapi.Empty, s pluginapi.D
 
 	for {
 		select {
-		case <-plugin.stop:
+		case <-plugin.healthCtx.Done():
```
guptaNswati, line 281: nit: add a log message
`klog.V(4).Info("Stopping health checks..")`

internal/plugin/server.go
```diff
@@ -271,9 +278,13 @@ func (plugin *nvidiaDevicePlugin) ListAndWatch(e *pluginapi.Empty, s pluginapi.D
...
 			return nil
-		case d := <-plugin.health:
+		case d, ok := <-plugin.health:
+			if !ok {
+				// Health channel closed, health checks stopped
```
guptaNswati: i would log here also. helps in debugging

internal/rm/health.go
```diff
@@ -71,13 +83,14 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		_ = eventSet.Free()
 	}()
 
+	// Construct the device maps.
```
guptaNswati, line 111: oh you already called it out. Yes, we should refractor it into helpers

internal/rm/health.go
```diff
@@ -40,8 +41,19 @@ const (
 	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 
+type nvmlHealthProvider struct {
```
guptaNswati: why not `nvmlDeviceHealthProvider`

internal/rm/health.go
```diff
@@ -276,25 +310,29 @@ func newHealthCheckXIDs(xids ...string) disabledXIDs {
...
+func (r *withDevicePlacements) getDevicePlacement(d *Device) (string, uint32, uint32, error) {
 	if !d.IsMigDevice() {
 		return d.GetUUID(), 0xFFFFFFFF, 0xFFFFFFFF, nil
 	}
 	return r.getMigDeviceParts(d)
 }
 
 // getMigDeviceParts returns the parent GI and CI ids of the MIG device.
-func (r *nvmlResourceManager) getMigDeviceParts(d *Device) (string, uint32, uint32, error) {
+func (r *withDevicePlacements) getMigDeviceParts(d *Device) (string, uint32, uint32, error) {
```
guptaNswati: nit: can we call it `dp` or `d` instead of `r`

internal/rm/health.go
```diff
@@ -40,8 +41,19 @@ const (
 	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 
+type nvmlHealthProvider struct {
+	nvmllib           nvml.Interface
+	devices           Devices
+	parentToDeviceMap map[string]*Device
```
guptaNswati, line 50: can we make it a type similar to `type devicePlacementMap map[string]map[uint32]map[uint32]*Device`
https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/main/cmd/gpu-kubelet-plugin/device_health.go#L36

internal/rm/health.go
guptaNswati: IMO, this is a good refactor as it separates NVML from resourceManager.
```
type nvmlResourceManager struct {
	resourceManager
	nvml nvml.Interface
}
```
This will also allow further refractor as called out here https://github.com/NVIDIA/k8s-device-plugin/pull/1601/changes#r2795621700

internal/rm/health.go
```diff
@@ -86,18 +99,35 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
...
+		deviceIDToCiMap:   deviceIDToCiMap,
+		xidsDisabled:      xids,
+	}
+	p.registerDeviceEvents(eventSet)
+
+	return p.runEventMonitor(ctx, eventSet)
+}
 
-		gpu, ret := r.nvml.DeviceGetHandleByUUID(uuid)
+func (r *nvmlHealthProvider) registerDeviceEvents(eventSet nvml.EventSet) {
```
guptaNswati: nit: r -> h or hp

internal/rm/health.go
```diff
@@ -86,18 +99,35 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		deviceIDToGiMap[d.ID] = gi
 		deviceIDToCiMap[d.ID] = ci
 		parentToDeviceMap[uuid] = d
+	}
+
+	p := nvmlHealthProvider{
+		nvmllib:           r.nvml,
```
guptaNswati, line 134: is this needed for tests?

guptaNswati:
sorry for the delayed review. @ArangoGutierrez thank you for your patience. 

I looked at it from the perspective of what we did for DRA health checks as the gaol is to align with DRA and integrate with  the device-api.

I have minor nits but overall the refractor LGTM. 

I still want to spend some more time on reading and trying the mocks tests.

ArangoGutierrez:
I have addressed all your comments @guptaNswati  / @cdesiniotis  PTAL

internal/rm/health.go
ArangoGutierrez: The goal of this refactor is to take the DevicePlugin to a similar state as our DRA after @guptaNswati Health check work. 
This in order to prepare both DRA and DevicePlugin for NVsentinel integration via the Device-API

AGENTS.md
ArangoGutierrez: DONE

internal/rm/health.go
guptaNswati: Lets add a detailed comment on how it will help. I also had to look into it. Because rn when we are only using NVML, its not making sense to decouple DevicePlacement logic from NVMLResourceManager. We need to note how this placement code can be reused with multiple providers and not just NVML.
```
type nvmlResourceManager struct {
	resourceManager
	nvml nvml.Interface
```

to having a new wrapper
```
type withDevicePlacements struct {
	nvml.Interface
}
```

internal/rm/health.go
guptaNswati: @ArangoGutierrez why do we need to have the lib here? nvmlResourceManager.nvml is used everywhere

guptaNswati:
Thank you. I dint get a chance to test it. Can you please paste the test logs

ArangoGutierrez:
> Thank you. I dint get a chance to test it. Can you please paste the test logs

Unfortunately, not, didn't save the logs, and the cluster I used is down

guptaNswati:
> > Thank you. I dint get a chance to test it. Can you please paste the test logs
> 
> Unfortunately, not, didn't save the logs, and the cluster I used is down

i know these are straightforward changes and you must have tested them very well. I feel more comfortable when i see the test logs so i will try to set up a cluster next week and see if we can test them.

ArangoGutierrez:
@guptaNswati 

```bash
I0216 18:53:08.797944  1 main.go:250] "Starting NVIDIA Device Plugin" version=v0.19.0-dev commit: d6c787ca9
I0216 18:53:08.797993  1 main.go:253] Starting FS watcher for /var/lib/kubelet/device-plugins
I0216 18:53:08.798043  1 main.go:260] Starting OS watcher.
I0216 18:53:08.798413  1 main.go:275] Starting Plugins.
I0216 18:53:08.798440  1 main.go:332] Loading configuration.
I0216 18:53:08.837087  1 server.go:203] Starting GRPC server for 'nvidia.com/gpu'
I0216 18:53:08.837916  1 server.go:130] Starting to serve 'nvidia.com/gpu' on /var/lib/kubelet/device-plugins/nvidia-gpu.sock
I0216 18:53:08.840309  1 server.go:137] Registered device plugin for 'nvidia.com/gpu' with Kubelet
I0216 18:53:08.843525  1 health.go:76] Ignoring the following XIDs for health checks: map[13:true 31:true 43:true 45:true 68:true 109:true]
```

guptaNswati:
> @guptaNswati
> 
> ```shell
> I0216 18:53:08.797944  1 main.go:250] "Starting NVIDIA Device Plugin" version=v0.19.0-dev commit: d6c787ca9
> I0216 18:53:08.797993  1 main.go:253] Starting FS watcher for /var/lib/kubelet/device-plugins
> I0216 18:53:08.798043  1 main.go:260] Starting OS watcher.
> I0216 18:53:08.798413  1 main.go:275] Starting Plugins.
> I0216 18:53:08.798440  1 main.go:332] Loading configuration.
> I0216 18:53:08.837087  1 server.go:203] Starting GRPC server for 'nvidia.com/gpu'
> I0216 18:53:08.837916  1 server.go:130] Starting to serve 'nvidia.com/gpu' on /var/lib/kubelet/device-plugins/nvidia-gpu.sock
> I0216 18:53:08.840309  1 server.go:137] Registered device plugin for 'nvidia.com/gpu' with Kubelet
> I0216 18:53:08.843525  1 health.go:76] Ignoring the following XIDs for health checks: map[13:true 31:true 43:true 45:true 68:true 109:true]
> ```

nice. thank you. so for device-plugin we always log which xids are ignored. We dint do that for DRA. Also can please share your mock test log triggering a xid error and the action taken. 

Is this a valid concern https://github.com/NVIDIA/k8s-device-plugin/pull/1601#discussion_r2800709523

i have a few other minor comments that i would like answered. 

we are very close to merging this. Thank you @ArangoGutierrez

ArangoGutierrez:
@guptaNswati I have addressed all your comments
cc @rajatchopra

guptaNswati:
@ArangoGutierrez this does seem like a legit bug https://github.com/NVIDIA/k8s-device-plugin/pull/1601#discussion_r2800709523. parentToDeviceMap overwrites when multiple MIG devices share a parent GPU. I see its from the old code, but now that we are refactoring, lets fix it. This is what i did for DRA `type devicePlacementMap map[string]map[uint32]map[uint32]*Device`
https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/main/cmd/gpu-kubelet-plugin/device_health.go#L36

is it possible to add test for mig devices?

internal/plugin/server.go
```diff
@@ -125,6 +125,7 @@ func (plugin *nvidiaDevicePlugin) Start(kubeletSocket string) error {
 	err := plugin.Serve()
 	if err != nil {
 		klog.Errorf("Could not start device plugin for '%s': %s", plugin.rm.Resource(), err)
+		plugin.healthCancel()
 		return err
 	}
```
guptaNswati, line 135: @ArangoGutierrez (nit) though`healthCancel()`properly cancels things but should these still be set to nil 
```
plugin.health = nil
plugin.server = nil
```

internal/plugin/server.go
ArangoGutierrez: The `close(plugin.health)` call was removed in a subsequent commit — context cancellation via `healthCtx` is now the sole shutdown mechanism. Since `Stop()` no longer touches `plugin.health`, the nil check is no longer needed. `ListAndWatch` exits via `<-plugin.healthCtx.Done()`.

internal/plugin/server.go
ArangoGutierrez: Yes — they're initialized at lines 122-123, which is before `Serve()` (line 125) and `Register()` (line 133). This is intentional: kubelet can call `ListAndWatch` as soon as the gRPC server is up, and `ListAndWatch` reads from both `plugin.health` and `plugin.healthCtx`. The health goroutine itself starts after registration (line 140), which is the right order since we want to ensure the server is fully registered before starting event monitoring.

internal/rm/health.go
ArangoGutierrez: Went with `nvmlDeviceHealthChecker` since the type actively checks health (monitors NVML events, marks unhealthy), rather than providing health data. This also aligns with the method names (`checkHealth`, `registerDeviceEvents`, `runEventMonitor`). Happy to change if you prefer `nvmlDeviceHealthProvider` though.

internal/rm/health.go
ArangoGutierrez: Good idea — the DRA pattern with `type devicePlacementMap map[string]map[uint32]map[uint32]*Device` is cleaner. That's a bigger refactor that changes how device lookup works in `runEventMonitor`. I'll create a follow-up PR to align DP with DRA's data model here.

internal/rm/health.go
ArangoGutierrez: Added a comment on the struct field explaining this. The NVML interface is stored on the health checker (rather than accessing it via `nvmlResourceManager`) to keep it decoupled and independently testable. The health checker methods (`registerDeviceEvents`, `runEventMonitor`) need `DeviceGetHandleByUUID` during event registration and processing.

internal/rm/health.go
```diff
@@ -86,18 +99,35 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		deviceIDToGiMap[d.ID] = gi
 		deviceIDToCiMap[d.ID] = ci
 		parentToDeviceMap[uuid] = d
```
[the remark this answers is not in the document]
ArangoGutierrez: Good catch — this is a pre-existing issue. When multiple MIG devices share the same parent GPU, only the last one gets stored. I'll track this as a separate fix (needs its own tests for the multi-MIG case) in a follow-up PR.

internal/rm/health.go
```diff
@@ -124,18 +156,20 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		}
 		if ret != nvml.SUCCESS {
 			klog.Infof("Error waiting for event: %v; Marking all devices as unhealthy", ret)
-			for _, d := range devices {
-				unhealthy <- d
+			for _, d := range h.devices {
+				h.unhealthy <- d
 			}
 			continue
```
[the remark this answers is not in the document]
ArangoGutierrez: This is pre-existing behavior — the busy loop on persistent NVML errors existed before this refactor. The new code does improve the situation: `markUnhealthy` is now context-aware, so the loop can be interrupted by cancellation even during sends. Adding backoff or error-count limits would be a good follow-up improvement.

internal/plugin/server.go
ArangoGutierrez: Good point. Now the error path also calls `plugin.server.Stop()` and nils both `plugin.server` and `plugin.health` to leave the struct in a clean state for potential retry. See latest commit.

ArangoGutierrez:
@guptaNswati comments addressed, PTAL

internal/rm/health.go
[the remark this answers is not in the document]
guptaNswati: yes. sounds good.

guptaNswati:
> @guptaNswati comments addressed, PTAL

Thank you @ArangoGutierrez 

LGTM. Approved.

Pls. open issue for follow-up, especially this on https://github.com/NVIDIA/k8s-device-plugin/pull/1601#issuecomment-4092748484

guptaNswati:
I dont seem to have approval rights. 

cc @tariq1890 @shivamerla @cdesiniotis can i get access rightd?

rajatchopra (approved):
/lgtm

@ArangoGutierrez needs rebase

guptaNswati (approved):
Pls rebase @ArangoGutierrez its g2g from @rajatchopra too.

ArangoGutierrez:
@cdesiniotis  / @tariq1890  Ready for last review and Merge

PR #1602, opened 2026-01-27
[improvement] : add github issue templates

rahulait:
This PR adds templates for common github issues (issues, features, questions).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,15 +1,14 @@
 ---
-name: Device Plugin Bug report
+name: Bug report
 about: Create a report to help us improve
-title: ''
-labels: ''
+title: '[Bug]: '
+labels: ["bug", "needs-triage"]
 
 ---
 
-_The template below is mostly useful for bug reports and support questions. Feel free to remove anything which doesn't apply to you and add more information where it makes sense._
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -35,4 +34,3 @@ Additional information that might help better understand your environment and re
  - [ ] Any relevant kernel output lines from `dmesg`
  - [ ] NVIDIA packages version from `dpkg -l '*nvidia*'` _or_ `rpm -qa '*nvidia*'`
  - [ ] NVIDIA container library version from `nvidia-container-cli -V`
```
elezar: Possibly out of scope of this PR, but we should revisit what actually makes sense when reporting a bug here. Let's rather have less if we're not sure that something is useful.

elezar: For example, we should prefer the `nvidia-ctk` version here instead.

rahulait: Sure, updated. If you have a list which we want to add here, let me know and I'll add that. Else, we can also revisit this in future.

rahulait:
@elezar Let me know if this looks good and we can merge this PR.

cdesiniotis (approved):

PR #1603, opened 2026-01-27
Add `--sleep-interval=infinite` support to GFD for running as a pod

rajathagasthya:
Reuse the existing --sleep-interval flag with support for 'infinite' as a special value.  This causes GFD to label once and sleep indefinitely, which is useful for running as a Kubernetes pod that should not exit.

### DIFF
--- a/api/config/v1/duration.go
+++ b/api/config/v1/duration.go
@@ -19,15 +19,29 @@ package v1
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"time"
 )
 
 // Duration wraps a time.Duration function with custom JSON marshaling/unmarshaling
 type Duration time.Duration
 
+// IsInfinite returns true if the duration represents an infinite sleep interval.
+func (d *Duration) IsInfinite() bool {
+	return d != nil && time.Duration(*d) == math.MaxInt64
+}
+
+// String returns a human-readable representation of the duration.
+func (d Duration) String() string {
+	if d.IsInfinite() {
+		return "infinite"
+	}
+	return time.Duration(d).String()
+}
+
 // MarshalJSON marshals 'Duration' to its raw bytes representation
 func (d Duration) MarshalJSON() ([]byte, error) {
-	return json.Marshal(time.Duration(d).String())
+	return json.Marshal(d.String())
 }
 
 // UnmarshalJSON unmarshals raw bytes into a 'Duration' type.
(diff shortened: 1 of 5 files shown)


cmd/gpu-feature-discovery/main.go
```diff
@@ -130,6 +136,15 @@ func main() {
 }
 
 func validateFlags(config *spec.Config) error {
+	// Check mutual exclusivity of oneshot flags
```
tariq1890: this comment is redundant. We can infer this from the logic and error message

cmd/gpu-feature-discovery/main.go
```diff
@@ -130,6 +136,15 @@ func main() {
...
+	if config.Flags.GFD != nil {
+		oneshot := config.Flags.GFD.Oneshot != nil && *config.Flags.GFD.Oneshot
+		oneshotDaemon := config.Flags.GFD.OneshotDaemon != nil && *config.Flags.GFD.OneshotDaemon
+		if oneshot && oneshotDaemon {
+			return fmt.Errorf("--oneshot and --oneshot-daemon cannot be used together")
```
tariq1890:
```suggestion
			return fmt.Errorf("only one of --oneshot and --oneshot-daemon may be used")
```

cdesiniotis (approved):

elezar (changes requested):
instead of adding a new flag, does it not make sense to reuse the sleep interval? We could add support for `infinite` as a special string that disables the main loop.

See: 
https://github.com/NVIDIA/k8s-device-plugin/blob/1e44e5705b91b8cf63681e71a8403015954f3a9c/api/config/v1/duration.go#L43-L50

Alternatively, a user COULD aready get close to this behaviour by setting this duration to something really long (years?).

api/config/v1/duration.go
```diff
@@ -19,12 +19,26 @@ package v1
...
+	"math"
 	"time"
 )
 
 // Duration wraps a time.Duration function with custom JSON marshaling/unmarshaling
 type Duration time.Duration
 
+// IsInfinite returns true if the duration represents an infinite sleep interval.
+func (d *Duration) IsInfinite() bool {
+	return d != nil && time.Duration(*d) == math.MaxInt64
```
elezar, line 31: Are we concerned with overvflow if we set this to `MaxInt64`? We could also reaslistically use `0` or `-1` as values here.

api/config/v1/duration.go
```diff
@@ -19,12 +19,26 @@ package v1
...
+}
+
+// String returns a human-readable representation of the duration.
+func (d Duration) String() string {
+	if d.IsInfinite() {
+		return "infinite"
```
elezar, line 37: Does this mean that this outputs as `infinite` when we log the config applied? Would we need to have a custom marshaller to do that, or does that just call the `String()` function? (Update: the `MarshalJSON` funciton needs to be updated to call `d.String()` directly).

elezar: OK. It seems not:
```
./gpu-feature-discovery --oneshot --sleep-interval=infinite 2>&1 | grep sleepInterval
      "sleepInterval": "2562047h47m16.854775807s",
```

api/config/v1/duration.go
```diff
@@ -19,12 +19,26 @@ package v1
...
+	}
+	return time.Duration(d).String()
+}
+
 // MarshalJSON marshals 'Duration' to its raw bytes representation
 func (d Duration) MarshalJSON() ([]byte, error) {
 	return json.Marshal(time.Duration(d).String())
```
elezar: This will fix outputing infinite:
```suggestion
	return json.Marshal(d.String())
```
```
 ./gpu-feature-discovery --oneshot --sleep-interval=infinite 2>&1 | grep sleepInterval
      "sleepInterval": "infinite",
```

api/config/v1/duration.go
rajathagasthya: I don't think there's an overflow risk here. We always check `IsInfinite()` before using the duration in `time.After()`, When `infinite`, we skip the timer entirely. `time.Duration(MaxInt64)` is also a valid duration, so I'd prefer to keep it this way than using `0` or `-1` sentinel values.

api/config/v1/duration.go
rajathagasthya: Thanks, I'll update this.

api/config/v1/duration.go
elezar: nit: We could introduce a constant for `Infinite` as well. Not a blocker for this PR.

rajathagasthya: Ack. I'll add that if there are other comments I need to address. If not, will do it in a follow up.

PR #1606, opened 2026-01-29
Add optional safety flag to prevent implicit GPU exposure without explicit resource requests

AkshatDudeja77:
This PR introduces an optional safety guard to prevent implicit GPU exposure when pods do not explicitly request `nvidia.com/gpu`.

### What this adds
- A new CLI flag `--require-gpu-request` (default: false)
- When enabled, the device plugin fails fast if the selected device-list strategy would allow GPUs to be exposed implicitly

### Motivation
While the current behavior is documented, it is easy to miss and can be unsafe in multi-tenant or hardened clusters. This opt-in flag allows operators to enforce explicit GPU requests without changing default behavior.

### Scope
- No config schema changes
- No behavior change unless the flag is explicitly enabled
- Enforcement happens early during startup before GPUs are registered with the kubelet

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -70,6 +70,12 @@ func main() {
 			Usage:   "fail the plugin if an error is encountered during initialization, otherwise block indefinitely",
 			EnvVars: []string{"FAIL_ON_INIT_ERROR"},
 		},
+		&cli.BoolFlag{
+			Name:    "require-gpu-request",
+			Value:   false,
+			Usage:   "require explicit nvidia.com/gpu requests; fail startup if GPUs would otherwise be exposed implicitly",
+			EnvVars: []string{"REQUIRE_GPU_REQUEST"},
+		},
 		&cli.StringFlag{
 			Name:    "driver-root",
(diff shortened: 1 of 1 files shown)


elezar:
@AkshatDudeja77 this is not in scope of the device plugin. Even when using a device list strategy such as `cdi-annotations`, `cdi-cri`, or `volume-mounts`, a container could still get access to GPUs using the `NVIDIA_VISIBLE_DEVICES` envvar if the *NVIDIA Container Toolkit* is configured to allow this.

Furthermore since containers that don't explicitly request GPUs don't pass through device plugin allocation, there is no way to detect that this is being done.

The flag *may* make sense in the context of the GPU Operator though, and I would recommend that you transfer this issue there to continue discussion.

AkshatDudeja77:
Thanks for the detailed explanation, that makes sense.

I agree that enforcing this at the device plugin level can’t fully address the issue given how NVIDIA_VISIBLE_DEVICES and the container toolkit behave outside of device plugin allocation.

I’ll close this PR and open an issue in the GPU Operator repo to continue the discussion at the appropriate layer. Appreciate the guidance.

PR #1615, opened 2026-02-06
update go version

rahulait:
Bump go version from 1.25.6 to 1.25.7


rahulait:
No longer required as we have bumped version in main via dependabot PR.

Issue #491, opened 2024-01-29
More flexible time-slicing strategy configuration


al42and:
_Note_: this is a copy of https://gitlab.com/nvidia/kubernetes/device-plugin/-/issues/8

### Issue or feature description

When a pod has requests/limits `nvidia.com/gpu: 2`, typically [one slice per physical device is allocated](https://gitlab.com/nvidia/kubernetes/device-plugin/-/merge_requests/224), and the container gets two devices, as intended.

However, when multiple pods are created on the host (e.g., in a CI system launching a batch of tasks), it both slices might end up on a single device. In this case, the CUDA application in the container only sees a single device, and the device-to-device communication tests cannot run.

It would be helpful to have a setting that makes the one-slice-per-device distribution a requirement rather than a best-effort attempt when distributing slices between devices. For our current case, a global setting would work.

We can envision the case for having this configurable per-pod. While we currently only run small CI tasks on our k8s cluster (which are not performance-sensitive), we are contemplating having more performance testing. For it, the opposite behavior would be needed: being able to request all slots on a single device. That, however, is not a near-term problem for our project: likely, any performance testing will be done on a separate infrastructure.

### Specific use case

- Small k8s cluster for running GitLab-based CI, with two NVIDIA T400 GPUs per node.
- CI tasks are pretty lightweight; time-slicing ("nvidia.com/gpu.replicas": "4") is used to utilize the hardware more efficiently.
- Three kinds of jobs: no GPU, single GPU, dual GPU. Latter require two different devices to test GPU-GPU communication (NVSHMEM, CUDA-aware MPI).
  - The number of devices can be checked by running `nvidia-smi -L` in the container.

For example, with a single node (2 GPUs, 4 slices each), allocating 1 then 5 then 2 slices triggers the problem for the last container:

```console
$ cat test_nvidia5.yml 
apiVersion: v1
kind: Pod
metadata:
  name: al42and-nvidia-test5
spec:
  containers:
    - image: registry.gitlab.com/gromacs/gromacs/ci-ubuntu-20.04-gcc-11-cuda-11.7.1
      name: al42and-nvidia-test5
      command:
        - /bin/sh
        - "-c"
        - "sleep 24h"
      resources:
        requests:
          cpu: 2
          memory: 2Gi
          nvidia.com/gpu: 2
        limits:
          cpu: 4
          memory: 4Gi
          nvidia.com/gpu: 2
  restartPolicy: Never


[...]

PR #1621, opened 2026-02-11
Add configurable allocation policy (packed/distributed) for replicated and MIG resources

wkd-woo:
## Summary

- Add `--allocation-policy` flag (`ALLOCATION_POLICY` env) with `distributed` (default) and `packed` options
- `packed` mode bin-packs replicated/MIG devices onto fewest physical GPUs, freeing up remaining GPUs for full-GPU workloads
- Default behavior (`distributed`) is unchanged — no breaking changes

## Motivation

The current `distributedAlloc` was designed for time-slicing, where distributing replicas across physical GPUs avoids compute contention. However, MIG devices also fall into this code path simply because `AlignedAllocationSupported()` returns `false` for them — not because distributed allocation is the right strategy.

MIG instances are hardware-isolated partitions with dedicated SMs and memory. Packing them onto fewer physical GPUs has no performance penalty, and frees up remaining GPUs for full-GPU workloads:

```
┌─────────────────────────────────────────────────────────────────────┐
│                Distributed Allocation (current default)              │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  GPU 0            GPU 1            GPU 2            GPU 3           │
│  ┌──────────┐    ┌──────────┐    ┌──────────┐    ┌──────────┐      │
│  │ ████░░░░ │    │ ████░░░░ │    │ ████░░░░ │    │ ░░░░░░░░ │      │
│  │ MIG 1/5  │    │ MIG 1/5  │    │ MIG 1/5  │    │ MIG 0/5  │      │
│  └──────────┘    └──────────┘    └──────────┘    └──────────┘      │
│   ⚠ partial       ⚠ partial       ⚠ partial       ░ empty          │
│                                                                     │
│  → Full GPU request arrives: only 1 GPU available (GPU 3)           │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────┐
│                    Packed Allocation (bin-packing)                   │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  GPU 0            GPU 1            GPU 2            GPU 3           │
│  ┌──────────┐    ┌──────────┐    ┌──────────┐    ┌──────────┐      │
│  │ ████████ │    │ ░░░░░░░░ │    │ ░░░░░░░░ │    │ ░░░░░░░░ │      │
│  │ MIG 3/5  │    │ MIG 0/5  │    │ MIG 0/5  │    │ MIG 0/5  │      │
│  └──────────┘    └──────────┘    └──────────┘    └──────────┘      │
│   ✓ packed         ★ free          ★ free          ★ free           │
│                                                                     │
│  → Full GPU request arrives: 3 GPUs available (GPU 1, 2, 3)        │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘
```

Relates to #491

## Design

```
getPreferredAllocation
├─ Full GPU (AlignedAllocationSupported && no annotations)
│   └─ alignedAlloc (topology-based, unaffected by this change)
│
└─ MIG / time-slicing / MPS
├─ allocationPolicy=packed  → packedAlloc (new)
└─ allocationPolicy=distributed (default) → distributedAlloc (existing, unmodified)
```

Key decisions:
- **`distributedAlloc` is completely untouched** — `packedAlloc` is a separate function following the existing `alignedAlloc`/`distributedAlloc` pattern
- **Full GPU nodes are unaffected** — `alignedAlloc` is selected before `allocationPolicy` is ever checked, so setting `packed` on a full-GPU node has no effect
- **Flag applies uniformly** — when `packed` is set, it applies to MIG, time-slicing, and MPS. Silently ignoring a user-set flag for specific device types would be inconsistent
- **Per-node config supported** — works with existing `config-manager` + ConfigMap + node label (`nvidia.com/device-plugin.config`) mechanism via YAML config

### CLI flag / Environment variable
```bash
--allocation-policy=packed
# or
ALLOCATION_POLICY=packed
Config file (per-node via ConfigMap + node label)

version: v1
flags:
  migStrategy: mixed
  plugin:
    allocationPolicy: packed
```

`kubectl label node mig-node nvidia.com/device-plugin.config=mig-packed`

Test plan
- [x] TestDistributedAlloc (6 cases) — existing distributed behavior regression
- [x] TestDistributedAllocIsDefault — default consistency verified over 10 iterations
- [x] TestPackedAlloc (6 cases) — bin-packing: same GPU priority, overflow to next GPU
- [x] TestPackedVsDistributedContrast — two strategies produce different results on same input
- [x] TestFullGPUNodeIgnoresAllocationPolicy (3 sub-cases) — full GPU / MIG / replicated branch path verification
 All existing internal/rm/ tests pass unchanged

---

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -143,6 +143,12 @@ func main() {
 			EnvVars: []string{"MPS_ROOT"},
 		},
 		&cli.StringFlag{
+			Name:    "shared-devices-allocation-policy",
+			Value:   spec.AllocationPolicyDistributed,
+			Usage:   "the allocation policy for replicated and MIG resources:\n\t\t[distributed | packed]",
+			EnvVars: []string{"SHARED_DEVICES_ALLOCATION_POLICY"},
+		},
+		&cli.StringFlag{
 			Name:    "device-discovery-strategy",
 			Value:   "auto",
 			Usage:   "the strategy to use to discover devices: 'auto', 'nvml', or 'tegra'",
@@ -207,6 +213,15 @@ func validateFlags(infolib nvinfo.Interface, config *spec.Config) error {
 
 	if *config.Flags.Plugin.DeviceIDStrategy != spec.DeviceIDStrategyUUID && *config.Flags.Plugin.DeviceIDStrategy != spec.DeviceIDStrategyIndex {
 		return fmt.Errorf("invalid --device-id-strategy option: %v", *config.Flags.Plugin.DeviceIDStrategy)
+	}
+
+	if config.Flags.Plugin.SharedDevicesAllocationPolicy != nil {
+		switch *config.Flags.Plugin.SharedDevicesAllocationPolicy {
+		case spec.AllocationPolicyDistributed:
+		case spec.AllocationPolicyPacked:
+		default:
+			return fmt.Errorf("invalid --shared-devices-allocation-policy option: %s", *config.Flags.Plugin.SharedDevicesAllocationPolicy)
+		}
 	}
 
 	if config.Sharing.SharingStrategy() == spec.SharingStrategyMPS {
--- a/internal/rm/allocate.go
+++ b/internal/rm/allocate.go
@@ -19,27 +19,67 @@ package rm
 import (
 	"fmt"
 	"sort"
+
+	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 )
 
-// distributedAlloc returns a list of devices such that any replicated
-// devices are distributed across all replicated GPUs equally. It takes into
-// account already allocated replicas to ensure a proper balance across them.
-func (r *resourceManager) distributedAlloc(available, required []string, size int) ([]string, error) {
-	// Get the set of candidate devices as the difference between available and required.
+// replicaCount tracks the total and available replica counts for a physical GPU.
+type replicaCount struct {
+	total     int
+	available int
+}
+
+// allocated returns the number of replicas currently allocated on the GPU.
+func (rc *replicaCount) allocated() int {
+	return rc.total - rc.available
+}
+
+// replicaComparator decides whether the physical GPU represented by i should
+// be preferred over the one represented by j when greedily selecting the next
+// device to allocate.
+type replicaComparator func(i, j *replicaCount) bool
+
+// allocationComparators maps each allocation policy to the comparator that
+// implements it. All policies share the same greedy selection loop
+// (greedyAlloc) and differ only in how the next best candidate is chosen.
+var allocationComparators = map[string]replicaComparator{
+	// distributed prefers GPUs with the fewest allocated replicas to spread
+	// workload evenly across physical GPUs.
+	spec.AllocationPolicyDistributed: func(i, j *replicaCount) bool {
+		return i.allocated() < j.allocated()
+	},
+	// packed prefers GPUs with the most allocated replicas to consolidate
+	// workloads onto fewer physical GPUs.
+	spec.AllocationPolicyPacked: func(i, j *replicaCount) bool {
+		return i.allocated() > j.allocated()
+	},
+}
+
+// comparatorForPolicy returns the comparator implementing the given
+// allocation policy. Unknown policies are rejected at startup, but fall back
+// to the default distributed policy here as a safety net.
+func comparatorForPolicy(policy string) replicaComparator {
+	if comparator, ok := allocationComparators[policy]; ok {
+		return comparator
+	}
+	return allocationComparators[spec.AllocationPolicyDistributed]
+}
+
+// prepareCandidates filters candidates from available devices (excluding required),
+// validates there are enough, and builds a per-GPU replica count map.
+func (r *resourceManager) prepareCandidates(available, required []string, size int) ([]string, map[string]*replicaCount, int, error) {
 	candidates := r.devices.Subset(available).Difference(r.devices.Subset(required)).GetIDs()
 	needed := size - len(required)
 
 	if len(candidates) < needed {
-		return nil, fmt.Errorf("not enough available devices to satisfy allocation")
+		return nil, nil, 0, fmt.Errorf("not enough available devices to satisfy allocation")
 	}
 
-	// For each candidate device, build a mapping of (stripped) device ID to
-	// total / available replicas for that device.
-	replicas := make(map[string]*struct{ total, available int })
+	replicas := make(map[string]*replicaCount)
 	for _, c := range candidates {
 		id := AnnotatedID(c).GetID()
 		if _, exists := replicas[id]; !exists {
-			replicas[id] = &struct{ total, available int }{}
+			replicas[id] = &replicaCount{}
 		}
 		replicas[id].available++
 	}
@@ -51,30 +91,40 @@ func (r *resourceManager) distributedAlloc(available, required []string, size in
 		replicas[id].total++
 	}
 
+	return candidates, replicas, needed, nil
+}
+
+// greedyAlloc returns a list of devices by repeatedly selecting the best
+// remaining candidate according to the supplied comparator. It takes into
+// account already allocated replicas so that consecutive allocations keep
+// following the policy the comparator implements.
+func (r *resourceManager) greedyAlloc(available, required []string, size int, preferred replicaComparator) ([]string, error) {
+	candidates, replicas, needed, err := r.prepareCandidates(available, required, size)
+	if err != nil {
+		return nil, err
+	}
+
 	// Track how many slots have already been picked from each physical device
-	// during this allocation. Used as the tie-break sort key below so the
-	// allocator rotates to a sibling physical device when the underlying
-	// "used" counts would otherwise tie.
+	// during this allocation. Used as the tie-break sort key below so that,
+	// when the comparator ranks two physical GPUs equally, the allocator
+	// rotates to a sibling device it has touched the least this round. This
+	// keeps the distributed policy spreading replicas across physical GPUs
+	// even when their allocated counts tie.
 	pickedFrom := make(map[string]int)
 
-	// Grab the set of 'needed' devices one-by-one from the candidates list.
-	// Before selecting each candidate, first sort the candidate list using the
-	// replicas map above. After sorting, the first element in the list will
-	// contain the device with the least difference between total and available
-	// replications (based on what's already been allocated). When two devices
-	// tie on that count, prefer the physical device we have not touched (or
-	// have touched the least) during this allocation. Add this device to the
-	// list of devices to allocate, remove it from the candidate list, down
-	// its available count in the replicas map, and repeat.
+	// Select devices one-by-one. The supplied comparator decides which
+	// physical GPU is preferred for the current policy. Comparators order
+	// solely by allocated() (see TestComparatorsOrderSolelyByAllocated), so
+	// equal allocated counts mean the comparator has no preference and the
+	// pickedFrom tie-break above applies.
 	var devices []string
 	for i := 0; i < needed; i++ {
 		sort.Slice(candidates, func(i, j int) bool {
 			iid := AnnotatedID(candidates[i]).GetID()
 			jid := AnnotatedID(candidates[j]).GetID()
-			idiff := replicas[iid].total - replicas[iid].available
-			jdiff := replicas[jid].total - replicas[jid].available
-			if idiff != jdiff {
-				return idiff < jdiff
+			ri, rj := replicas[iid], replicas[jid]
+			if ri.allocated() != rj.allocated() {
+				return preferred(ri, rj)
 			}
 			return pickedFrom[iid] < pickedFrom[jid]
 		})
@@ -85,8 +135,5 @@ func (r *resourceManager) distributedAlloc(available, required []string, size in
 		candidates = candidates[1:]
 	}
 
-	// Add the set of required devices to this list and return it.
-	devices = append(required, devices...)
-
-	return devices, nil
+	return append(required, devices...), nil
 }
--- a/api/config/v1/consts.go
+++ b/api/config/v1/consts.go
@@ -48,6 +48,12 @@ const (
 	DeviceIDStrategyIndex = "index"
 )
 
+// Constants to represent the various allocation policies
+const (
+	AllocationPolicyDistributed = "distributed"
+	AllocationPolicyPacked      = "packed"
+)
+
 // Constants related to generating CDI specifications
 const (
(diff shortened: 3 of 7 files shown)


ottowhite:
Hi! Very cool idea, I actually really want a very similar custom allocation policy for replicated (non-MIG) resources, but for enabling bin-packing replicas with topology-awareness. 

For example, this would could enable the allocation of 1.6 GPUs to a single model, with two tensor parallel ranks across 2 GPUs, 0.8 GPUs allocated to each tensor parallel tank, and both within a single NVLink domain (see my pretty diagram for an example). This would avoid excessive communication overhead during matrix multiplications across both tensor parallel model ranks, while leaving the 0.4 GPUs left over to be allocated to a smaller model, decreasing latency and increasing throughput in (surprisingly common) multi-model deployment scenarios including RAG with LLMs + Embedding models.

image

I believe the implementation of this policy is essentially a composite between what you've implemented, and the [alignedAlloc code path](https://github.com/NVIDIA/k8s-device-plugin/blob/fdbb5e093e3078547f2a89d0070cfcdda438b610/internal/rm/nvml_manager.go#L100-L109) which uses [go-gpuallocator](https://github.com/NVIDIA/go-gpuallocator) (gaining topology-awareness through NVML). At present, any of the replicated resource code paths discard topology awareness in favour of evenly load balancing GPU fractions with the distributed allocation policy, which prevents the use-case that I'm after.

Would be great if you're interested in discussing this more. I wonder if it's possible to slip it in with this PR? Also happy to contribute if that helps.

internal/rm/allocate.go
```diff
@@ -78,3 +78,61 @@ func (r *resourceManager) distributedAlloc(available, required []string, size in
...
+	// replications (i.e. the most already allocated). This packs allocations
+	// onto GPUs that are already in use, freeing up other GPUs entirely.
+	var devices []string
+	for i := 0; i < needed; i++ {
+		sort.Slice(candidates, func(i, j int) bool {
+			iid := AnnotatedID(candidates[i]).GetID()
+			jid := AnnotatedID(candidates[j]).GetID()
+			idiff := replicas[iid].total - replicas[iid].available
+			jdiff := replicas[jid].total - replicas[jid].available
+			return idiff > jdiff
```
ottowhite: Was the only change in this function and distributedAlloc the comparator between idiff and jdiff? Can we just pass in an enum and then have a condition on the enum to dictate this behavior? There is a lot of code duplication at the moment.

wkd-woo: Thanks for the review!

The duplication is intentional — I wanted distributedAlloc to stay completely untouched so reviewers can verify there's zero behavioral change just by looking at the diff.

On merging them: the two functions share the same structure because packed is just the inverse of distributed (flipped comparator). But as you mentioned with topology-aware bin-packing, a future policy would need NVML topology data and different sorting logic entirely — so I'd rather keep them separate now than merge and split again later.

That said, happy to extract the common setup (candidate filtering + replica counting) into a shared helper if that feels cleaner.

ottowhite: No problem :) to be honest I think it would actually be easier to review/maintain if you generalised the allocation function slightly. 

It could be generalised to fractionalAlloc, taking an enum that allows distributedFractionPlacement or packedFractionPlacement policies. Then internally you just pick the comparator based on the policy. 

What you suggested with pulling out the common helpers such as candidate filtering and replica counting is probably even better than the generalisation I suggested, as it enables the policies to be kept completely separate, and is more extensible in that it enables easier implementation of new policies again such as the topology-aware one.

When I was reviewing the code, I pulled up both distributedAlloc and packedAlloc side by side and was cross-checking for a while before I saw the flipped comparator. I expect other reviewers and contributors to go through a similar process. This is just my two cents though. 

You've also improved the state of testing so that's an additional signal that this is functionally equivalent to what there was before.

In terms of the different topology-aware placement policy, I think it makes sense to keep it separate for now. But cool that you're opening this up and making it more configurable! I may build off of this myself.

wkd-woo: @ottowhite I've extracted prepareCandidates in the second commit. 
Each policy function now only has its sorting logic, so the difference should be immediately visible

ottowhite: Looking better! There's still quite a bit of duplication across both of the Alloc functions. What is common is that they're both greedy allocation algorithms (repeatedly evaluating and selecting the next best option, without looking further ahead). Could you extract this into another helper greedyDeviceAlloc or something similar that makes the difference between these two functions even clearer? Could pass in a candidate comparator. It will also be much clearer to extend with other greedy allocation policies. For example the one that I spoke about is another greedy allocation algorithm that happens to use topology-awareness in it's comparator.

ottowhite:
Also, is there a reason why the alignedAlloc happens to be in nvml_manager.go? It seems like just another allocation policy that would be more suited to be in allocate.go rather than there. I know this is not due to your code change but moving it there could improve the organisation of the code, and ease to extend with different allocation policies.

wkd-woo:
Thank @ottowhite for your review :)
On moving alignedAlloc — agreed it could improve organization, but I'd rather keep this PR focused on the new feature and not restructure existing code!

I'd like to hear from a maintainer at this point to see if the current direction works.

ottowhite:
Fair enough! Though I think those refactors would be a minor improvement for organisation/extensibility, I think the testing is strong and implementation is solid. 

LGTM!

Will be great to hear what the maintainers have to say. Also really keen on this feature and the more extensible policy opportunities that it opens up.

ottowhite:
Hey! By the way we're running your packedAlloc allocation policy and it's working for multi-GPU deployments. We will imminently be forking and building off it it.

wkd-woo:
Hi @elezar, just checking in to see if there's anything else needed from my side to move this forward. I'm happy to address any feedback :)

internal/rm/allocate.go
rajatchopra: I agree that we should have a map of comparator functions to choose from. And a helper function that keeps the common portion in one place.

wkd-woo: Done in 4afcd2e3d :) `distributedAlloc`/`packedAlloc` are gone — there is now a single `greedyAlloc` that takes a comparator, plus an `allocationComparators` map keyed by policy. The difference between the two policies ends up being just `<` vs `>` in the map. Unknown policies fall back to distributed (covered by a new test), and the existing tests pass unchanged.

wkd-woo:
Hi @rajatchopra, could you take a look at this PR when you have a chance? 
I'd appreciate your feedback.

rajatchopra (approved):
/lgtm

Pending:
- rebase after #1788 merges
- squash commits

tariq1890:
@wkd-woo Please rebase your PR on top of the latest `main`.

wkd-woo:
@rajatchopra @tariq1890 Rebased onto latest `main` (now includes #1788) and squashed into a single commit as requested.

**Heads-up — this was not a clean rebase, so this needs another look before merge.** #1788 landed the day after your `/lgtm` and it touches the exact same allocation path this PR refactors, so I had to reconcile them by hand:

- #1788 added a `pickedFrom` tie-break to distributed allocation (rotate to a sibling physical GPU when two GPUs tie on allocated-replica count) plus a regression test `TestDistributedAlloc_PartiallyAllocated_DistributesAcrossDistinctGPUs`.
- This PR had already replaced `distributedAlloc`/`packedAlloc` with a single `greedyAlloc(comparator)`, so #1788's fix no longer had a place to live.

To keep both, I folded the tie-break into `greedyAlloc` as a secondary sort key applied only when the comparator ranks two GPUs equally:

```go
 sort.Slice(candidates, func(i, j int) bool {
     iid := AnnotatedID(candidates[i]).GetID()
     jid := AnnotatedID(candidates[j]).GetID()
-    return preferred(replicas[iid], replicas[jid])
+    ri, rj := replicas[iid], replicas[jid]
+    if preferred(ri, rj) {
+        return true
+    }
+    if preferred(rj, ri) {
+        return false
+    }
+    // comparator ranks the two GPUs equally -> prefer the physical
+    // device touched least during this allocation
+    return pickedFrom[iid] < pickedFrom[jid]
 })
```

This is the only logic change beyond the code you approved. It's required rather than optional: without it, #1788's regression test now fails on this branch (both replicas stack onto `GPU-0`); with it, the distributed policy keeps spreading. It's a no-op for `packed`, whose comparator dominates once packing starts. All `internal/rm` tests pass (`TestDistributedAlloc`, `TestPackedAlloc`, `TestDistributedAllocIsDefault`, `TestPackedVsDistributedContrast`, `TestFullGPUNodeIgnoresAllocationPolicy`, and #1788's `TestDistributedAlloc_PartiallyAllocated_...`).

PTAL when you get a chance.

wkd-woo:
cc @tariq1890 — flagging for the merge: this was rebased onto `main` after #1788 with **one small logic change beyond @rajatchopra's `/lgtm`** (folded #1788's distributed-allocation tie-break into this PR's new `greedyAlloc` — details and diff in the comment above). It's a no-op for `packed`, and all `internal/rm` tests pass including #1788's regression test. Should be ready to merge once CI is green — thanks!

rajatchopra:
@wkd-woo  >> To keep both, I folded the tie-break into greedyAlloc as a secondary sort key applied only when the comparator ranks two GPUs equally
That is good work resolving the conflicts. Thanks. Needs another benign rebase though.

rajatchopra (approved):
/lgtm

wkd-woo:
@rajatchopra Rebased onto latest main. Clean, no conflicts, and all tests pass. Thanks!

cc. @tariq1890

cmd/nvidia-device-plugin/main.go
```diff
@@ -209,6 +215,15 @@ func validateFlags(infolib nvinfo.Interface, config *spec.Config) error {
 		return fmt.Errorf("invalid --device-id-strategy option: %v", *config.Flags.Plugin.DeviceIDStrategy)
 	}
 
+	if config.Flags.Plugin.AllocationPolicy != nil {
+		switch *config.Flags.Plugin.AllocationPolicy {
+		case spec.AllocationPolicyDistributed:
+		case spec.AllocationPolicyPacked:
+		default:
+			return fmt.Errorf("invalid --allocation-policy option: %v", *config.Flags.Plugin.AllocationPolicy)
```
tariq1890:
```suggestion
			return fmt.Errorf("invalid --allocation-policy option: %s", *config.Flags.Plugin.AllocationPolicy)
```

internal/rm/allocate.go
```diff
@@ -19,27 +19,66 @@ package rm
...
 )
 
-// distributedAlloc returns a list of devices such that any replicated
-// devices are distributed across all replicated GPUs equally. It takes into
-// account already allocated replicas to ensure a proper balance across them.
-func (r *resourceManager) distributedAlloc(available, required []string, size int) ([]string, error) {
-	// Get the set of candidate devices as the difference between available and required.
+// replicaCount tracks the total and available replica counts for a physical GPU.
+type replicaCount struct {
+	total, available int
```
tariq1890:
```suggestion
	total     int
	available int
```

cmd/nvidia-device-plugin/main.go
```diff
@@ -142,6 +142,12 @@ func main() {
 			Usage:   "the path on the host where MPS-specific mounts and files are created by the MPS control daemon manager",
 			EnvVars: []string{"MPS_ROOT"},
 		},
+		&cli.StringFlag{
+			Name:    "allocation-policy",
+			Value:   spec.AllocationPolicyDistributed,
+			Usage:   "the allocation policy for replicated and MIG resources:\n\t\t[distributed | packed]",
+			EnvVars: []string{"ALLOCATION_POLICY"},
```
tariq1890: Can we come up with a more specific name here? The name `ALLOCATION_POLICY` does not lend itself to an allocation policy used for replicated and MIG devices.

internal/rm/allocate.go
```diff
@@ -51,30 +90,41 @@ func (r *resourceManager) distributedAlloc(available, required []string, size in
...
-			idiff := replicas[iid].total - replicas[iid].available
-			jdiff := replicas[jid].total - replicas[jid].available
-			if idiff != jdiff {
-				return idiff < jdiff
+			ri, rj := replicas[iid], replicas[jid]
+			if preferred(ri, rj) {
+				return true
+			}
+			if preferred(rj, ri) {
+				return false
```
tariq1890: Would this work?
```suggestion
			if ri.allocated() != rj.allocated() {
				return preferred(ri, rj)
			}
```

tariq1890:
@wkd-woo Thank you very much for your patience with this PR and I appreciate all the work you've put in. Just a few more comments, but we are definitely looking to merge this and ensure that this lands in the next release of the device-plugin.

cmd/nvidia-device-plugin/main.go
tariq1890: How about `SHARED_DEVICES_ALLOCATION_POLICY`?

cmd/nvidia-device-plugin/main.go
wkd-woo: 👍
Will Apply.

internal/rm/allocate.go
wkd-woo: 👍
Will Apply.

cmd/nvidia-device-plugin/main.go
wkd-woo: Agreed.

Will rename to `SHARED_DEVICES_ALLOCATION_POLICY` / `--shared-devices-allocation-policy`, and update the config field to match.

tariq1890: Thank you !

internal/rm/allocate.go
wkd-woo: Yes, that works. Both comparators only order by `allocated()`, so it comes out the same. I went with the two-way check to avoid assuming what the comparator compares, but with just these two policies that's probably overkill. 
Will apply it, and add a small test on the comparators to cover that assumption.

wkd-woo:
@tariq1890 Applied all four in [b8c575ef](https://github.com/NVIDIA/k8s-device-plugin/pull/1621/commits/b8c575ef654b0c98764df14b712e7fd3bbc0bd24), tests pass 🙏

tariq1890:
Can you squash your commit history?

wkd-woo:
@tariq1890 Squashed !

tariq1890 (approved):

tariq1890:
@wkd-woo. Your commit is not signed. See [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) for context.

wkd-woo:
@tariq1890 Signed the commit. Thank you for your patience.

PR #1623, opened 2026-02-16
add additional health checks

aburan28:
check for retired pages and temperature


rajatchopra:
This PR adds very useful set of health checks which we should definitely include in the device plugin.

However there is an effort going on to obtain the GPU health from a comprehensive tool (like [NvSentinel](https://github.com/nvidia/nvsentinel)) through a common API.
cc @ArangoGutierrez @dims @lalitadithya

PR #1625, opened 2026-02-19
Allow unsetting runtimeClassName

sklirg:
Default value to nil in case it is unset.
If the value previously has been set, but is now unset, the conditional branch would remove the field from the generated manifest, resulting in a manifest merge to retain the field as it was not specified to be removed/nulled.


rajatchopra:
Not sure if this is the right solution.
` | default nil` will result in rendering `runtimeClassName: <nil>` which isn't valid.

Closing this PR. Please re-open if anyone thinks it is still the right path.

PR #1667, opened 2026-03-19
add log verbosity configuration to plugin components

armourstill:
Sometimes we need to inspect logs from nvinfo, but its logging interface is not compatible with klog. To address this, I brought back the logger package and added a logging adapter for nvinfo. A fixed verbosity level of 1 is used to make debug logging as easy to enable as possible.


deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -37,6 +37,7 @@ gdrcopyEnabled: null
 gdsEnabled: null
 mofedEnabled: null
 deviceDiscoveryStrategy: null
+logVerbosity: null
```
rajatchopra, line 41: Do you want this to be set to `0`? To let the default be visible, as set [here](https://github.com/NVIDIA/k8s-device-plugin/pull/1667/changes#diff-bedb6af12e5f6e41bf5826b9433ea77929ba591a91f2026c6810ade99ff0871bR167)?

internal/logger/adapter.go
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+**/
+
+package logger
+
+import "k8s.io/klog/v2"
+
+// NvInfoAdapter is an adapter that implements the basicLogger interface using klog.
+type NvInfoAdapter struct {
+	DebugV int
```
rajatchopra: Call it verbosity?

rajatchopra: Or maybe level?

cmd/nvidia-device-plugin/main.go
```diff
@@ -347,6 +359,7 @@ func startPlugins(c *cli.Context, o *options) ([]plugin.Interface, bool, error)
 		nvinfo.WithRoot(string(driverRoot)),
 		nvinfo.WithNvmlLib(nvmllib),
 		nvinfo.WithDeviceLib(devicelib),
+		nvinfo.WithLogger(&logger.NvInfoAdapter{DebugV: 1}),
```
rajatchopra: hardcoded to `1`?

rajatchopra (changes requested):

internal/logger/adapter.go
armourstill: Good point. I renamed DebugV to Verbosity since it represents the klog verbosity level rather than a separate debug flag.

deployments/helm/nvidia-device-plugin/values.yaml
armourstill: I’d prefer to keep it as null, which means the Helm chart does not set LOG_VERBOSITY, so the binary uses its built-in default of 0. Setting it to 0 would be functionally equivalent, but it would make the chart explicitly override the application default rather than inherit it.

cmd/nvidia-device-plugin/main.go
armourstill: This was intentional. DebugV: 1 only maps nvinfo's Debugf messages to klog verbosity level 1; it does not enable them by default. The default global log-verbosity is still 0, so these messages are only shown when users explicitly increase verbosity.

armourstill: Also, the verbosity setting is now encapsulated in NewNvInfoAdapter(), so the call sites no longer need to hardcode it.

internal/logger/adapter.go
```diff
@@ -0,0 +1,42 @@
...
+package logger
+
+import "k8s.io/klog/v2"
+
+// defaultNvInfoVerbosity maps nvinfo debug messages to klog verbosity level 1.
+const defaultNvInfoVerbosity = 1
+
+// NvInfoAdapter is an adapter that implements the basicLogger interface using klog.
+type NvInfoAdapter struct {
+	Verbosity int
```
abrarshivani, line 26: Can we have this adapter accept a `logr.Logger` and construct it with `klog.Background()` instead of depending directly on global klog? This would keep it testable and backend-agnostic while preserving the current klog behavior, and provide a straightforward path to remove the adapter if `go-nvlib` supports `logr` directly in the future.

armourstill: Makes sense. I updated the adapter to take a logr.Logger and pass klog.Background() from the call sites. This keeps the current behavior while making the adapter easier to test and less tied to klog.

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -204,6 +204,10 @@ spec:
         {{- if typeIs "string" .Values.deviceDiscoveryStrategy }}
           - name: DEVICE_DISCOVERY_STRATEGY
             value: {{ .Values.deviceDiscoveryStrategy }}
+        {{- end }}
+        {{- if or (typeIs "string" .Values.logVerbosity) (typeIs "int64" .Values.logVerbosity) (typeIs "float64" .Values.logVerbosity) }}
```
kvalliyurnatt, line 208: Seems like this was not added for the MPS control daemonset

armourstill: Added LOG_VERBOSITY to the MPS control daemon daemonset as well

PR #1674, opened 2026-03-25
feat: exclude HAMI virtualized GPUs from NVIDIA device plugin via Node annotation watch

Zhangxiurui520:
### Background

In clusters where HAMi virtualizes selected GPUs, Node annotation hami.io/node-nvidia-register contains the GPU UUIDs managed by HAMi (virtualized GPUs).
The NVIDIA should avoid reporting these GPUs to kubelet to prevent resource overlap and scheduling conflicts.

### What this PR changes

1. Adds a Node annotation watcher in [main.go]:

- Uses Kubernetes Watch API (not polling) to watch only current Node (metadata.name=<NODE_NAME>).

- Watches annotation hami.io/node-nvidia-register.

- Parses GPU UUIDs (supports both GPU-... and hami-core:GPU-... token format).

- Triggers plugin update when annotation changes.

2. Extends plugin interface for dynamic filtering:

- Adds [HandleAllowedDeviceIDs([]string)] in [api.go].

- Device plugin implementation now handles runtime device-set updates.

3. Enables dynamic re-reporting to kubelet:

- In [server.go], adds an update signal channel.

- ListAndWatch now re-sends device list on filter updates, so kubelet sees changes without restarting plugin.

4. Adds filtering capability in resource manager:

- In [rm.go], stores both:
 full discovered device set ([allDevices]
，currently exposed device set ([devices]

- HandleAllowedDeviceIDs excludes GPUs present in HAMi annotation UUID list.

- Empty UUID list restores full device set.

- Adds RW mutex protection for concurrent read/update access.

5. Keeps resource manager constructors aligned:

NVML/Tegra resource managers initialize both [allDevices] and [devices] so runtime filtering is safe and reversible.

### Behavior
On startup: watcher does an initial Node GET sync.
On Node annotation update: plugin recalculates exposed GPUs and updates kubelet via ListAndWatch.
On annotation cleanup/removal: plugin restores full discovered GPU set.

### Why Watch instead of polling
Faster reaction to annotation changes.
Lower API overhead compared with periodic GET loops.
Cleaner event-driven behavior for production clusters.

### Compatibility / Notes

- Backward compatible when HAMi annotation is absent: plugin behavior remains unchanged.

- Requires running in cluster (InClusterConfig) and Node read/watch RBAC.

- Recommended to inject NODE_NAME via Downward API ([spec.nodeName] for accurate node targeting.

image
image

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -17,11 +17,14 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"syscall"
 	"time"
 
@@ -30,6 +33,12 @@ import (
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"github.com/fsnotify/fsnotify"
 	"github.com/urfave/cli/v2"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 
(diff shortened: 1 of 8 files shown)


rajatchopra:
I doubt if we should skip GPU enumeration based on a hard coded label key.
We need a better, generic design, though the dynamic part in this PR looks good.

kvalliyurnatt:
@Zhangxiurui520 are you still intersted in pursuing this and making changes based on the last comment ?

kvalliyurnatt:
There have been no updates on the requested changes, it has been a couple of months, closing this PR for now. Please reopen if you disagree

Issue #1677, opened 2026-03-28
[Bug]: Helm: gpu-feature-discovery DaemonSet missing volumeMounts for nvidiaDriverRoot + custom env needed


plevart:
Description:
In a Pure CDI architecture (deviceListStrategy: cdi-cri) without the nvidia-container-runtime wrapper, infrastructure pods like GFD are "blind" to NVIDIA libraries because they do not trigger CDI injection.

The Problem:
While the Helm chart provides the nvidiaDriverRoot variable to define the host path and creates the Volumes entry, it only adds the volumeMounts to the nvidia-device-plugin container. The gpu-feature-discovery container is missing these mounts.

Without the mount and a corresponding LD_PRELOAD to the mount path, GFD fails to load libnvidia-ml.so.1 on any system not using the legacy runtime wrapper.

Proposed Fix:
Sync the GFD template with the Device Plugin template to include:

volumeMounts: Mount nvidiaDriverRoot to a neutral path (e.g., /driver-root).

env: Add an optional custom env list so user can specify LD_PRELOAD pointing to the library within that mount to preload needed library(ies).

tariq1890:
Thank you @plevart for raising an issue. In regards to your statement

> In a Pure CDI architecture (deviceListStrategy: cdi-cri) without the nvidia-container-runtime wrapper, infrastructure pods like GFD are "blind" to NVIDIA libraries because they do not trigger CDI injection.

The is not entirely true. The GFD pod leverages a CDI device named `management.nvidia.com/gpu=all` CDI device. This CDI device is injected into the pod since it has the `nvidia` runtime class and has `NVIDIA_VISIBLE_DEVICES` (this holds true for GPU Operator versions v25.10.0 and above). This CDI device spec is generated by the nvidia-container-toolkit pod and you can see its contents in `/var/run/cdi/management.nvidia.com-gpu.yaml` on your host.

I would recommend looking at the following areas

- Container runtime config file. Ensure that the config file has been overridden correctly and you can see the `nvidia` runtime handler
- You see the management CDI device spec file generated in the `/var/run/cdi/` host directory.

Both of the above require the toolkit daemonset pod to run successfully.

plevart:
As I said, I'm not running the GPU operator. I'm not configuring special runtime (I'm using plain CRI-O). I just deployed node-feature-discovery Helm chart (standalone) with almost default values and nvidia-device-plugin Helm chart with the following values:

```yaml
gfd:
[...]

PR #1678, opened 2026-03-28
Helm chart: add volumeMounts for nvidiaDriverRoot + custom env. for NFD pod

plevart:
This is a proposed fix for https://github.com/NVIDIA/k8s-device-plugin/issues/1677

With this patch, I can specify something like the following in the Helm chart custom values.yaml:

```yaml
deviceListStrategy: cdi-cri
nvidiaDriverRoot: "/"
gfd:
  enabled: true
  env:
    # when deviceListStrategy=cdi-cri, there is no runc wrapper to inject the library
    - name: LD_PRELOAD
      value: /driver-root/usr/lib64/libnvidia-ml.so.1
```

...and the gfd Pod is happy.

I also cleaned-up and aligned the logic of this part of template with device plugin template.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -207,7 +207,7 @@ spec:
           - name: kubelet-device-plugins-dir
             mountPath: /var/lib/kubelet/device-plugins
         {{- if typeIs "string" .Values.nvidiaDriverRoot }}
-          # We always mount the driver root at /driver-root in the container.
+          # We mount the driver root at /driver-root in the container.
           # This is required for CDI detection to work correctly.
           - name: driver-root
             mountPath: /driver-root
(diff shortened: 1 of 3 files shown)


elezar (commented):
Although it may be useful to set the envvars for the GFD pod, I would argue that using this to set the location of libnvidia-ml.so.1 is not the correct mechanism. We already support loading a SPECIFIC path to libnvidia-ml.so.1 and the initialisation of the library here https://github.com/NVIDIA/k8s-device-plugin/blob/51f7ecebbd2a3b4e4eb53790691af29244b45c3d/cmd/gpu-feature-discovery/main.go#L182 should be updated.

See for example how this is done in the DRA driver here https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/ceb2f34a763def3531062efef01593f40861258e/cmd/gpu-kubelet-plugin/nvlib.go#L54-L69

Note that in the case of the GPU operator, the Device Plugin and GFD are consisdered "management" containers and are explicitly run using a specific runtime class to ensure that they have access to the requried devices and libraries. In cases where native CDI is used, we have an NRI plugin that ensures that these pods have GPU access based on pod annotations.

rajatchopra:
@plevart do you need to use this specifically for loading correct libnvidia-ml.so.1? If yes, then please set the correct root to nvml.New! Like Evan mentions, using LD_PRELOAD isn't incorrect but out of pattern.

closing this PR. Please absolutely re-open if you feel otherwise.

PR #1684, opened 2026-04-07
Allow setting the NVIDIA_DRIVER_CAPABILITIES environment variable

linkages:
This will allow setting the NVIDIA_DRIVER_CAPABILITIES environment variable on the nvidia-device-plugin-ctr container in the daemonset.

In my environment I need to change it from the default of "compute,utility" to "all".

This is required in my environment where I am using MiG. If this is not set then we get Insufficient Permissions errors from the container when trying to read memory information.

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -199,8 +199,10 @@ spec:
         {{- end }}
           - name: NVIDIA_VISIBLE_DEVICES
             value: all
+        {{ if typeIs "string" .Values.nvidiaDriverCapabilties }}
           - name: NVIDIA_DRIVER_CAPABILITIES
-            value: compute,utility
+            value: {{ .Values.nvidiaDriverCapabilities }}
+        {{- end }}
         securityContext:
           {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
         volumeMounts:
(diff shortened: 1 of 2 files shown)


rajatchopra (approved):
/lgtm

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -199,8 +199,10 @@ spec:
         {{- end }}
           - name: NVIDIA_VISIBLE_DEVICES
             value: all
+        {{ if typeIs "string" .Values.nvidiaDriverCapabilties }}
```
cdesiniotis, line 202:
```suggestion
        {{ if typeIs "string" .Values.nvidiaDriverCapabilities }}
```

tariq1890:
The templatisation is required here as well - https://github.com/NVIDIA/k8s-device-plugin/blob/db1ea9481054448d97ae43bd082147e7d6ba5501/deployments/helm/nvidia-device-plugin/templates/daemonset-mps-control-daemon.yml#L175

tariq1890:
Thank you for your PR. After addressing the comments, please ensure the following are taken care of as well:

- Commits are pushed with `--signoff`; this satisfies the DCO check
- Commits are signed and verified
- Commit history is squashed

abrarshivani:
@linkages gentle ping. Can you please address the review comments once you get a chance?

PR #1716, opened 2026-04-22
include component-specific DaemonSet selector labels

rahulait:
include component-specific DaemonSet selector labels to avoid selector overlap

Changes include:
- opt-in Helm setting to include component-specific DaemonSet selector labels to avoid selector overlap across device-plugin, GFD, and MPS control daemon
- disabled by default to avoid breaking upgrades (DaemonSet selectors are immutable and enabling requires DaemonSet recreation)

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -29,6 +29,9 @@ spec:
   selector:
     matchLabels:
       {{- include "nvidia-device-plugin.selectorLabels" . | nindent 6 }}
+      {{- with (include "nvidia-device-plugin.componentLabel" (dict "root" . "component" "device-plugin") | trim) }}
+      {{- . | nindent 6 }}
+      {{- end }}
   {{- with .Values.updateStrategy }}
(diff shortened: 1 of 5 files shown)


deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -29,6 +29,9 @@ spec:
   selector:
     matchLabels:
       {{- include "nvidia-device-plugin.selectorLabels" . | nindent 6 }}
+      {{- with (include "nvidia-device-plugin.componentLabel" (dict "root" . "component" "device-plugin") | trim) }}
```
rajatchopra, line 32: "device-plugin", or maybe "k8s-device-plugin"? Since we may have multiple ones in a cluster given node pools for confidential-containers.

rajatchopra (commented):
/lgtm

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
rahulait: Updated to be "k8s-device-plugin"

rajatchopra (approved):

Issue #1695, opened 2026-04-15
[Question]: Make Sharing.MPS.FailRequestsGreaterThanOne authoritative


radu-malliu:
Currently, `Sharing.MPS.FailRequestsGreaterThanOne` is a valid config item. However, its value is [ignored](https://github.com/NVIDIA/k8s-device-plugin/blob/eb98db48620947fd0235ca2bc66e06f8d3ac4bb9/internal/rm/rm.go#L101).

We have a use case whereby we need to be able to request all GPUs available on a node. Device plugin promises around what that request means in terms of time-on-device have no bearing, we know that this workload will have exclusivity when running. So, the only requirement is to be able to request multiple (all) replicas, in order to have access to all GPUs.

Based on this, is there interest in making the value of `Sharing.MPS.FailRequestsGreaterThanOne` authoritative when MPS is enabled, instead of it being ignored? Happy to PR the change if yes.

tariq1890:
> Device plugin promises around what that request means in terms of time-on-device have no bearing, we know that this workload will have exclusivity when running. So, the only requirement is to be able to request multiple (all) replicas, in order to have access to all GPUs.

So it looks like you want to be able to access all the GPUs while not taking away from the Allocatable GPU count for normal workloads. If I am understanding this correctly, would it suffice if you just set the `runtimeClassName: nvidia` and set the `NVIDIA_VISIBLE_DEVICES=all` envar in your pod which needs access to all of the GPUs?

pamelia:
Thanks, that may work as a debug/runtime escape hatch, but I do not think it is equivalent to what we need here.

For this case we do want the pod GPU access to go through the normal Kubernetes extended-resource allocation path. The verification pod is expected to reserve/request all MPS replicas while it runs, so the allocation is visible to kubelet/scheduler accounting rather than being an out-of-band runtime env override.

There is also an MPS-specific part of the device-plugin Allocate response: when MPS is enabled the plugin injects the MPS pipe/log directory env and mounts for the selected resource. A pod that relies only on `runtimeClassName: nvidia` and `NVIDIA_VISIBLE_DEVICES=all` would bypass the multi-replica request validation path this PR is changing, and depending on whether it requests a GPU resource at all, may not get the MPS allocation response data either.

[...]

PR #1762, opened 2026-05-12
Allow MPS failRequestsGreaterThanOne to be configured

pamelia:
## Summary
- preserve the existing conservative MPS default by treating an omitted `sharing.mps.failRequestsGreaterThanOne` as `true`
- honor an explicit `sharing.mps.failRequestsGreaterThanOne: false` for MPS configs
- validate MPS multi-unit shared GPU requests using the configured value instead of always rejecting them

## Motivation
This makes the MPS setting authoritative in the same way operators would expect from the config field, while keeping current default behavior unchanged. It is related to #1695.

## Tests
- `go test ./api/config/v1 ./internal/rm`

### DIFF
--- a/internal/rm/rm.go
+++ b/internal/rm/rm.go
@@ -92,20 +92,14 @@ func (r *resourceManager) ValidateRequest(ids AnnotatedIDs) error {
 	// error out if more than one resource is being allocated.
 	includesReplicas := ids.AnyHasAnnotations()
 	numRequestedDevices := len(ids)
+	failRequestsGTOne := r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne
 	switch r.config.Sharing.SharingStrategy() {
 	case spec.SharingStrategyTimeSlicing:
-		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
+		if includesReplicas && numRequestedDevices > 1 && failRequestsGTOne != nil && *failRequestsGTOne {
 			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
 		}
 	case spec.SharingStrategyMPS:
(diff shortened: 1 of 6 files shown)


internal/rm/rm.go
```diff
@@ -92,20 +92,14 @@ func (r *resourceManager) ValidateRequest(ids AnnotatedIDs) error {
...
 	numRequestedDevices := len(ids)
+	failRequestsGTOne := r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne
 	switch r.config.Sharing.SharingStrategy() {
 	case spec.SharingStrategyTimeSlicing:
-		if includesReplicas && numRequestedDevices > 1 && r.config.Sharing.ReplicatedResources().FailRequestsGreaterThanOne {
+		if includesReplicas && numRequestedDevices > 1 && failRequestsGTOne != nil && *failRequestsGTOne {
 			return fmt.Errorf("%w: maximum request size for shared resources is 1; found %d", errInvalidRequest, numRequestedDevices)
 		}
 	case spec.SharingStrategyMPS:
-		// For MPS sharing, we explicitly ignore the FailRequestsGreaterThanOne
```
shivamerla, line 101: I’m trying to understand the original intent behind this comment. MPS and time-slicing configs already expose different APIs today, so I’m not sure why we needed to extend the API further here.

pamelia: My read is that the old comment was explaining why MPS deliberately ignored the shared `ReplicatedResources.FailRequestsGreaterThanOne` field and always rejected multi-device requests.

This PR intentionally does not add another MPS-specific API. Since MPS already uses `ReplicatedResources` and the field is already present in `sharing.mps`, this just makes the existing field authoritative for MPS too. The default remains conservative because omitted MPS config is normalized to `true`; the only behavior change is for an explicit `sharing.mps.failRequestsGreaterThanOne: false`.

shivamerla: yes, that makes sense.

shivamerla (approved):

dims (approved):

PR #1763, opened 2026-05-13
[ci] run e2e tests remotely on the provisioned holodeck host

tariq1890:
This PR addresses the failing e2e tests that started surfacing since merging the holodeck version bump PR

#### Problem

The new versions of Holodeck include various security tightening improvements. Part of this is [restricting](https://github.com/NVIDIA/holodeck/pull/629) the permissions of the generated `kubeconfig` file from Holodeck. Given that the generated kubeconfigs are no longer world readable, we can no longer access the kubeconfig that is generated by holodeck from outside.

#### Solution

To address this, we copy the files from the github workspace over to the EC2 instance created by holodeck and run the e2e tests from inside that EC2 instance. This model of e2e test execution is already followed in [gpu-operator](https://github.com/NVIDIA/gpu-operator/tree/main/tests/scripts), so we follow the same approach here

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -55,24 +55,41 @@ jobs:
         aws_ssh_key: ${{ secrets.AWS_SSH_KEY }}
         holodeck_config: "tests/e2e/infra/aws.yaml"
 
+    - name: Get public dns name
+      id: get_public_dns_name
+      uses: mikefarah/yq@v4
+      with:
+        cmd: yq '.status.properties[] | select(.name == "public-dns-name") | .value' /github/workspace/.cache/holodeck.yaml
+
+    - name: Set test environment
+      env:
+        PUBLIC_DNS_NAME: ${{ steps.get_public_dns_name.outputs.result }}
+      run: |
+        echo "instance_hostname=ubuntu@${PUBLIC_DNS_NAME}" >> $GITHUB_ENV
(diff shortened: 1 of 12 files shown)


.github/workflows/e2e.yaml
```diff
@@ -55,24 +55,40 @@ jobs:
...
         E2E_IMAGE_REPO: ghcr.io/nvidia/k8s-device-plugin
         E2E_IMAGE_TAG: ${{ inputs.version }}
         E2E_IMAGE_PULL_POLICY: Always
         NVIDIA_DRIVER_ENABLED: true
-        LOG_ARTIFACTS: ${{ github.workspace }}/e2e_logs
       run: |
-        make -f tests/e2e/Makefile test
-
+        ./tests/ci-run-e2e.sh ${E2E_IMAGE_REPO} ${E2E_IMAGE_TAG} ${NVIDIA_DRIVER_ENABLED}
+        ./tests/scripts/pull.sh /tmp/logs logs
```
rajathagasthya, line 85: I think this steps runs these commands as `ci-run-e2e.sh && pull.sh`. This means if the test fails, `pull.sh` never runs and no logs are uploaded. We might need to run something like `trap pull.sh EXIT`.

tariq1890: Agreed. Thanks for catching that!

tariq1890: Fixed.

rajathagasthya (approved):

PR #1787, opened 2026-05-19
enable per-physical-GPU replica counts in time-slicing and MPS

### LINKED ISSUE #1786: Support per-physical-GPU replica counts in sharing.timeSlicing and sharing.mps
## Summary

Allow operators to configure **different replica counts for different physical GPUs on the same node** under the existing `sharing.timeSlicing` and `sharing.mps` config. The schema for this already exists (`ReplicatedResource.Devices` + `ReplicatedResource.Rename`); the device-map construction in `internal/rm/device_map.go` already supports it. The only thing blocking the feature is one helper (`disableResoureRenaming` in `api/config/v1/replicas.go`) that strips these fields on config load.

## Motivation

Today, every physical GPU on a node ends up with the same replica count because `sharing.timeSlicing.resources` collapses to a single homogeneous entry. That's fine for symmetric nodes, but it forces an all-or-nothing trade-off: either every GPU on the node gets aggressive sharing (more slices, smaller share each) or none does.

Real-world cases where per-GPU replica counts would help:

- **Light + heavy on the same node.** Reserve one GPU at `replicas: 2` for latency-sensitive inference (50% share each, two co-tenants max) and another at `replicas: 8` for batch jobs that don't mind small slices.
- **Mixed GPU classes on one node** (different ages or memory tiers). With distinct `rename`s, each GPU advertises as a different resource name — and consumers explicitly request the tier they need.
- **MPS specifically.** Because MPS enforces `CUDA_MPS_ACTIVE_THREAD_PERCENTAGE = 100 / replicas`, per-GPU replicas translates directly into per-GPU compute caps. This is where the feature is most semantically clean.

Per-node configs (via the GPU Operator's labelled-config map) only solve this when entire nodes are homogeneous. Within a single node, no current option works.

## Proposal

After the change, this config would do what it reads as:

```yaml
version: v1
sharing:
  timeSlicing:
    resources:
      - name: nvidia.com/gpu
        rename: nvidia.com/gpu-light
        devices: ["0"]
        replicas: 2
      - name: nvidia.com/gpu
        rename: nvidia.com/gpu-heavy
        devices: ["1"]
        replicas: 8
```

Pods request `nvidia.com/gpu-light` or `nvidia.com/gpu-heavy` explicitly. Behavior for configs that omit `devices:` / `rename:` is unchanged (existing single-resource configs keep working).

## What this is NOT trying to do

- **Not changing time-slicing compute semantics.** Time-slicing on a single physical GPU round-robins CUDA contexts in the closed-source driver. Holding more time-slicing replicas on the same physical GPU does *not* give a pod a larger compute share — the device plugin has no way to inject scheduler weight there. The `failRequestsGreaterThanOne: true` default still applies and is still the right guardrail for time-slicing.
  - For **MPS**, the picture is cleaner: replicas → `CUDA_MPS_ACTIVE_THREAD_PERCENTAGE`, which the MPS daemon actually enforces. So MPS users get the intuitive "more replicas per GPU = more compute share" behavior, and the per-GPU `replicas` config becomes a per-GPU compute-cap knob.
- **Not introducing new k8s scheduling semantics.** The existing per-resource-name model is unchanged. Pods just see two (or more) distinct resource names instead of one.

## Prior art

The same feature gap has been raised before. None of these were closed by a decision; most went stale:

- #628 "Dedicated GPU's for time slicing on multi GPU set ups." Closed stale. @frittentheke identified the exact `disableResoureRenaming` gate.
- #491 "More flexible time-slicing strategy configuration." Open. @klueska: *"not opposed in principle"* but raised scheduler-integration concerns.
- #1422 "Heterogeneous GPU MPS Replicas Configuration." Closed; the answer was "use per-node configs via the GPU Operator," which doesn't help within-node heterogeneity.
- #1018 (PR) "bugfix: Allow for optional TimeSlicing configuration." Open. Adjacent area.
- #1621 (PR) "Add configurable allocation policy (packed/distributed) for replicated and MIG resources." Open since Feb 2026; touches the same allocator. Happy to coordinate with @wkd-woo to avoid conflicts.

The comment on `DisableResourceNamingInConfig` already hints at intent: *"This may be reenabled in a future release."*

## Next steps

We have a working implementation with tests against `5c15cbe` and will open a PR shortly. Filing this issue first so the design context isn't buried inside the PR description.

cc @elezar @klueska

jonathan-meiri:
## Summary

Enable per-physical-GPU replica counts in `sharing.timeSlicing` and `sharing.mps` config. See #1786 for the design context (prior art, what this does and doesn't try to do, MPS vs time-slicing semantics).

Closes #1786

## Changes

- `api/config/v1/replicas.go`: replace `disableResoureRenaming` with `applyDefaults`. The new helper still fills the existing defaults (auto-rename when `renameByDefault: true` and `Rename` is unset; default `Devices.All = true` when no selector is set) but no longer strips user-supplied `Rename` or `Devices.List` / `Devices.Count`. The two "not yet supported in the config" warnings go away.
- `api/config/v1/config.go`: update `DisableResourceNamingInConfig` to call the new helper. Function name and public signature unchanged.
- `internal/rm/allocate.go`: in `distributedAlloc`, change the sort key from `min(total - available)` to `max(available)`. The two orderings are equivalent in the homogeneous case, but the previous key biased toward exhausting smaller physical GPUs first when replica counts differ. The now-unused `total` field and its initialization loop are removed.

Device-map construction in `internal/rm/device_map.go` is unchanged — it already handles `Devices.All` / `Devices.Count` / `Devices.List` correctly. Net behavior change is ~50 LoC plus tests.

Commits are DCO-signed.

### DIFF
--- a/api/config/v1/replicas.go
+++ b/api/config/v1/replicas.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 
 	"github.com/google/uuid"
-	"k8s.io/klog/v2"
 )
 
 // ReplicatedResources defines generic options for replicating devices.
@@ -33,36 +32,27 @@ type ReplicatedResources struct {
 	Resources                  []ReplicatedResource `json:"resources,omitempty"                  yaml:"resources,omitempty"`
 }
 
-func (rrs *ReplicatedResources) disableResoureRenaming(id string) {
+// applyDefaults fills in default values for entries in rrs.Resources, leaving
+// any user-supplied Rename or Devices selection in place. A nil receiver is a
+// no-op.
+//
+// Two defaults are applied:
+//   - If RenameByDefault is true and an entry has no explicit Rename, the
+//     entry's Rename is set to the default shared name for its Name.
+//   - If an entry leaves Devices unset (no All / Count / List), Devices.All
+//     is set to true so that the entry replicates every matching device.
+func (rrs *ReplicatedResources) applyDefaults() {
 	if rrs == nil {
 		return
 	}
-	renameByDefault := rrs.RenameByDefault
-	setsNonDefaultRename := false
-	setsDevices := false
 	for i, r := range rrs.Resources {
-		if !renameByDefault && r.Rename != "" {
-			setsNonDefaultRename = true
-			rrs.Resources[i].Rename = ""
-		}
-		if renameByDefault && r.Rename != r.Name.DefaultSharedRename() {
-			setsNonDefaultRename = true
+		if rrs.RenameByDefault && r.Rename == "" {
 			rrs.Resources[i].Rename = r.Name.DefaultSharedRename()
 		}
-		if !r.Devices.All {
-			setsDevices = true
+		if !r.Devices.All && r.Devices.Count == 0 && len(r.Devices.List) == 0 {
 			rrs.Resources[i].Devices.All = true
-			rrs.Resources[i].Devices.Count = 0
-			rrs.Resources[i].Devices.List = nil
 		}
 	}
-	if setsNonDefaultRename {
-		klog.Warningf("Setting the 'rename' field in sharing.%s.resources is not yet supported in the config. Ignoring...", id)
-	}
-	if setsDevices {
-		klog.Warningf("Customizing the 'devices' field in sharing.%s.resources is not yet supported in the config. Ignoring...", id)
-	}
-
 }
 
 func (rrs *ReplicatedResources) isReplicated() bool {
--- a/cmd/mps-control-daemon/mps/daemon.go
+++ b/cmd/mps-control-daemon/mps/daemon.go
@@ -234,8 +234,23 @@ func (d *Daemon) EchoPipeToControl(command string) (string, error) {
 	return out.String(), nil
 }
 
+// sharedDevices returns the subset of the daemon's devices that are actually
+// MPS-shared (i.e., replicated with annotated IDs). Devices that end up in
+// the same resource manager but are not being replicated (per the sharing
+// config's Devices selector) are excluded so their compute mode and memory
+// limits are not touched by MPS setup.
+func (d *Daemon) sharedDevices() rm.Devices {
+	result := make(rm.Devices)
+	for id, dev := range d.rm.Devices() {
+		if rm.AnnotatedID(id).HasAnnotations() {
+			result[id] = dev
+		}
+	}
+	return result
+}
+
 func (d *Daemon) setComputeMode(mode computeMode) error {
-	for _, uuid := range d.Devices().GetUUIDs() {
+	for _, uuid := range d.sharedDevices().GetUUIDs() {
 		cmd := exec.Command(
 			"nvidia-smi",
 			"-i", uuid,
@@ -253,7 +268,7 @@ func (d *Daemon) setComputeMode(mode computeMode) error {
 func (m *Daemon) perDevicePinnedDeviceMemoryLimits() map[string]string {
 	totalMemoryInBytesPerDevice := make(map[string]uint64)
 	replicasPerDevice := make(map[string]uint64)
-	for _, device := range m.Devices() {
+	for _, device := range m.sharedDevices() {
 		index := device.Index
 		totalMemoryInBytesPerDevice[index] = device.TotalMemory
 		replicasPerDevice[index] += 1
@@ -271,10 +286,11 @@ func (m *Daemon) perDevicePinnedDeviceMemoryLimits() map[string]string {
 }
 
 func (m *Daemon) activeThreadPercentage() string {
-	if len(m.Devices()) == 0 {
+	shared := m.sharedDevices()
+	if len(shared) == 0 {
 		return ""
 	}
-	replicasPerDevice := len(m.Devices()) / len(m.Devices().GetUUIDs())
+	replicasPerDevice := len(shared) / len(shared.GetUUIDs())
 
 	return fmt.Sprintf("%d", 100/replicasPerDevice)
 }
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -78,8 +78,13 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
 	return config, nil
 }
 
-// DisableResourceNamingInConfig temporarily disable the resource renaming feature of the plugin.
-// This may be reenabled in a future release.
+// DisableResourceNamingInConfig disables the top-level resource renaming
+// feature (config.Resources.GPUs / config.Resources.MIGs) and applies
+// default values to the per-resource sharing entries.
+//
+// Per-entry Rename and Devices selection in Sharing.TimeSlicing and
+// Sharing.MPS are preserved as the user configured them; only unset fields
+// receive defaults.
 func DisableResourceNamingInConfig(config *Config) {
 	// Disable resource renaming through config.Resource
 	if len(config.Resources.GPUs) > 0 || len(config.Resources.MIGs) > 0 {
@@ -88,10 +93,9 @@ func DisableResourceNamingInConfig(config *Config) {
 	config.Resources.GPUs = nil
 	config.Resources.MIGs = nil
 
-	// Disable renaming / device selection in Sharing.TimeSlicing.Resources
-	config.Sharing.TimeSlicing.disableResoureRenaming("timeSlicing")
-	// Disable renaming / device selection in Sharing.MPS.Resources
-	config.Sharing.MPS.disableResoureRenaming("mps")
+	// Apply defaults (auto-rename, default Devices.All) to sharing entries.
+	config.Sharing.TimeSlicing.applyDefaults()
+	config.Sharing.MPS.applyDefaults()
 }
 
 // parseConfig parses a config file as either YAML of JSON and unmarshals it into a Config struct.
--- a/api/config/v1/config_test.go
+++ b/api/config/v1/config_test.go
@@ -57,6 +57,84 @@ sharing:
 	require.False(t, *config.Sharing.MPS.FailRequestsGreaterThanOne)
 }
 
+func TestDisableResourceNamingInConfig_PreservesPerDeviceTimeSlicingSelection(t *testing.T) {
+	cfg := &Config{
+		Sharing: Sharing{
+			TimeSlicing: ReplicatedResources{
+				Resources: []ReplicatedResource{
+					{
+						Name:     ResourceName("nvidia.com/gpu"),
+						Rename:   ResourceName("nvidia.com/gpu-light"),
+						Devices:  ReplicatedDevices{List: []ReplicatedDeviceRef{"0"}},
+						Replicas: 2,
+					},
+					{
+						Name:     ResourceName("nvidia.com/gpu"),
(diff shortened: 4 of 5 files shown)


jonathan-meiri:
Contributing on behalf of @runatom-ai.

Dorsivan:
This can be extremely useful, is there any way I can help advance this?

jonathan-meiri:
Hi — friendly nudge on this one. It's been open a couple of months and hasn't had a maintainer look yet, so I wanted to resurface it.

It's a small change (~200 LoC in `api/config/v1`) that lets operators set per-GPU `devices:` and `rename:` in both time-slicing and MPS sharing configs. It also fixes the MPS daemon so it only adjusts the GPUs it's actually sharing, rather than every GPU on the node.

Happy to rebase or address any feedback whenever someone has a chance to take a look. Thanks!

cc @rajatchopra @tariq1890

myeolenv:
Thanks for the contribution. We will review this soon.

api/config/v1/replicas.go
```diff
@@ -33,36 +32,27 @@ type ReplicatedResources struct {
...
-			rrs.Resources[i].Rename = ""
-		}
-		if renameByDefault && r.Rename != r.Name.DefaultSharedRename() {
-			setsNonDefaultRename = true
+		if rrs.RenameByDefault && r.Rename == "" {
 			rrs.Resources[i].Rename = r.Name.DefaultSharedRename()
 		}
-		if !r.Devices.All {
-			setsDevices = true
+		if !r.Devices.All && r.Devices.Count == 0 && len(r.Devices.List) == 0 {
```
kvalliyurnatt, line 52: Multiple entries with the same Name re-add each other’s selected devices. For [0] -> gpu-light and [1] -> gpu-heavy, both GPUs also remain under nvidia.com/gpu.

Can we collect selected IDs per source resource, then add the difference once?
```
selectedByName[r.Name][id] = device
// After processing all entries:
remaining := oDevices[name].Difference(selectedByName[name])
```

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -234,8 +234,23 @@ func (d *Daemon) EchoPipeToControl(command string) (string, error) {
...
 }
 
+// sharedDevices returns the subset of the daemon's devices that are actually
+// MPS-shared (i.e., replicated with annotated IDs). Devices that end up in
+// the same resource manager but are not being replicated (per the sharing
+// config's Devices selector) are excluded so their compute mode and memory
+// limits are not touched by MPS setup.
+func (d *Daemon) sharedDevices() rm.Devices {
+	result := make(rm.Devices)
+	for id, dev := range d.rm.Devices() {
```
kvalliyurnatt, line 255: The plugin still enables MPS for unannotated resources, even though the daemon manager skips them. It then waits for a daemon that was never created. Could we add a check like below  in getMPSOptions?
```
if !rm.AnnotatedIDs(resourceManager.Devices().GetIDs()).AnyHasAnnotations() {
    return mpsOptions{}, nil
}
```

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -271,10 +286,11 @@ func (m *Daemon) perDevicePinnedDeviceMemoryLimits() map[string]string {
 }
 
 func (m *Daemon) activeThreadPercentage() string {
-	if len(m.Devices()) == 0 {
+	shared := m.sharedDevices()
+	if len(shared) == 0 {
 		return ""
 	}
-	replicasPerDevice := len(m.Devices()) / len(m.Devices().GetUUIDs())
+	replicasPerDevice := len(shared) / len(shared.GetUUIDs())
```
kvalliyurnatt: I think ends up averaging replica count across the GPUs. With 2 and 4 replicas, both GPUs get 33% instead of 50% and 25%.

kvalliyurnatt (changes requested):

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -234,8 +234,23 @@ func (d *Daemon) EchoPipeToControl(command string) (string, error) {
 	return out.String(), nil
 }
 
+// sharedDevices returns the subset of the daemon's devices that are actually
+// MPS-shared (i.e., replicated with annotated IDs). Devices that end up in
+// the same resource manager but are not being replicated (per the sharing
+// config's Devices selector) are excluded so their compute mode and memory
+// limits are not touched by MPS setup.
+func (d *Daemon) sharedDevices() rm.Devices {
```
abrarshivani, line 253: Currently, `sharedDevices()` scopes the compute mode and memory setup to the GPUs in this resource manager, but `nvidia-cuda-mps-control -d` is still started without `CUDA_VISIBLE_DEVICES`. This means MPS control daemons created for different resource managers can still see the same GPUs, so `set_default_active_thread_percentage` may not apply as expected since it is a control-daemon setting rather than a per-device setting. Can we set `CUDA_VISIBLE_DEVICES` so that each MPS control daemon is limited to the GPUs in its resource manager?

jonathan-meiri: Good point, each control daemon starts without CUDA_VISIBLE_DEVICES, so their scopes overlap. (The thread-% part is now moot since I moved it to a per-client env,  but the isolation gap still applies to the memory/compute-mode setup and to which GPUs a daemon's clients are confined to.)

I've opened #2056 to handle this as a focused follow-up once this PR merges.
it's a self-contained isolation change, and setting CUDA_VISIBLE_DEVICES also shifts the device indices the control commands use, so the memory-limit indexing needs remapping.

api/config/v1/replicas.go
jonathan-meiri: Fixed.
selected IDs are now unioned per source resource and leftovers computed once, so same-Name entries don't re-add each other's devices. Added a test.

cmd/mps-control-daemon/mps/daemon.go
jonathan-meiri: Fixed.
getMPSOptions now skips resources with no annotated devices, matching the daemon manager, so it won't wait on a daemon that was never created. Added a test.

cmd/mps-control-daemon/mps/daemon.go
jonathan-meiri: Fixed.
moved thread % off the averaged daemon default to a per-client CUDA_MPS_ACTIVE_THREAD_PERCENTAGE env from the device's replica count (mixed containers capped by 100/max). Added a test.

cmd/mps-control-daemon/mps/daemon.go
kvalliyurnatt: @jonathan-meiri I think I prefer addressing this as part of the current PR instead of splitting it up, it is a substantial change but I think it needs to be part of the current changes since it is directly related to the change  in this PRand would prefer we don't merge this PR without that change

jonathan-meiri: Fair.
I pushed a new commit where each MPS control daemon is now scoped via CUDA_VISIBLE_DEVICES to its resource's GPUs, with per-device pinned-mem limits keyed by the ordinal in that visible set (+ unit tests). Closing #2056 as folded in here.

Thanks for the thoughtful review, really appreciate it !

internal/rm/device_map.go
kvalliyurnatt, line 256: While not a part of the changes here, we don't have a determinstic order of the devices, so the plugin and the mps daemon might end up selecting different devices, I think we should sort them by UUID before we pick the devcies

internal/plugin/mps.go
```diff
@@ -71,13 +77,19 @@ func (m *mpsOptions) waitForDaemon() error {
 	return nil
 }
 
-func (m *mpsOptions) updateReponse(response *pluginapi.ContainerAllocateResponse) {
+func (m *mpsOptions) updateReponse(response *pluginapi.ContainerAllocateResponse, ids []string) {
 	if m == nil || !m.enabled {
 		return
 	}
 	// TODO: We should check that the deviceIDs are shared using MPS.
 	response.Envs["CUDA_MPS_PIPE_DIRECTORY"] = m.daemon.PipeDir()
```
kvalliyurnatt, line 85: I think we need to address this TODO as well, earlier mixed mode under a resource manager was not possible, now with the PRs changes we can have something like
```
nvidia.com/gpu:
  GPU-0::0   # MPS shared
  GPU-0::1   # MPS shared
  GPU-1      # unshared
  GPU-2      # unshared
  GPU-3      # unshared
```

When there is no rename in the configuration, In such mixed resource managers, we might select an unshared GPU as part of the allocation and pass in the pipe directory, but the MPS daemon does not manager this GPU resource at all.

I think we should
1. Disallow such mixed mode
something like
```
func classifyMPSRequest(ids []string) (bool, error) {
	var hasShared, hasUnshared bool
	for _, id := range ids {
		if rm.AnnotatedID(id).HasAnnotations() {
			hasShared = true
		} else {
			hasUnshared = true
		}
	}
	if hasShared && hasUnshared {
		return false, fmt.Errorf("cannot mix shared and unshared GPUs in one MPS allocation")
	}
	return hasShared, nil
}
```
2. Only pass in the pipe if the RM has MPS shared devices
```
func (m *mpsOptions) updateResponse(
	response *pluginapi.ContainerAllocateResponse,
	ids []string,
) error {
	if m == nil || !m.enabled {
		return nil
	}
	usesMPS, err := classifyMPSRequest(ids)
	if err != nil {
		return err
	}
	if !usesMPS {
		return nil
	}
	response.Envs["CUDA_MPS_PIPE_DIRECTORY"] = m.daemon.PipeDir()
	// Existing thread-percentage and mount handling.
	return nil
}
```

internal/rm/device_map.go
```diff
@@ -311,11 +317,7 @@ func updateDeviceMapWithReplicas(replicatedResources *spec.ReplicatedResources,
 		if len(ids) == 0 {
 			continue
 		}
-
-		// Add any devices we don't want replicated directly into the device map.
-		for _, d := range oDevices[r.Name].Difference(oDevices[r.Name].Subset(ids)) {
-			devices.insert(r.Name, d)
-		}
+		selectedIDs[r.Name] = append(selectedIDs[r.Name], ids...)
```
kvalliyurnatt, line 320: I think we can have an issue where the same physical GPU is part of multiple resources, something like
```
sharing:
  mps:
    resources:
      - name: nvidia.com/gpu
        rename: nvidia.com/gpu-light
        devices: ["0"]
        replicas: 2
      - name: nvidia.com/gpu
        rename: nvidia.com/gpu-heavy
        devices: ["0"] # Same physical GPU
        replicas: 4
```

I think here we should make sure they are not 
maybe we can do something like
```
selected := selectedIDs[r.Name]
if selected == nil {
	selected = make(map[string]struct{})
	selectedIDs[r.Name] = selected
}
for _, id := range ids {
	if _, exists := selected[id]; exists {
		return nil, fmt.Errorf(
			"device %q is selected by multiple entries for resource %q",
			id,
			r.Name,
		)
	}
	selected[id] = struct{}{}
}
```

kvalliyurnatt (changes requested):

PR #1788, opened 2026-05-19
fix distributedAlloc tie-break to preserve distribution across physical GPUs

jonathan-meiri:
## Summary

`distributedAlloc` is designed to spread allocations across replicated physical GPUs evenly, but a tie in its sort key — easy to reach once any prior pod has consumed a slot on one of the GPUs — falls through to `sort.Slice`'s arbitrary order. In practice this means the loop keeps picking the next slot on the GPU it just picked from, instead of rotating to a sibling physical GPU that still has capacity.

## What's here

Two commits, TDD-style:

1. **A failing test** describing the bug — 2 physical GPUs, 1 slot on GPU-1 already allocated to another pod, new pod requests 2 slots. Today both slots deterministically land on GPU-0; the test asserts 1-from-each.
2. **The fix** — introduce a small `pickedFrom` map and consult it as a *tie-break* sort key. The existing ordering remains primary. Behavior is unchanged whenever the primary sort key differs; the change only touches the previously-arbitrary tied case.

Commits are DCO-signed.

### DIFF
--- a/internal/rm/allocate.go
+++ b/internal/rm/allocate.go
@@ -51,13 +51,21 @@ func (r *resourceManager) distributedAlloc(available, required []string, size in
 		replicas[id].total++
 	}
 
+	// Track how many slots have already been picked from each physical device
+	// during this allocation. Used as the tie-break sort key below so the
+	// allocator rotates to a sibling physical device when the underlying
+	// "used" counts would otherwise tie.
+	pickedFrom := make(map[string]int)
+
 	// Grab the set of 'needed' devices one-by-one from the candidates list.
 	// Before selecting each candidate, first sort the candidate list using the
 	// replicas map above. After sorting, the first element in the list will
 	// contain the device with the least difference between total and available
-	// replications (based on what's already been allocated). Add this device
-	// to the list of devices to allocate, remove it from the candidate list,
-	// down its available count in the replicas map, and repeat.
+	// replications (based on what's already been allocated). When two devices
+	// tie on that count, prefer the physical device we have not touched (or
+	// have touched the least) during this allocation. Add this device to the
+	// list of devices to allocate, remove it from the candidate list, down
+	// its available count in the replicas map, and repeat.
 	var devices []string
 	for i := 0; i < needed; i++ {
 		sort.Slice(candidates, func(i, j int) bool {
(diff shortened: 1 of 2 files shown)


dims:
cc @cdesiniotis @tariq1890

tariq1890:
@jonathan-meiri Can you squash your commit history?

jonathan-meiri:
@tariq1890 done 😄

rajatchopra:
The changes look good to me. However, I feel the entire loop should be optimized. Currently, `n` needed allocations sort the gpus by available capacity `n` times. One sort should be enough and then we can decide on distributed allocation or a packed one, or a topology driven one. Also see https://github.com/NVIDIA/k8s-device-plugin/pull/1621

jonathan-meiri:
Thanks @rajatchopra — both points well taken.

On the strategy framework: agreed that's a useful direction. I see #1621 is exploring similar ground; once this lands I'd be glad to help extend that work or contribute additional strategies as follow-ups.

On the loop: you're right, a heap keyed by `(used, pickedFrom)` brings it to `O(n log m)`. I drafted it on top of this branch in #1826 (opened as draft, not asking for review yet). Happy to keep it as a separate follow-up after this lands, or fold it in here if you'd prefer.

wkd-woo:
Author of #1621 here :) nice catch — I had not noticed the tie case falling through to sort order.

Our PRs touch the same loop in `allocate.go`, so whichever lands second will need a rebase. Happy to handle that on my side if yours goes in first. Your tie-break should slot into the refactored shape of #1621 pretty naturally too: `replicaCount` can carry a per-allocation picked count, and the distributed comparator just falls back to it on ties.

tariq1890:
@jonathan-meiri Can you rebase your PR?

rajatchopra (approved):
/lgtm

Lets attack the loop optimization in #1826 next. Thanks.

jonathan-meiri:
@tariq1890 done

tariq1890 (approved):

tariq1890:
@jonathan-meiri Can you address the failing `gofmt` check?

jonathan-meiri:
@tariq1890 done

tariq1890:
Thank you for your contributon @jonathan-meiri !

jonathan-meiri:
Thanks @tariq1890 and @rajatchopra for the careful review, and @wkd-woo for the coordination heads-up. Appreciate it!

PR #1793, opened 2026-05-20
Ship static busybox shell in k8s-device-plugin image

### LINKED ISSUE #1744: Ship static busybox shell in k8s-device-plugin image
Part of NVIDIA/cloud-native-team#299.

NVIDIA STIG policy is dropping `-dev` distroless tags as approved
parent images. To move the k8s-device-plugin image onto a non-`-dev`
base **without changing the entrypoint or any Kubernetes manifests**,
we ship a static busybox shell inside the image.

## Approach

Add a `debian:trixie-slim` shell-builder stage to
`deployments/container/Dockerfile` that apt-installs `busybox-static`.
COPY `/bin/busybox` into the final stage and run
`busybox --install -s /bin` to populate `/bin/sh`, `/bin/rm`,
`/bin/ln`, etc. Flip the final base from
`nvcr.io/nvidia/distroless/go:<ver>-dev` to the matching non-`-dev`
variant. Drop the existing `SHELL ["/busybox/sh", "-c"]` directive
and the `RUN ln -s /busybox/sh /bin/sh` symlink — the static busybox
install replaces them.

```dockerfile
FROM debian:trixie-slim AS shell-builder
RUN apt-get update \
 && apt-get install --no-install-recommends -y busybox-static \
 && rm -rf /var/lib/apt/lists/*

FROM nvcr.io/nvidia/distroless/go:<version>
COPY --from=shell-builder /bin/busybox /bin/busybox
RUN ["/bin/busybox", "--install", "-s", "/bin"]
```

The existing `nvidia-device-plugin` / GFD / MPS daemon entrypoints
and Helm manifests remain untouched.

Precedent:
[`kubernetes-sigs/dra-driver-nvidia-gpu`](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu)
uses the same pattern with `bash-static`.

## Tracking

No PR yet — separate companion PR will be opened against this repo.

## Acceptance

- [ ] Dockerfile no longer references a `-dev` distroless tag
- [ ] No `/busybox/sh` references remain; `/bin/sh` is provided by static busybox
- [ ] Image builds successfully on amd64 and arm64
- [ ] device-plugin / GFD / MPS daemon binaries run as before when
      consumed by gpu-operator

rajathagasthya:
Flip the base from `*-dev*` to non-`*-dev*` distroless and source a static busybox from `debian:trixie-slim`. The device-plugin entrypoint and any shell-based lifecycle hooks continue to work via `/bin/sh` and busybox applet symlinks layered into the final image.

Part of NVIDIA/cloud-native-team#299.
Resolves #1744.

### DIFF
--- a/deployments/container/Dockerfile
+++ b/deployments/container/Dockerfile
@@ -40,11 +40,24 @@ ARG VERSION="N/A"
 ARG GIT_COMMIT="unknown"
 RUN make PREFIX=/artifacts cmds
 
-FROM nvcr.io/nvidia/distroless/go:v4.0.6-dev AS application
+# Build a static busybox layout: one binary plus applet symlinks (sh, rm,
+# ln, sleep, cat, ...) so PATH-resolved commands in init-container wrappers
+# and lifecycle hooks keep working on the non-*-dev* distroless base.
+FROM debian:trixie-slim AS shell
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends busybox-static \
+ && rm -rf /var/lib/apt/lists/* \
+ && mkdir /busybox \
+ && cp /bin/busybox /busybox/busybox \
+ && /busybox/busybox --install -s /busybox
+
+FROM nvcr.io/nvidia/distroless/go:v4.0.6 AS application
 
 USER 0:0
-SHELL ["/busybox/sh", "-c"]
-RUN ln -s /busybox/sh /bin/sh
+
+COPY --from=shell /busybox /busybox
+RUN ["/busybox/ln", "-s", "/busybox/sh", "/bin/sh"]
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/busybox
 
 ENV NVIDIA_DISABLE_REQUIRE="true"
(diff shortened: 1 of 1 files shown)


deployments/container/Dockerfile
```diff
@@ -40,11 +40,24 @@ ARG VERSION="N/A"
...
+ && /busybox/busybox --install -s /busybox
+
+FROM nvcr.io/nvidia/distroless/go:v4.0.5 AS application
 
 USER 0:0
-SHELL ["/busybox/sh", "-c"]
-RUN ln -s /busybox/sh /bin/sh
+
+COPY --from=shell /busybox /busybox
+RUN ["/busybox/ln", "-s", "/busybox/sh", "/bin/sh"]
```
rajatchopra, line 59: It isn't quite a departure from '-dev' if we use busybox-static. Why not bash-static?
Is it because we use extensive shell utilities?

rajathagasthya: Yeah, the deployment manifests use shell utilities (`rm`, `cat`, etc.), which we don't get if use `bash-static`. 

The `-dev` images we use also ship busybox in the same `/busybox` layout, so it's not really a departure.

cdesiniotis (approved):

Issue #143, opened 2019-10-23
How to use specific NVIDIA GPU type(model) in pod yaml


estherxyz:
### 1. Issue or feature description
There are different nvidia gpu model install on k8s cluster(or single machine).
How could I specific different NVIDIA GPU type in k8s yaml?

Because different GPU type maybe use different cuda version. And docker image need to support same cuda version with GPU device.

### 3. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)

Common error checking:
 - [ ] The output of `nvidia-smi -a` on your host

```
==============NVSMI LOG==============

Timestamp                           : Wed Oct 23 13:55:31 2019
Driver Version                      : 435.21
CUDA Version                        : 10.1

Attached GPUs                       : 1
GPU 00000000:01:00.0
    Product Name                    : GeForce GTX 1080 Ti
    Product Brand                   : GeForce
    Display Mode                    : Disabled
    Display Active                  : Disabled
    Persistence Mode                : Disabled
    Accounting Mode                 : Disabled
    Accounting Mode Buffer Size     : 4000
    Driver Model
        Current                     : N/A
        Pending                     : N/A
    Serial Number                   : N/A
    GPU UUID                        : GPU-c93652fd-383f-3e3b-dd65-ce5b718eee3e
    Minor Number                    : 0
    VBIOS Version                   : 86.02.39.40.19
    MultiGPU Board                  : No
    Board ID                        : 0x100
    GPU Part Number                 : N/A
    Inforom Version
        Image Version               : G001.0000.01.04
        OEM Object                  : 1.1
        ECC Object                  : N/A
        Power Management Object     : N/A
    GPU Operation Mode
        Current                     : N/A
        Pending                     : N/A
    GPU Virtualization Mode
        Virtualization mode         : None
    IBMNPU
        Relaxed Ordering Mode       : N/A
    PCI
        Bus                         : 0x01
        Device                      : 0x00
        Domain                      : 0x0000
        Device Id                   : 0x1B0610DE
        Bus Id                      : 00000000:01:00.0
        Sub System Id               : 0x376B1458
        GPU Link Info
            PCIe Generation
                Max                 : 3
                Current             : 1
            Link Width
                Max                 : 16x
                Current             : 8x
        Bridge Chip
            Type                    : N/A
            Firmware                : N/A
        Replays Since Reset         : 0
        Replay Number Rollovers     : 0
[...]

PR #1825, opened 2026-06-02
api/config/v1: keep user-specified rename/devices for shared resources

jonyhy96:
`disableResoureRenaming` currently resets the `Rename` and `Devices` fields of every entry under `sharing.timeSlicing.resources` (and `sharing.mps.resources`) to their defaults. That makes it impossible to:

  * pin a shared resource to a specific subset of GPU UUIDs (per-UUID time-slicing), and
  * expose two resource names on the same node, e.g. `nvidia.com/gpu` for full cards and `nvidia.com/gpu.shared` for the time-sliced subset.

Preserve both fields and emit a warning instead, so that user intent is honored. Downstream resource-manager code already supports per-UUID device lists and custom rename targets, so no additional plumbing is required.

### DIFF
--- a/api/config/v1/replicas.go
+++ b/api/config/v1/replicas.go
@@ -33,10 +33,37 @@ type ReplicatedResources struct {
 	Resources                  []ReplicatedResource `json:"resources,omitempty"                  yaml:"resources,omitempty"`
 }
 
+// mpsSharingID is the id value passed to disableResoureRenaming for the MPS
+// sharing path (see DisableResourceNamingInConfig).
+const mpsSharingID = "mps"
+
 func (rrs *ReplicatedResources) disableResoureRenaming(id string) {
 	if rrs == nil {
 		return
 	}
+
+	// The MPS control daemon can only apply a single active thread percentage
+	// per node, so per-GPU device selection and multiple rename targets are
+	// structurally unsupportable on the MPS path. Keep the historical
+	// behaviour: reset user-specified `rename` and `devices` fields.
+	if id == mpsSharingID {
+		rrs.resetRenameAndDevices(id)
+		return
+	}
+
+	// For time-slicing (and any future non-MPS sharing mode) there is no
+	// equivalent per-node constraint, so preserve user-specified `rename`
+	// and `devices` fields instead of resetting them. This enables per-UUID
+	// time-slicing, where only a subset of GPUs on a node is exposed under
+	// a renamed shared resource (e.g. nvidia.com/gpu.shared) while the
+	// remaining devices stay on the original resource name.
+	rrs.keepRenameAndDevices(id)
+}
+
+// resetRenameAndDevices restores the historical behaviour: clear any
+// user-specified `rename` field and force `Devices.All=true` on every entry,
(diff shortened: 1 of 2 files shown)


rajatchopra:
@jonyhy96 Can you provide a corresponding unit test case here? Possibly one that shows how two resource names are used within a node.

jonyhy96:
@rajatchopra Thanks for the review! I've pushed a unit test TestDisableResoureRenamingKeepsUserSpec in api/config/v1/replicas_test.go (commit 85954f6) covering four cases:

1. Per-UUID rename preserved when renameByDefault=false.

2. Per-UUID devices list preserved when renameByDefault=true with the default rename target.

3. Two resource names on a single node — exactly the scenario you asked about: nvidia.com/gpu for full cards and nvidia.com/gpu.shared carrying only a UUID-selected subset (devices: ["GPU-<uuid>"]). Any GPU on the node that isn't listed stays on the original nvidia.com/gpu resource name, so the two resource names coexist on the same node.

4. nil receiver no-op for defensive coverage.

All four sub-tests pass locally (go test ./api/config/v1/... is green). PTAL.

henry118:
Looking at the #490 that introduced the `disableResoureRenaming` logic, resetting `Rename` and `Devices` fields seems intentional to me. Particularly [this comment](https://github.com/NVIDIA/k8s-device-plugin/pull/490#discussion_r1470981431) explained the limitation of the MPS implementation.

IIUC MPS requires all GPUs on a node to share the same replica count. This is enforced by `disableResoureRenaming` which forces `Devices.All = true` (preventing per-GPU device selection) and clears custom rename fields (preventing multiple resource names).

jonyhy96:
@henry118  Good catch — thanks for flagging the MPS constraint. You're right that MPS and time-slicing share this helper but have very different invariants: the MPS control daemon can only apply one active_thread_percentage per node, so per-GPU device selection and multiple rename targets aren't representable on the MPS path, whereas time-slicing has no equivalent per-node limit.

I've narrowed the change so it only affects the time-slicing branch:

- disableResoureRenaming("mps") still clears user-specified Rename and forces Devices.All=true, with the original warning wording — the MPS invariant is preserved byte-for-byte.

- Only disableResoureRenaming("timeSlicing") (and any future non-MPS sharing modes) now keeps user-specified rename / devices, so per-UUID time-slicing on a subset of GPUs (e.g. some GPUs on the node stay on nvidia.com/gpu while others are sliced into nvidia.com/gpu.shared) works end-to-end.

- Added TestDisableResoureRenamingMPSStillResets as a regression test that pins the MPS reset behaviour so this can't drift in the future.

Force-pushed as f30be88de. PTAL.

henry118:
@jonyhy96 Thanks for bearing with me on this one. At this stage, we can't take this patch, since we don't think custom rename for shared resources is ready enough for prod use cases broadly. And honestly our stance on this hasn't changed since we reverted it before. see https://github.com/NVIDIA/k8s-device-plugin/issues/143#issuecomment-1325017300.

I'm going to close this PR. Happy to talk through the underlying use case if there's a different approach, but this one we can't take right now.

PR #1826, opened 2026-06-02
perf(rm): replace per-iteration sort in distributedAlloc with a min-heap

jonathan-meiri:
## Summary

Follow-up optimization stacked on #1788. Until #1788 lands, this PR's cumulative diff includes that commit too — the new work here is in the **second commit** (the heap refactor). **Not for review until #1788 lands.**

Replaces the per-iteration `sort.Slice` in `distributedAlloc` with a min-heap keyed by `(used, pickedFrom)`. Brings the loop from `O(n² log n)` to `O(n log m)` where `n` is replicas requested and `m` is the number of physical GPUs touched in this allocation. Same correctness as #1788; same tests pass.

Practically, `n` and `m` are small in real configurations and the wall-clock impact is invisible — this is structural cleanliness, not a hot-path speedup.

Opened as **draft** so it doesn't enter the review queue alongside #1788. Happy to mark it ready as a separate follow-up after #1788 lands, or fold the change into #1788 if that's preferred.

### DIFF
--- a/internal/rm/allocate.go
+++ b/internal/rm/allocate.go
@@ -17,8 +17,8 @@
 package rm
 
 import (
+	"container/heap"
 	"fmt"
-	"sort"
 
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 )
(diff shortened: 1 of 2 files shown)


jonathan-meiri:
Hi @rajatchopra — following up on your comment in #1788 ("Lets attack the loop optimization in #1826 next"). This PR is out of draft and ready for review: a single commit with the heap-based `O(n log m)` refactor of `distributedAlloc`. Would appreciate your thoughts when you have a moment.

rajatchopra:
Heap logic looks good. Though this PR must wait for #1621 first. Will need a rebase after that.

jonathan-meiri:
Rebased on top of #1621 (now merged). The heap now sits on top of the new `greedyAlloc` / `replicaComparator` structure — both `distributed` and `packed` policies get the O(n log m) improvement. Existing tests pass unchanged.

Ready for another look when you have a moment, @rajatchopra.

henry118:
@jonathan-meiri Do you have the benchmark data that shows the improvement?

jonathan-meiri:
@henry118 The improvement is real, added `BenchmarkGreedyAlloc` (3be20fb),
same signature so it runs on `main` too for a direct before/after:
- n=64 (8 GPUs ×8):   137µs → 18µs  (7.7×)
- n=128 (8 ×16):      518µs → 36µs  (14×)

plus ~10× fewer allocations. To be clear on scope: this isn't a hot path (runs once per pod admission), so it's not a latency fix. The change came out of @rajatchopra's [suggestion](https://github.com/NVIDIA/k8s-device-plugin/pull/1788#issuecomment-4596838324) on #1788 to do one sort instead of N

henry118:
Thanks @jonathan-meiri, numbers look legit.

Allocs dropping is good. Raw us savings won't be noticeable in practice, but the alloc/GC reduction is worth it on its own.

Please rebase and sign off the last commit.

jonathan-meiri:
@henry118  Done, rebased on latest main and signed off the benchmark commit (DCO green).
Thanks for the review!

henry118 (approved):

jonathan-meiri:
Thanks for the review and merge, @henry118 🙏

FYI the follow-up #1936 (a `spread` allocation policy) builds directly on this greedyAlloc/comparator structure, I just took it out of draft now that this PR landed. No rush at all, just flagging the connection while it's fresh.

PR #1836, opened 2026-06-08
Add configurable --sysfs-root for NUMA detection and PCI discovery

### LINKED ISSUE #1833: [Feature]: Add a configurable `--sysfs-root` for device NUMA detection (`GetNumaNode`)
## Summary

`GetNumaNode` reads a device's NUMA node from a hardcoded `/sys/bus/pci/devices/<busID>/numa_node`, with no way to redirect the sysfs root. 
This is the device-plugin counterpart of [NVIDIA/k8s-test-infra#264](https://github.com/NVIDIA/k8s-test-infra/issues/264) (same `--sysfs-root` request for the DRA driver); both consume the synthetic tree from [#263](https://github.com/NVIDIA/k8s-test-infra/issues/263) (`render-pci-sysfs`).

## Problem

`internal/rm/nvml_devices.go:150-162`:

```go
busID := strings.ToLower(strings.TrimPrefix(uint8Slice(info.BusId[:]).String(), "0000"))
b, err := os.ReadFile(fmt.Sprintf("/sys/bus/pci/devices/%s/numa_node", busID))
if err != nil {
    return false, 0, nil  // error swallowed -> device reported with no NUMA
}
```

The `/sys` prefix is a literal; nothing overrides it (`--nvidia-driver-root` / `--nvidia-dev-root` feed only CDI / library resolution, not this read). When NVML reports devices whose bus IDs are not backed by host sysfs (simulated/mock GPUs), the read fails, `hasNuma=false`, and `BuildDevice` (`internal/rm/devices.go:93-101`) leaves `Device.Topology` nil, so the device is advertised with no NUMA and the Topology Manager / podresources see no affinity. 

## Proposed solution

Flag `--sysfs-root` (default `/sys`), matching #264 and the existing `--nvidia-*-root` family:

```go
os.ReadFile(filepath.Join(sysfsRoot, "bus/pci/devices", busID, "numa_node"))
```

In-tree precedent for the same read: go-nvlib's `WithPCIDevicesRoot` (`pkg/nvpci/nvpci.go:187,317`, mock at `mock.go:104`) and k8s `dynamic-resource-allocation`'s `WithFSFromRoot` (which #264 builds on). Default unchanged means zero impact on real deployments.

## Alternatives

- **`nvmlDeviceGetNumaNodeId`:** doesn't help. A mock NVML returns `NVML_ERROR_NOT_SUPPORTED` for it, and on real hardware it signals "no NUMA" via return code (not the sysfs `-1`) and is absent on older drivers, so it would need return-code gating plus a sysfs fallback anyway.
- **Subpath-mount the tree over `/sys/bus/pci/devices`:** possible (#264's fallback) but shadows the real PCI dir per-pod; a flag is cleaner.

Scoped to `GetNumaNode` (NVML full GPU + MIG, which delegates to the parent). A second hardcoded root at `internal/vgpu/pciutil.go:42` (GFD vGPU detection) could adopt the same flag later. Happy to send a PR.

giuliocalzo:
## Summary

- Adds `--sysfs-root` / `$SYSFS_ROOT` (default `/sys`) as a shared flag for the device plugin and GFD, matching the pattern used by `--nvidia-driver-root`
- Threads the sysfs root into `GetNumaNode` for GPU and MIG device NUMA topology, and into GFD `pciutil` for vGPU PCI scanning
- Adds Helm values (`sysfsRoot`, `sysfsHostPath`) for injecting `SYSFS_ROOT` and optionally mounting a synthetic PCI sysfs tree

Closes #1833

## Motivation

Mock-GPU test environments (e.g. k8s-test-infra `render-pci-sysfs`) need the plugin to read NUMA affinity from a synthetic PCI sysfs tree rather than the hardcoded `/sys` path. Without this, `GetNumaNode` silently reports no NUMA and devices are advertised without topology.

## Test plan

- [x] `go test ./internal/rm/ ./internal/vgpu/ ./api/config/v1/`
- [x] `go build ./cmd/nvidia-device-plugin/ ./cmd/gpu-feature-discovery/`
- [x] Helm template renders `SYSFS_ROOT` and `sysfs-root` volume when `sysfsRoot` + `sysfsHostPath` are set
- [ ] k8s-test-infra integration with synthetic PCI sysfs tree (downstream)

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ deploying the plugin via `helm`.
 | `--mig-strategy`         | `$MIG_STRATEGY`         | `"none"`        |
 | `--fail-on-init-error`   | `$FAIL_ON_INIT_ERROR`   | `true`          |
 | `--nvidia-driver-root`   | `$NVIDIA_DRIVER_ROOT`   | `"/"`           |
+| `--sysfs-root`           | `$SYSFS_ROOT`           | `"/sys"`        |
 | `--pass-device-specs`    | `$PASS_DEVICE_SPECS`    | `false`         |
 | `--device-list-strategy` | `$DEVICE_LIST_STRATEGY` | `"envvar"`      |
 | `--device-id-strategy`   | `$DEVICE_ID_STRATEGY`   | `"uuid"`        |
@@ -234,6 +235,7 @@ flags:
   migStrategy: "none"
   failOnInitError: true
   nvidiaDriverRoot: "/"
+  sysfsRoot: "/sys"
   plugin:
     passDeviceSpecs: false
     deviceListStrategy: "envvar"
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -69,6 +69,11 @@ func NewConfig(c *cli.Context, flags []cli.Flag) (*Config, error) {
 		config.Flags.NvidiaDevRoot = config.Flags.NvidiaDriverRoot
 	}
 
+	if config.Flags.SysfsRoot == nil || *config.Flags.SysfsRoot == "" {
+		sysfsRoot := DefaultSysfsRoot
+		config.Flags.SysfsRoot = &sysfsRoot
+	}
+
 	// Preserve the historical MPS behavior unless the config explicitly relaxes it.
 	if config.Sharing.MPS != nil && config.Sharing.MPS.FailRequestsGreaterThanOne == nil {
 		t := true
--- a/api/config/v1/consts.go
+++ b/api/config/v1/consts.go
@@ -53,4 +53,5 @@ const (
 	DefaultCDIAnnotationPrefix = cdiapi.AnnotationPrefix
 	DefaultNvidiaCTKPath       = "/usr/bin/nvidia-ctk"
 	DefaultContainerDriverRoot = "/driver-root"
+	DefaultSysfsRoot           = "/sys"
 )
--- a/api/config/v1/flags.go
+++ b/api/config/v1/flags.go
@@ -62,6 +62,7 @@ type CommandLineFlags struct {
 	MpsRoot                 *string                 `json:"mpsRoot,omitempty"          yaml:"mpsRoot,omitempty"`
 	NvidiaDriverRoot        *string                 `json:"nvidiaDriverRoot,omitempty" yaml:"nvidiaDriverRoot,omitempty"`
 	NvidiaDevRoot           *string                 `json:"nvidiaDevRoot,omitempty"    yaml:"nvidiaDevRoot,omitempty"`
+	SysfsRoot               *string                 `json:"sysfsRoot,omitempty"        yaml:"sysfsRoot,omitempty"`
 	GDRCopyEnabled          *bool                   `json:"gdrcopyEnabled"             yaml:"gdrcopyEnabled"`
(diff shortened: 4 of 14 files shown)


giuliocalzo:
@tariq1890 @cdesiniotis Could you please review? This implements #1833 (--sysfs-root for NUMA detection and GFD PCI scanning).

tariq1890:
Thanks @giuliocalzo ! Is this primarily motivated to serve test frameworks? I was wondering if there are scenarios where a live environment would have a custom sysfs root.

giuliocalzo:
hi @tariq1890  — the primary motivation is test frameworks, specifically mock-GPU environments in k8s-test-infra that use synthetic PCI sysfs trees from `render-pci-sysfs` (see #1833 and NVIDIA/k8s-test-infra#264 for the DRA driver counterpart).

Today `GetNumaNode` reads a hardcoded `/sys/bus/pci/devices/<busID>/numa_node`. When NVML reports bus IDs that aren't backed by host sysfs (simulated GPUs), the read fails silently and devices are advertised without NUMA topology. The same applies to GFD vGPU PCI scanning in `internal/vgpu/pciutil.go`.

I don't expect live production clusters to set a custom sysfs root — the default stays `/sys`, so existing deployments are unchanged. The flag is there so CI/test infra can mount a rendered tree at an alternate path (e.g. `/rendered-sysfs`) without shadowing the real `/sys` mount per-pod.

Happy to trim scope or adjust naming if you'd prefer to keep this test-infra-only in documentation.

tariq1890:
Thank you for the context. In general, I'd like to avoid making significant changes to the code if it is to get the project to work with mocking and test frameworks.

Have you looked into `afero`? Linking the [docs](https://github.com/spf13/afero#quick-start-the-power-of-abstraction) here.

eliranw:
Hi, just chiming in that we also plan on using this for e2e testing for KAI-scheduler together with NVML-Mock for NUMA awareness, so this would be really helpful for us. It looks like the change is in line with the pattern used by --nvidia-driver-root, as @giuliocalzo wrote. Happy to talk through it more if that would help.

tariq1890:
> It looks like the change is in line with the pattern used by --nvidia-driver-root, as @giuliocalzo wrote.

Configurable nvidia driver root is needed to support the driver container use case, where as custom sysfs root is not motivated by a use-case that could exist in live environments. Modifying the k8s-device-plugin CLI interface to accommodate a test framework seems like an architectural smell to me.

dims:
@tariq1890 looking at https://www.kernel.org/doc/html/v4.19/admin-guide/sysfs-rules.html i see:

```
sysfs is always at /sys
Parsing /proc/mounts is a waste of time. Other mount points are a system configuration bug you should not try to solve. For test cases, possibly support a SYSFS_PATH environment variable to overwrite the application’s behavior, but never try to search for sysfs. Never try to mount it, if you are not an early boot script.
```

Could we strip down the changes in this PR to a minimal `SYSFS_PATH` env var as suggested there?

Here's an attempt with a few more suggestions:
https://github.com/dims/k8s-device-plugin/commit/a1dd26b425e5061f68c17eb474692d0ebfca1dec

including:
- using go-nvlib/nvpci
- call NVML's GetNumaNodeId

PR #1837, opened 2026-06-08
revert default enablement of features mofed, gdrcopy and mofed

### LINKED ISSUE #1692: [Question]: Enabling MOFED by default is a breaking change.
With the the release of 0.19.0 device plugin, MOFED is enabled by default #1550 
This is an unexpected behavior and is considered a breaking change, as MOFED will mount all "/dev/infiniband/uverbs<N>" devices, this is not just related to Mellanox IB, but as ibverbs is being used by other vendors (e.g: AWS EFA) this conflicts with other device plugins and DRA driver (DRANET) that are not expecting another device plugin to handle those devices.

Should this either be disabled by default, or mention clearly as a breaking change.

tariq1890:
This PR fixes #1692

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -104,19 +104,16 @@ func main() {
 		},
 		&cli.BoolFlag{
 			Name:    "gdrcopy-enabled",
-			Value:   true,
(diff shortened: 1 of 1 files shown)


cdesiniotis (approved):

rahulait (approved):
LGTM

PR #1930, opened 2026-07-17
fix: mark all devices sharing a physical GPU unhealthy on Xid errors

### LINKED ISSUE #1929: [Bug]: Health check misses devices that share a physical GPU
* OS/Version(e.g. RHEL8.6, Ubuntu22.04): Amazon Linux 2023
* Kernel Version: Kernel-6.1
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker): Containerd
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): EKS(v1.34)

* **Without MIG (time slicing or MPS)**. When multiple logical GPUs (replicas) correspond to one physical GPU UUID, an Xid error on that UUID should report all of them as unhealthy to the kubelet. Currently only one logical GPU is reported. **Since the physical GPU itself is broken, every logical GPU tied to it should become unhealthy**.
* **With MIG**. Even when multiple MIG instances inside one physical GPU are broken, at most one is reported as unhealthy, and in some cases none at all. If the value of `parentToDeviceMap[uuid]` happens to be a pointer to a device that is not broken, the GI and CI check rejects the event and **nothing is reported**.

tryuuu:
Fixes https://github.com/NVIDIA/k8s-device-plugin/issues/1929

`checkHealth` kept only one `*Device` per physical GPU UUID in `parentToDeviceMap`, so devices sharing a physical GPU overwrote each other. As a result, an Xid event marked at most one device as unhealthy.

With replication (time-slicing or MPS), the other replicas of the affected GPU remained schedulable. With MIG, an event could be dropped entirely if the single device retained in the map did not match the event's GI and CI.

This PR groups all logical devices by their parent GPU UUID and processes each device associated with the event UUID. Replicated devices are all reported as unhealthy, while MIG devices are filtered using each available event ID independently. If either the GI or CI is unavailable, the remaining available ID is still used to identify the affected devices.

### DIFF
--- a/internal/rm/health.go
+++ b/internal/rm/health.go
@@ -71,7 +71,7 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		_ = eventSet.Free()
 	}()
 
-	parentToDeviceMap := make(map[string]*Device)
+	parentToDeviceMap := make(map[string][]*Device)
 	deviceIDToGiMap := make(map[string]uint32)
 	deviceIDToCiMap := make(map[string]uint32)
 
@@ -85,7 +85,7 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		}
 		deviceIDToGiMap[d.ID] = gi
 		deviceIDToCiMap[d.ID] = ci
-		parentToDeviceMap[uuid] = d
+		parentToDeviceMap[uuid] = append(parentToDeviceMap[uuid], d)
 
 		gpu, ret := r.nvml.DeviceGetHandleByUUID(uuid)
 		if ret != nvml.SUCCESS {
@@ -151,23 +151,25 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 			continue
 		}
 
-		d, exists := parentToDeviceMap[eventUUID]
+		ds, exists := parentToDeviceMap[eventUUID]
 		if !exists {
 			klog.Infof("Ignoring event for unexpected device: %v", eventUUID)
 			continue
 		}
 
-		if d.IsMigDevice() && e.GpuInstanceId != 0xFFFFFFFF && e.ComputeInstanceId != 0xFFFFFFFF {
-			gi := deviceIDToGiMap[d.ID]
-			ci := deviceIDToCiMap[d.ID]
-			if gi != e.GpuInstanceId || ci != e.ComputeInstanceId {
-				continue
+		for _, d := range ds {
+			if d.IsMigDevice() && e.GpuInstanceId != 0xFFFFFFFF && e.ComputeInstanceId != 0xFFFFFFFF {
+				gi := deviceIDToGiMap[d.ID]
+				ci := deviceIDToCiMap[d.ID]
+				if gi != e.GpuInstanceId || ci != e.ComputeInstanceId {
+					continue
+				}
+				klog.Infof("Event for mig device %v (gi=%v, ci=%v)", d.ID, gi, ci)
 			}
-			klog.Infof("Event for mig device %v (gi=%v, ci=%v)", d.ID, gi, ci)
-		}
 
-		klog.Infof("XidCriticalError: Xid=%d on Device=%s; marking device as unhealthy.", e.EventData, d.ID)
-		unhealthy <- d
+			klog.Infof("XidCriticalError: Xid=%d on Device=%s; marking device as unhealthy.", e.EventData, d.ID)
+			unhealthy <- d
+		}
 	}
 }
 
--- a/internal/rm/health_test.go
+++ b/internal/rm/health_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


tariq1890:
Thank you @tryuuu for your contribution. On an initial glance, this looks fine to me. Let me run some more tests against this branch and get back to you sometime next week

tryuuu:
@tariq1890 
Hi, just following up on this PR.
Have you had a chance to test this yet? Please let me know if there’s anything else I can help with!

internal/rm/health.go
```diff
@@ -151,23 +151,25 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
...
 			continue
 		}
 
-		if d.IsMigDevice() && e.GpuInstanceId != 0xFFFFFFFF && e.ComputeInstanceId != 0xFFFFFFFF {
-			gi := deviceIDToGiMap[d.ID]
-			ci := deviceIDToCiMap[d.ID]
-			if gi != e.GpuInstanceId || ci != e.ComputeInstanceId {
-				continue
+		for _, d := range ds {
+			if d.IsMigDevice() && e.GpuInstanceId != 0xFFFFFFFF && e.ComputeInstanceId != 0xFFFFFFFF {
```
kvalliyurnatt, line 161: NVML I believe reports GPU instance Id and Compute Instance Id independently. I believe we can have an Xid event like 
GI=0, CI=FFFFFFFF
GI=1, CI=0

In this case GI 1 should remain healthy, but currently we end up skipping the entire GPU and mark all instances of that GPU unhealthy, is that intended ?.  I think we should individually verify the GIs and CIs
```
for _, d := range ds {
	if d.IsMigDevice() {
		gi := deviceIDToGiMap[d.ID]
		ci := deviceIDToCiMap[d.ID]
		if e.GpuInstanceId != 0xFFFFFFFF && gi != e.GpuInstanceId {
			continue
		}
		if e.ComputeInstanceId != 0xFFFFFFFF && ci != e.ComputeInstanceId {
			continue
		}
	}
	klog.Infof("XidCriticalError: Xid=%d on Device=%s; marking device as unhealthy.", e.EventData, d.ID)
	unhealthy <- d
}
```

We should also add a unit test for this case.

aryangorwade:
@tryuuu About the testing changes: this small map fan-out fix adds five NVML fakes and ~266 lines. Existing tests isolate deterministic data transformations instead of simulating raw NVML (eg. testing `DeviceMap`'s `setEntry`/insert methods). Could we extract and directly test the parent-UUID-to-logical-devices grouping, instead of testing the whole NVML event loop end to end?

I'm thinking we could create a type in `health.go` like this: 
```go
type placedDevice struct {
	parentUUID string
	device     *Device
}

func groupByParent(devices []placedDevice) map[string][]*Device {
	grouped := make(map[string][]*Device)
	for _, d := range devices {
		grouped[d.parentUUID] = append(grouped[d.parentUUID], d.device)
	}
	return grouped
}
```
You would call `placedDevices := make([]placedDevice, 0, len(devices))`, `placeDevices.append`, and then ultimately set `parentToDeviceMap := groupByParent(placedDevices)`.

This allows for easy unit testing:  
```go
func TestGroupByParent(t *testing.T) {
	parentA := "GPU-A"
	parentB := "GPU-B"

	deviceA0 := &Device{Device: pluginapi.Device{ID: "GPU-A::0"}}
	deviceA1 := &Device{Device: pluginapi.Device{ID: "GPU-A::1"}}
	deviceB0 := &Device{Device: pluginapi.Device{ID: "GPU-B::0"}}

	grouped := groupByParent([]placedDevice{
		{parentUUID: parentA, device: deviceA0},
		{parentUUID: parentA, device: deviceA1},
		{parentUUID: parentB, device: deviceB0},
	})

	require.Equal(t, []*Device{deviceA0, deviceA1}, grouped[parentA])
	require.Equal(t, []*Device{deviceB0}, grouped[parentB])
}
```

internal/rm/health.go
aryangorwade, line 104: Duplicate gpu registration has been an issue before this PR as well (and still is). Just something to note for a future fix.

internal/rm/health.go
tryuuu: Good catch. I agree that the GI and CI IDs should be checked independently. 
I addressed this and added unit tests in commit https://github.com/NVIDIA/k8s-device-plugin/pull/1930/changes/5397a2b5388a78321f8b88518944862d577eb85b.

tryuuu:
@kvalliyurnatt @aryangorwade 
Thank you both for the helpful and insightful comments. I agreed with your suggestions and have updated the PR accordingly. Could you please take another look when you have a chance? 
Please feel free to let me know if any further changes are needed.

internal/rm/health.go
tryuuu: Once this PR is merged, we can use `parentToDeviceMap` to register events once per unique parent GPU UUID, addressing #2052. 
I’d be happy to open a follow-up PR for that as well.

kvalliyurnatt (approved):

aryangorwade (approved):

aryangorwade:
@tryuuu Can you rebase on top of main and squash your changes into one commit?

tryuuu:
@aryangorwade done

internal/rm/health.go
```diff
@@ -85,7 +108,7 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan any, devices Devices, unhe
 		}
 		deviceIDToGiMap[d.ID] = gi
 		deviceIDToCiMap[d.ID] = ci
-		parentToDeviceMap[uuid] = d
+		placedDevices = append(placedDevices, placedDevice{parentUUID: uuid, device: d})
```
tariq1890, line 107: why construct this slice as an intermediate step only to use it for building a map? Can't we build the final map directly here? I don't see the `placedDevices` used for anything else.

tryuuu:
@aryangorwade done

internal/rm/health.go
```diff
@@ -40,6 +40,29 @@ const (
...
+
+func matchesMigEvent(deviceGI, deviceCI, eventGI, eventCI uint32) bool {
+	if eventGI != 0xFFFFFFFF && deviceGI != eventGI {
+		return false
+	}
+	if eventCI != 0xFFFFFFFF && deviceCI != eventCI {
+		return false
+	}
+	return true
+}
```
tariq1890, line 60: This is more readable IMO. Instead of relying on a "fall through to true", we can flip it around to explicitly express the condition we want to evaluate to true.
```suggestion
func matchesMigEvent(deviceGI, deviceCI, eventGI, eventCI uint32) bool {
      giMatches := eventGI == nvml.GPU_INSTANCE_ID_ANY || deviceGI == eventGI
      ciMatches := eventCI == nvml.COMPUTE_INSTANCE_ID_ANY || deviceCI == eventCI
      return giMatches && ciMatches
}
```

tariq1890 (changes requested):

internal/rm/health.go
tryuuu: Agreed, I think this is a good change. I've updated it. https://github.com/NVIDIA/k8s-device-plugin/pull/1930/changes/f5880f8e0200ffd7904e36f90bdd5b3430e4476b

internal/rm/health.go
tryuuu: I used `placedDevices` so the parent-UUID grouping can be tested directly, as suggested in the earlier review. I agree the slice is only used to build the map. I thought keeping this small unit test might be preferable to bringing back the larger NVML fakes. Do you see a way to build the map directly while keeping the grouping easy to unit test?

tariq1890: Understood, I think it's fine. This was a minor comment. Let's keep it as-is. Resolving this

tariq1890:
Thanks for your patience @tryuuu . The changes LGTM. Only two minor things left:

- Squash commit history
- Ensure your commit is signed with a GPG key (or similar). See [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

tariq1890 (approved):

Issue #1683, opened 2026-04-07
[Bug]: MIG: error getting parent memory info: Insufficient Permissions


dylex:
### 1. Quick Debug Information
* OS/Version: Rocky 8.10
* Kernel Version: 4.18.0-553.111.1.el8_10.x86_64
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker): cri-docker with nvidia-container-runtime
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): k8s 1.34 (also fails on 1.32)

### 2. Issue or feature description
With MIG enabled on A100 and MIG_STRATEGY=single:

```
E0407 94 main.go:188] error starting plugins: error getting plugins: unable to create plugins: failed to construct resource managers: error building device map: error building device map from config.resources: error building MIG device map: error visiting devices: error visiting device: error visiting device: error visiting MIG device: error visiting MIG device: error getting MIG profile for MIG device at index '(0, 0)': error getting parent memory info: Insufficient Permissions
```

Commit e3323ce5530fed8e6a9a835a79adb1599c41a643 seems relevant, as `nvidia-smi` inside the container also fails to get memory info (`N/A`) though it works on the host. When running without MIG, the warning is logged, but with MIG this is an error.

This used to work with the exact same config and hardware with some earlier combination of k8s-device-plugin/cuda/docker/kernel but I haven't been able to find a working one now.

dylex:
Upon downgrading nvidia-container-toolkit to version 1.17.8 this is working again (and memory also shows up in nvidia-smi).  It fails with 1.18.2 and 1.19.0, so perhaps this is a bug there instead.  It looks like this issue was reported in https://github.com/NVIDIA/nvidia-container-toolkit/issues/842 but not acknowledged.

dylex:
Also https://github.com/NVIDIA/nvidia-container-toolkit/issues/1740

cdesiniotis:
@dylex as a workaround, can you try generating a management CDI spec and requesting the management device explicitly in the device-plugin pod as suggested in https://github.com/NVIDIA/nvidia-container-toolkit/issues/1740#issuecomment-4081263793?

dylex:
Yes, that seems to work.

1. Upgraded back to nvidia-container-toolkit 1.19.0
[...]

PR #1935, opened 2026-07-20
Make visible devices configurable

sstrk:
As adjusting this env var is necessary for the current suggested workaround for issues [1] and [2].
This saves overhead for operators who previously had to adjust the env var via a kustomization or helm post-renderer.

  [1]: https://github.com/NVIDIA/k8s-device-plugin/issues/1683
  [2]: https://github.com/NVIDIA/nvidia-container-toolkit/issues/1740

### DIFF
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -200,7 +200,7 @@ spec:
             value: {{ .Values.deviceDiscoveryStrategy }}
         {{- end }}
           - name: NVIDIA_VISIBLE_DEVICES
-            value: all
+            value: {{ .Values.visibleDevices }}
           - name: NVIDIA_DRIVER_CAPABILITIES
             value: compute,utility
         securityContext:
--- a/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
+++ b/deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
@@ -199,10 +199,9 @@ spec:
           - name: DEVICE_DISCOVERY_STRATEGY
             value: {{ .Values.deviceDiscoveryStrategy }}
         {{- end }}
-          - name: NVIDIA_VISIBLE_DEVICES
-            value: all
-          - name: NVIDIA_DRIVER_CAPABILITIES
-            value: compute,utility
+        {{- with .Values.devicePlugin.containers.main.env }}
+        {{- toYaml . | nindent 10 }}
+        {{- end }}
         securityContext:
           {{- include "nvidia-device-plugin.securityContext" . | nindent 10 }}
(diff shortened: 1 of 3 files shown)


myeolenv:
Thanks for the contribution. We will review this PR.

cdesiniotis:
@sstrk my preference would be to not add this option to the top-level helm values, as those are typically reserved for common config shared by both k8s-device-plugin and GFD (which can both be deployed by the chart). I would be more open to introducing  a common field for passing arbitrary envvars to the device-plugin main container. For example:
```
devicePlugin:
  containers:
    init: {}
    main:
      env:
        - name: NVIDIA_VISIBLE_DEVICES
          value: "all"
    sidecar: {}
```

@tariq1890 @rahulait let me know if you have any strong opinions.

sstrk:
@cdesiniotis thanks for your reply and pointing out that top-level values are used for common configuration. I wasn't aware of that. I've updated the supplied changes based on your feedback. I'm happy to receive further feedback from you

kvalliyurnatt (approved):

tariq1890:
@cdesiniotis I completely agree with your suggestion!

deployments/helm/nvidia-device-plugin/templates/daemonset-device-plugin.yml
```diff
@@ -199,10 +199,9 @@ spec:
           - name: DEVICE_DISCOVERY_STRATEGY
             value: {{ .Values.deviceDiscoveryStrategy }}
         {{- end }}
-          - name: NVIDIA_VISIBLE_DEVICES
-            value: all
-          - name: NVIDIA_DRIVER_CAPABILITIES
-            value: compute,utility
```
tariq1890, line 205: I don't think we should remove this. I prefer introducing a new block that will populate the additional env vars.

Note that K8s takes the last occurrence of an env var as the value to be applied if a pod spec has duplicate env var entries.

deployments/helm/nvidia-device-plugin/values.yaml
```diff
@@ -106,6 +106,14 @@ devicePlugin:
   # This can be useful in non-overlay network environments (e.g., AWS, Azure, GCP)
   # to avoid unnecessary IP allocation for daemon pods.
   # enableHostNetwork: false
+  containers:
```
kvalliyurnatt, line 109: @sstrk can we actually introduce a schema that aligns with what we have in the [operator ](https://github.com/NVIDIA/gpu-operator/blob/42052c45a53f8c50e607b72ddc410ecfb9865060/deployments/gpu-operator/values.yaml#L273)

This could look something like
```
devicePlugin:
  extraEnvs:
  - name: NVIDIA_VISIBLE_DEVICES
    value: "all"
```

This would only apply to the main container, the above shape helps keep somewhat of a schema parity between the operator and the device plugin and also makes this more targeted to just the environment variables. what do you think ?

PR #1936, opened 2026-07-22
Add `spread` allocation policy for maximum per-pod distinct-GPU coverage

jonathan-meiri:
## Summary

- Add `spread` as a third value for `--shared-devices-allocation-policy`, alongside `distributed` (default) and `packed`.
- `spread` prefers physical GPUs the **current pod** has not yet picked from, maximizing distinct-GPU coverage for a single multi-slot allocation.
- Default behavior (`distributed`) is unchanged.

**Depends on #1826** (heap refactor) — the current diff includes that commit until #1826 lands, then collapses to just the `spread` policy addition. Opened as draft for now.

## Motivation

`distributed` minimizes **cluster-wide** load imbalance — it prefers the GPU with the fewest replicas already allocated across all pods. `packed` does the opposite for bin-packing. Neither guarantees that a single pod's multi-slot request touches distinct physical GPUs.

Consider a node where GPU-0 has 3 free replicas (5 allocated) and GPU-1 has 5 free (3 allocated). A pod requests 2:

```
┌─────────────────────────────────────────────────────────────────────┐
│              distributed (current default)                          │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  GPU 0                       GPU 1                                  │
│  ┌──────────┐               ┌──────────┐                            │
│  │ █████░░░ │               │ ███░░░░░ │      before                │
│  │  5 alloc │               │  3 alloc │                            │
│  └──────────┘               └──────────┘                            │
│                                                                     │
│  Pod requests 2 slots → distributed picks GPU with fewer allocated  │
│                                                                     │
│  ┌──────────┐               ┌──────────┐                            │
│  │ █████░░░ │               │ █████░░░ │      after                 │
│  │  5 alloc │               │  5 alloc │                            │
│  └──────────┘               └──────────┘                            │
│   ⚠ untouched                 ✓ 2 slots (pod concentrated)          │
│                                                                     │
│  → Pod runs entirely on GPU 1. No NVLink/NCCL across GPUs.          │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────┐
│              spread (new)                                           │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  GPU 0                       GPU 1                                  │
│  ┌──────────┐               ┌──────────┐                            │
│  │ █████░░░ │               │ ███░░░░░ │      before                │
│  │  5 alloc │               │  3 alloc │                            │
│  └──────────┘               └──────────┘                            │
│                                                                     │
│  Pod requests 2 slots → spread prefers GPUs untouched by THIS pod   │
│                                                                     │
│  ┌──────────┐               ┌──────────┐                            │
│  │ ██████░░ │               │ ████░░░░ │      after                 │
│  │  6 alloc │               │  4 alloc │                            │
│  └──────────┘               └──────────┘                            │
│   ★ 1 slot                    ★ 1 slot (pod spans both GPUs)        │
│                                                                     │
│  → Multi-GPU workloads (NCCL, DDP, tensor-parallel) get real        │
│    hardware diversity for a single request.                         │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘
```

Concrete workloads that benefit:
- **Data-parallel / distributed training** — each rank on a distinct physical GPU so ring-allreduce / NCCL spans real hardware
- **Fault-isolated inference** — a pod requesting 2 GPUs specifically to survive a single-card failure
- **NVLink topology plays** — a first step toward topology-aware allocation (weak preference: at least distinct cards)

@rajatchopra explicitly floated additional policies on #1788:
> "we can decide on distributed allocation or a packed one, or a topology driven one"

This adds the third policy that #1621's abstraction was built for.

## Design

Small refactor: the `replicaComparator` signature is enriched from `func(i, j *replicaCount) bool` to `func(i, j *gpuAllocState) bool`. Each policy now owns both its primary ordering and tie-break — the queue's `Less` is a pure passthrough:

```go
distributed: func(i, j *gpuAllocState) bool {
    if i.count.allocated() != j.count.allocated() {
        return i.count.allocated() < j.count.allocated()   // primary
    }
    return i.pickedFrom < j.pickedFrom                     // tie-break
}
packed: func(i, j *gpuAllocState) bool {
    if i.count.allocated() != j.count.allocated() {
        return i.count.allocated() > j.count.allocated()   // primary
    }
    return i.pickedFrom < j.pickedFrom                     // tie-break
}
spread: func(i, j *gpuAllocState) bool {
    if i.pickedFrom != j.pickedFrom {
        return i.pickedFrom < j.pickedFrom                 // primary
    }
    return i.count.allocated() < j.count.allocated()       // tie-break
}
```

`distributed` and `packed` are behavior-preserving — the primary+tie-break they used to get from `greedyAlloc`'s wrapping is now spelled out in the comparator body itself. All existing tests (`TestDistributedAlloc`, `TestPackedAlloc`, `TestPackedVsDistributedContrast`) pass unchanged.

## Usage

```bash
--shared-devices-allocation-policy=spread
# or
SHARED_DEVICES_ALLOCATION_POLICY=spread
```

Config file:
```yaml
version: v1
flags:
  plugin:
    sharedDevicesAllocationPolicy: spread
```

Commits are DCO-signed.

### DIFF
--- a/cmd/nvidia-device-plugin/main.go
+++ b/cmd/nvidia-device-plugin/main.go
@@ -219,6 +219,7 @@ func validateFlags(infolib nvinfo.Interface, config *spec.Config) error {
 		switch *config.Flags.Plugin.SharedDevicesAllocationPolicy {
 		case spec.AllocationPolicyDistributed:
 		case spec.AllocationPolicyPacked:
+		case spec.AllocationPolicySpread:
 		default:
 			return fmt.Errorf("invalid --shared-devices-allocation-policy option: %s", *config.Flags.Plugin.SharedDevicesAllocationPolicy)
 		}
--- a/internal/rm/allocate.go
+++ b/internal/rm/allocate.go
@@ -34,24 +34,57 @@ func (rc *replicaCount) allocated() int {
 	return rc.total - rc.available
 }
 
+// gpuAllocState is the per-physical-GPU bookkeeping the greedy allocator
+// tracks while it consumes candidates. A comparator sees both the shared
+// cluster-wide replicaCount and the per-allocation pickedFrom counter so it
+// can express policies that mix the two axes (e.g. spread, which orders
+// primarily by pickedFrom and only tie-breaks by allocated()).
+type gpuAllocState struct {
+	count      *replicaCount // shared reference to this GPU's replicaCount
+	pickedFrom int           // slots picked from this GPU during this allocation
+	replicas   []string      // remaining annotated-ID candidates for this GPU
+}
+
 // replicaComparator decides whether the physical GPU represented by i should
 // be preferred over the one represented by j when greedily selecting the next
-// device to allocate.
-type replicaComparator func(i, j *replicaCount) bool
+// device to allocate. Comparators are complete Less functions — they own both
+// the primary ordering and the tie-break — so a new policy can freely choose
+// which axis of gpuAllocState matters most.
+type replicaComparator func(i, j *gpuAllocState) bool
 
 // allocationComparators maps each allocation policy to the comparator that
 // implements it. All policies share the same greedy selection loop
 // (greedyAlloc) and differ only in how the next best candidate is chosen.
 var allocationComparators = map[string]replicaComparator{
 	// distributed prefers GPUs with the fewest allocated replicas to spread
-	// workload evenly across physical GPUs.
-	spec.AllocationPolicyDistributed: func(i, j *replicaCount) bool {
-		return i.allocated() < j.allocated()
+	// workload evenly across physical GPUs. Equal allocated counts fall back
+	// to pickedFrom so the pod's own picks also rotate.
+	spec.AllocationPolicyDistributed: func(i, j *gpuAllocState) bool {
+		if i.count.allocated() != j.count.allocated() {
+			return i.count.allocated() < j.count.allocated()
+		}
+		return i.pickedFrom < j.pickedFrom
 	},
 	// packed prefers GPUs with the most allocated replicas to consolidate
-	// workloads onto fewer physical GPUs.
-	spec.AllocationPolicyPacked: func(i, j *replicaCount) bool {
-		return i.allocated() > j.allocated()
+	// workloads onto fewer physical GPUs. Equal allocated counts still
+	// rotate by pickedFrom so a single pod that overflows a GPU spreads its
+	// overflow rather than concentrating arbitrarily.
+	spec.AllocationPolicyPacked: func(i, j *gpuAllocState) bool {
+		if i.count.allocated() != j.count.allocated() {
+			return i.count.allocated() > j.count.allocated()
+		}
+		return i.pickedFrom < j.pickedFrom
+	},
+	// spread maximizes the number of distinct physical GPUs the current
+	// allocation touches: it prefers GPUs the pod has not yet picked from,
+	// falling back to less-allocated when pickedFrom ties. Useful for
+	// multi-GPU workloads (e.g. distributed training / NCCL) that benefit
+	// from spanning physical hardware.
+	spec.AllocationPolicySpread: func(i, j *gpuAllocState) bool {
+		if i.pickedFrom != j.pickedFrom {
+			return i.pickedFrom < j.pickedFrom
+		}
+		return i.count.allocated() < j.count.allocated()
 	},
 }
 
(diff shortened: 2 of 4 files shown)


tariq1890 (dismissed):

tariq1890:
THanks @jonathan-meiri ! Can you rebase the PR?

cmd/nvidia-device-plugin/main.go
abrarshivani: Can we add `spread` here as well?

jonathan-meiri:
Rebased on latest main 🙏 
Thanks for the review, @tariq1890!

cmd/nvidia-device-plugin/main.go
jonathan-meiri: Good catch, done in a82bb9834. Updated the flag usage to [distributed | packed | spread]. Thanks!

internal/rm/allocate.go
```diff
@@ -37,21 +37,35 @@ func (rc *replicaCount) allocated() int {
...
+	spec.AllocationPolicyPacked: func(i, j *gpuAllocState) bool {
+		if i.count.allocated() != j.count.allocated() {
+			return i.count.allocated() > j.count.allocated()
+		}
+		return i.pickedFrom < j.pickedFrom
+	},
+	// spread prefers GPUs the current allocation has touched least, to span
+	// distinct physical GPUs.
+	spec.AllocationPolicySpread: func(i, j *gpuAllocState) bool {
+		if i.pickedFrom != j.pickedFrom {
```
abrarshivani, line 65: Looks like `pickedFrom` doesn't account for `required` / `MustIncludeDeviceIDs` here. If GPU0 already has a required replica, it still starts with `pickedFrom=0`, so `spread` may treat it as untouched and pick GPU0 again. Should we initialize `pickedFrom` from `required`?

jonathan-meiri: fixed in 046b76b by seeding `pickedFrom` from `required`, plus a test (TestSpreadAccountsForRequired) that stacks both slots on gpu0 without the fix. Thanks!

abrarshivani: Thanks @jonathan-meiri  for fixing this.

abrarshivani (approved):
@jonathan-meiri Thanks for working on this. LGTM.

internal/rm/allocate.go
```diff
@@ -37,21 +37,35 @@ func (rc *replicaCount) allocated() int {
...
+			return i.pickedFrom < j.pickedFrom
+		}
+		return i.count.allocated() < j.count.allocated()
```
tariq1890, line 69: What about MIG devices?

With MIG, `spread` would treat each MIG UUID as a distinct physical GPU, yes?. Candidate buckets use device UUIDs, while multiple MIG devices can share one parent. A multi-slot allocation can remain on one physical GPU.

internal/rm/allocate.go
```diff
@@ -151,6 +160,16 @@ func (r *resourceManager) greedyAlloc(available, required []string, size int, pr
...
+	// Seed pickedFrom from required so a GPU that already holds a required
+	// replica counts as touched by this allocation. Otherwise spread, which
+	// orders primarily by pickedFrom, could pick that same physical GPU again
+	// instead of spanning a new one.
+	for _, req := range required {
+		if item, ok := byGPU[AnnotatedID(req).GetID()]; ok {
+			item.pickedFrom++
+		}
+	}
+
```
tariq1890: If I am not wrong, this changes the tie-breaking logic and it would apply to the `distributed` and `packed` allocation policies. This goes beyond the scope of this PR - which is an introduction of a new allocation policy

internal/rm/allocate.go
jonathan-meiri: Good catch, you're right. Buckets key on the device UUID (GetID), and each MIG instance has its own UUID, so spread spans distinct MIG instances that can share one physical GPU. This is pre-existing (distributed/packed too).

I've kept this PR to the spread comparator and opened #2036 to bucket by parent GPU (from Device.Index) for all policies. 

Let me know if you'd prefer to fold that into this PR, or if you're happy keeping it separate.

internal/rm/allocate.go
jonathan-meiri: Good point, seeding pickedFrom leaked into distributed/packed. Fixed: pickedFrom stays pure, required accounting is now spread-only (touched()), with a test to lock it in. 

Packed's tie-break could improve too, but that's pre-existing. I'll follow up separately.

internal/rm/allocate.go
tariq1890: This godoc needs to be updated now that the `spread` comparator has been updated to use the `touched()` method as a primary sort key.

tariq1890 (approved):

tariq1890:
@jonathan-meiri Thanks for your hard work on this. These changes are  good to go. I can merge them after your PR branch is rebased.

jonathan-meiri:
Thanks @tariq1890  and @abrarshivani  for the thorough reviews and the merge, really appreciate it! 

The MIG parent-GPU bucketing follow-up we discussed is up as #2036 whenever you have a chance.

PR #1946, opened 2026-08-02
Wait for MPS daemon readiness before advertising shared resources

jonathan-meiri:
## Summary

The device plugin can advertise MPS-shared resources to kubelet before the MPS daemon has finished configuring per-device memory limits and thread percentages. A pod scheduled in that window runs against MPS with the daemon defaults (no pinned memory limit, 100% threads) instead of the configured limits — silently bypassing the intended isolation.

This gates readiness on the MPS daemon's `.ready` file, closing the two existing TODOs in `waitForDaemon`.

## The race

`internal/plugin/server.go:Start()` calls `plugin.mps.waitForDaemon()` before `Serve()` + `Register()`. Previously `waitForDaemon` ran a single `AssertHealthy()`:

```go
// TODO: Check the .ready file here.
// TODO: Have some retry strategy here.
if err := m.daemon.AssertHealthy(); err != nil { ... }
```

`AssertHealthy()` issues `get_default_active_thread_percentage`, which only proves the **control pipe is responsive**. In `Daemon.Start()` the ordering is:

1. `mpsControlBin -d` starts → **pipe becomes responsive** (`AssertHealthy` passes here)
2. per-device pinned memory limits applied
3. active thread percentage applied
4. `.ready` file created (after all daemons' `Start()` return)

If the single check lands between (1) and (4), the plugin registers as ready while the configured MPS limits are not yet in place.

The MPS control daemon already creates a node-global `.ready` file only after full initialization, but nothing consumed it — the two TODOs noted exactly this gap. Both components share the same `/mps` hostPath mount, so the file is visible to the device plugin.

## Changes

- `mps/root.go`: add `Root.ReadyFilePath()` so the marker path has a single definition.
- `mps-control-daemon/main.go`: create/remove `.ready` via `ReadyFilePath()` instead of the hardcoded `/mps/.ready`, keeping writer and reader in sync.
- `mps/daemon.go`: add `Daemon.Ready()`, reporting whether the `.ready` file exists.
- `internal/plugin/mps.go`: rewrite `waitForDaemon` to poll `checkDaemonReady` (`Ready()` **and** `AssertHealthy()`) every 5s up to a 5m bound. On timeout the caller fails and is retried by the plugin manager, so the bound is per-attempt.

## Test plan

- [x] `go test ./...` — full suite passes; new `TestReadyFilePath` and `TestDaemonReady` cover the readiness primitive (readiness reflects `.ready` file existence).
- [x] `make build`, `make check-modules`, gofmt — clean.

Commits are DCO-signed.

### DIFF
--- a/cmd/mps-control-daemon/main.go
+++ b/cmd/mps-control-daemon/main.go
@@ -213,7 +213,7 @@ func startDaemons(c *cli.Context, cfg *Config) ([]*mps.Daemon, bool, error) {
 			return mpsDaemons, true, nil
 		}
 	}
-	readyFile, err := os.Create("/mps/.ready")
+	readyFile, err := os.Create(mps.ContainerRoot.ReadyFilePath())
 	if err != nil {
 		return mpsDaemons, true, fmt.Errorf("failed to create .ready file")
 	}
@@ -223,7 +223,7 @@ func startDaemons(c *cli.Context, cfg *Config) ([]*mps.Daemon, bool, error) {
 }
 
 func stopDaemons(mpsDaemons ...*mps.Daemon) error {
-	if err := os.Remove("/mps/.ready"); err != nil {
+	if err := os.Remove(mps.ContainerRoot.ReadyFilePath()); err != nil {
 		klog.Warningf("Failed to remove .ready file: %v", err)
 	}
 	klog.Info("Stopping MPS daemons.")
--- a/cmd/mps-control-daemon/mps/daemon.go
+++ b/cmd/mps-control-daemon/mps/daemon.go
@@ -206,6 +206,16 @@ func (d *Daemon) AssertHealthy() error {
 	return err
 }
 
+// Ready returns true once the MPS daemons have signalled that initialization
+// has completed by creating the node-global .ready file under the MPS root.
+// AssertHealthy only proves the control pipe is responsive, which happens
+// before per-device memory limits and thread percentages are applied; Ready
+// gates on the full configuration being in place.
+func (d *Daemon) Ready() bool {
+	_, err := os.Stat(d.root.ReadyFilePath())
+	return err == nil
+}
+
 // EchoPipeToControl sends the specified command to the MPS control daemon.
 func (d *Daemon) EchoPipeToControl(command string) (string, error) {
 	var out bytes.Buffer
--- a/cmd/mps-control-daemon/mps/daemon_test.go
+++ b/cmd/mps-control-daemon/mps/daemon_test.go
@@ -1,0 +1,49 @@
+/**
+# Copyright 2026 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package mps
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReadyFilePath(t *testing.T) {
+	require.Equal(t, "/mps/.ready", ContainerRoot.ReadyFilePath())
+	require.Equal(t, "/custom/root/.ready", Root("/custom/root").ReadyFilePath())
+}
+
+func TestDaemonReady(t *testing.T) {
+	root := t.TempDir()
+	d := &Daemon{root: Root(root)}
+
+	// No .ready file yet: the daemon must not report ready. This is the window
+	// in which AssertHealthy can already succeed (control pipe responsive)
+	// while per-device memory/thread configuration is not yet applied.
+	require.False(t, d.Ready(), "daemon must not be ready before the .ready file exists")
+
+	// Once the MPS control daemon has finished initialization it creates the
+	// .ready file; the daemon must then report ready.
+	require.NoError(t, os.WriteFile(filepath.Join(root, ".ready"), nil, 0o644))
+	require.True(t, d.Ready(), "daemon must be ready once the .ready file exists")
+
+	// Removing the file (e.g. on daemon stop) flips readiness back to false.
+	require.NoError(t, os.Remove(filepath.Join(root, ".ready")))
+	require.False(t, d.Ready(), "daemon must not be ready after the .ready file is removed")
+}
--- a/cmd/mps-control-daemon/mps/root.go
+++ b/cmd/mps-control-daemon/mps/root.go
@@ -52,6 +52,16 @@ func (r Root) startedFile(resourceName spec.ResourceName) string {
 	return r.Path(string(resourceName), ".started")
 }
 
+// ReadyFilePath returns the path to the node-global .ready file. Unlike the
+// per-resource .started file, this single marker is created only after all MPS
+// daemons have completed initialization (compute mode, pinned memory limits,
+// and active thread percentage). Consumers such as the device plugin use it to
+// avoid advertising MPS-shared resources before the daemons are fully
+// configured.
+func (r Root) ReadyFilePath() string {
+	return r.Path(".ready")
+}
+
 // Path returns a path relative to the MPS root.
 func (r Root) Path(parts ...string) string {
 	pathparts := append([]string{string(r)}, parts...)
--- a/internal/plugin/mps.go
+++ b/internal/plugin/mps.go
@@ -19,6 +19,7 @@ package plugin
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"k8s.io/klog/v2"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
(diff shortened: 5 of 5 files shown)


myeolenv:
Thanks for the contribution. We will review this soon.

internal/plugin/mps.go
```diff
@@ -62,12 +72,36 @@ func (m *mpsOptions) waitForDaemon() error {
...
+	}
+}
+
+// checkDaemonReady reports whether the MPS daemon has finished initialization.
+// It requires both that the daemon has signalled readiness via its .ready file
+// (created only after compute mode, pinned memory limits, and thread
+// percentages are applied) and that the control pipe is responsive. Checking
+// only AssertHealthy is insufficient: the pipe becomes responsive before the
+// per-device configuration is in place, so a pod scheduled in that window
+// could run without the configured MPS memory/thread limits.
```
tariq1890: Please consider trimming this down. We don't need verbose comments. Here the number of lines in the comments exceed the lines of code in the method.

cmd/mps-control-daemon/mps/root.go
```diff
@@ -52,6 +52,16 @@ func (r Root) startedFile(resourceName spec.ResourceName) string {
 	return r.Path(string(resourceName), ".started")
 }
 
+// ReadyFilePath returns the path to the node-global .ready file. Unlike the
+// per-resource .started file, this single marker is created only after all MPS
+// daemons have completed initialization (compute mode, pinned memory limits,
+// and active thread percentage). Consumers such as the device plugin use it to
+// avoid advertising MPS-shared resources before the daemons are fully
+// configured.
```
tariq1890: This is too verbose.

cmd/mps-control-daemon/mps/daemon_test.go
```diff
@@ -0,0 +1,49 @@
...
+	require.False(t, d.Ready(), "daemon must not be ready before the .ready file exists")
+
+	// Once the MPS control daemon has finished initialization it creates the
+	// .ready file; the daemon must then report ready.
+	require.NoError(t, os.WriteFile(filepath.Join(root, ".ready"), nil, 0o644))
+	require.True(t, d.Ready(), "daemon must be ready once the .ready file exists")
+
+	// Removing the file (e.g. on daemon stop) flips readiness back to false.
+	require.NoError(t, os.Remove(filepath.Join(root, ".ready")))
+	require.False(t, d.Ready(), "daemon must not be ready after the .ready file is removed")
```
tariq1890: The comments here seem unnecessary.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -206,6 +206,16 @@ func (d *Daemon) AssertHealthy() error {
 	return err
 }
 
+// Ready returns true once the MPS daemons have signalled that initialization
+// has completed by creating the node-global .ready file under the MPS root.
+// AssertHealthy only proves the control pipe is responsive, which happens
+// before per-device memory limits and thread percentages are applied; Ready
+// gates on the full configuration being in place.
```
tariq1890: This is also very verbose. Consider keeping the comments to not more than 2 lines.

jonathan-meiri:
Trimmed all four, comments now ≤2 lines. Thanks @tariq1890!

internal/plugin/mps.go
```diff
@@ -62,12 +70,31 @@ func (m *mpsOptions) waitForDaemon() error {
...
+			klog.InfoS("MPS daemon is ready", "resource", m.resourceName)
+			return nil
+		}
+		if time.Now().After(deadline) {
+			return fmt.Errorf("timed out waiting for MPS daemon for %v: %w", m.resourceName, err)
+		}
+		klog.InfoS("Waiting for MPS daemon to be ready", "resource", m.resourceName, "reason", err)
+		time.Sleep(mpsReadyCheckInterval)
+	}
+}
```
tariq1890: Can you use the `wait.PollUntilContextTimeout` method for this? See [here](https://github.com/NVIDIA/k8s-device-plugin/blob/a8a935135bb05413d895ebd1aff96843f86c08fe/tests/e2e/e2e_test.go#L270) for an example

jonathan-meiri: Done, switched to wait.PollUntilContextTimeout. Thanks!

cmd/mps-control-daemon/mps/daemon_test.go
```diff
@@ -0,0 +1,43 @@
+/**
+# Copyright 2026 NVIDIA CORPORATION
```
tariq1890:
```suggestion
# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
```

internal/plugin/mps.go
```diff
@@ -62,12 +72,26 @@ func (m *mpsOptions) waitForDaemon() error {
 	if m == nil || !m.enabled {
 		return nil
 	}
-	// TODO: Check the .ready file here.
-	// TODO: Have some retry strategy here.
+
+	return wait.PollUntilContextTimeout(context.Background(), mpsReadyCheckInterval, mpsReadyCheckTimeout, true, func(context.Context) (bool, error) {
```
tariq1890: let's propagate a context from the caller instead  of intitialising a new one here.

internal/plugin/mps.go
tariq1890:
```suggestion
func (m *mpsOptions) waitForDaemon(ctx context.Context) error {
```

internal/plugin/mps.go
tariq1890: Thanks for the responsiveness!

cmd/mps-control-daemon/main.go
```diff
@@ -213,7 +213,7 @@ func startDaemons(c *cli.Context, cfg *Config) ([]*mps.Daemon, bool, error) {
 			return mpsDaemons, true, nil
 		}
 	}
-	readyFile, err := os.Create("/mps/.ready")
+	readyFile, err := os.Create(mps.ContainerRoot.ReadyFilePath())
```
abrarshivani, line 222: Can we remove `.ready` before starting the mps daemons? `/mps` is a hostPath, so the file can survive a container restart. A stale `.ready` can make the device plugin think mps is ready before the new daemon has finished applying the new config.

cmd/mps-control-daemon/mps/daemon.go
```diff
@@ -206,6 +206,13 @@ func (d *Daemon) AssertHealthy() error {
 	return err
 }
 
+// Ready reports whether the .ready file exists, i.e. the MPS daemons have
+// finished initialization (compute mode, memory limits, thread percentages).
+func (d *Daemon) Ready() bool {
+	_, err := os.Stat(d.root.ReadyFilePath())
+	return err == nil
```
abrarshivani: Can `Ready()` return an error as well? Right now any `os.Stat` error is treated as `not ready`, so a missing file looks the same as a permission or filesystem error.

cmd/mps-control-daemon/mps/daemon_test.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func TestDaemonReady(t *testing.T) {
+	root := t.TempDir()
+	d := &Daemon{root: Root(root)}
+
+	require.False(t, d.Ready(), "not ready before the .ready file exists")
+
+	require.NoError(t, os.WriteFile(filepath.Join(root, ".ready"), nil, 0o644))
+	require.True(t, d.Ready(), "ready once the .ready file exists")
```
abrarshivani: Can we add regression tests for this change? The current tests only check `.ready` file presence. They don't cover the race this PR fixes, such as the mps pipe being healthy before `.ready` is created.

internal/plugin/mps.go
jonathan-meiri: Done, copyright header fixed, and context is now propagated from the caller
(Start(ctx, ...) → waitForDaemon(ctx)). Thanks!

cmd/mps-control-daemon/mps/daemon_test.go
jonathan-meiri: All three addressed:
- Clear stale .ready at daemon startup (hostPath can survive restarts).
- Ready() now returns (bool, error) so real stat errors surface.
- Added regression tests: readiness is withheld until .ready exists (before the pipe is checked), plus a stat-error case.
Thanks for the careful review!

internal/plugin/mps.go
```diff
@@ -58,16 +68,34 @@ func (o *options) getMPSOptions(resourceManager rm.ResourceManager) (mpsOptions,
...
+		}
+		klog.InfoS("MPS daemon is ready", "resource", m.resourceName)
+		return true, nil
+	})
+}
+
+// checkDaemonReady requires the .ready file (written after full configuration)
+// and a responsive pipe; AssertHealthy alone responds before config is applied.
+func (m *mpsOptions) checkDaemonReady() error {
+	ready, err := m.daemon.Ready()
```
abrarshivani, line 89: One case I think is still possible: say the current config is A and it changes to B. The device plugin can observe B before the MPS daemon does. In that window, `.ready` may still be from A and the MPS pipe may still be healthy, so this check can succeed before MPS has applied B.

This should converge once the MPS daemon processes the update, so I don't think it needs to block this PR, but ideally readiness should also confirm that MPS is running the expected config

cmd/mps-control-daemon/mps/daemon_test.go
abrarshivani: Thanks, these address the earlier comments. I left one more comment about a possible config update race.

internal/plugin/mps.go
jonathan-meiri: Good catch, you're right this gates startup but not a config change. Opened #2055 to track it; I'll follow up there once this PR merges. 
Thanks!

PR #1950, opened 2026-08-07
Add THIRD_PARTY_NOTICES.md and a generator for it

abrarshivani:
Adds `THIRD_PARTY_NOTICES.md`, listing every Go dep with the verbatim text of
its license, and the generator that produces it.

- `make third-party-notices` regenerates it
- `make check-third-party-notices` regenerates and diffs it, and runs on every build

Covers the modules linked into the four `cmd/` binaries in the released image.

## What to review

96% of the diff is generated. Hand-written:

| File | Lines |
|---|---:|
| `hack/generate-third-party-notices.sh` | 370 |
| `.github/workflows/third-party-notices-check.yaml` | 40 |
| `Makefile` | 16 |
| `.github/workflows/basic-checks.yaml` | 7 |
| `.gitignore` | 1 |
| `deployments/devel/tools.go` | 1 |

Generated: `THIRD_PARTY_NOTICES.md` (10966 lines), `deployments/devel/go.mod`
and `go.sum` (164).

## How it works

1. Use `./bin/go-licenses` so the version pinned in `deployments/devel/go.mod`
   runs. `tools.go` gains the blank import so `make install-tools` and
   `make -C deployments/devel check-modules` keep that pin honest
2. Check the platform list against `deployments/container/multi-arch.mk`, fail
   if out of sync, so a new image target cannot silently produce a short file
3. Run `go-licenses save` and `csv` over `./cmd/...` once per released platform,
   from `vendor/`, then merge. It resolves only the host platform, build-tagged
   sources differ per target, and there is no union mode
4. Join all licenses per package instead of picking one
5. Add `module@version` from `vendor/modules.txt` by longest-prefix match
6. Write to a temp file, then move it into place

## Implementation notes

- **CGO stays on.** `make cmds` does not turn it off, and the released binaries
  link NVML through cgo. With `CGO_ENABLED=0`, build constraints exclude every
  file in `github.com/NVIDIA/go-nvml/pkg/dl`, and both `go-licenses save` and
  `csv` fail with `build constraints exclude all Go files` and exit 1, so the
  run stops rather than writing a short file. No C compiler is needed, since
  `go-licenses` only lists and parses.
- **Only the local module goes to `--ignore`.** It matches raw string prefixes,
  not path segments. Passing the stdlib names adds the token `go`, which
  silently drops 15 of the 74 entries, including gRPC, protobuf, all of
  `golang.org/x/*` and all of `gopkg.in/*`, with a zero exit. Measured both
  ways.
- **Licenses are joined, not picked.** `go-licenses` emits a row per recognised
  license, so keeping one would hide `filepath-securejoin`'s MPL-2.0 behind its
  BSD-3-Clause.
- **Rows carry `module@version`, not a URL.** In vendor mode `go-licenses` links
  into this repository at `HEAD`, which stops describing the released content
  once `main` advances and names this copy rather than upstream.
- **Output must match on every machine**, since CI diffs it. `LC_ALL=C` on every
  sort and grep. Whole-line `sort -u` instead of by key, which BSD and GNU sort
  tie-break differently. The awk counts instead of using `in`, which mawk and
  BSD awk disagree on. `grep -a` so a NUL byte cannot change the fence width.
- **An unidentified license fails the run.** `go-licenses` reports a license it
  cannot classify as `Unknown` and still exits 0, so without a guard the file
  would ship an entry that attributes nothing. Licenses are joined, so the check
  is anchored on both sides: it catches `BSD-3-Clause / Unknown` as well as a
  bare `Unknown`, and leaves one legitimately named `Unknown-Something` alone.
  An empty license field is caught too, since it would otherwise render as
  `Unknown` in the document.
- **Moved into place with `mv`, not `cp`.** `cp` truncates the target first, so
  an interrupted run would leave a half-written file.

## Scope

`./cmd/...` is what `make cmds` builds: `CMDS` is a wildcard over `./cmd/*`, and
the Dockerfile runs `make PREFIX=/artifacts cmds` then copies all four binaries
into the image. Against `./...` the only extra is the local
`internal/resource/testing` helper, which brings in no third-party dep.

Not covered:

- The build tools in `deployments/devel`. They are not redistributed, so they
  carry no attribution obligation.
- The distroless base image, covered by that image's own compliance process. The
  preamble links to its OSS listing.
- The static BusyBox binary and its applet symlinks, handled separately,
  including any source-distribution obligations it carries. The preamble records
  that it is GPLv2.
- The e2e suite under `tests/`, a separate module with its own `go.mod` and
  vendor tree, so its deps cannot reach this graph.

## Shipping

Committed to the repository. Not copied into the image, so image size is
unchanged, and not attached to releases, so the release flow is untouched.

`make check-third-party-notices` runs on every build, not behind a changed-paths
filter. The list is the import set of `./cmd/...`, so it goes stale when an
ordinary `.go` file changes its imports, not only when `go.mod` or `vendor/`
move. A filter keyed on manifests would pass green and the failure would land on
`main` instead of the PR that caused it.

## Testing

- Same sha256 on macOS/arm64 (BSD awk, BSD sort) and in `golang:1.26.5` (mawk
  1.3.4, GNU coreutils):
  `c401458af84d245cf45a0f438a9899a3e9e97f88c43316dbf683386bc9756d2a`. Two runs
  on one host also match.
- Completeness, derived without `go-licenses`: `go list -deps ./cmd/...` on
  `linux/amd64` and `linux/arm64` gives 513 non-stdlib, non-local packages,
  mapped by longest prefix to 74 license dirs in `vendor/`. Expected 74, present
  74, 0 missing, 0 extra.
- 74 index rows, 74 license-text sections, one per row and none orphaned.
- 0 `Unknown` licenses, 0 "License text unavailable", 0 unresolved
  `module@version`.
- The three multi-licensed entries (`filepath-securejoin`, `sigs.k8s.io/json`,
  `sigs.k8s.io/yaml`) match the files in `vendor/`.
- `make check-third-party-notices` passes on the committed file, and fails with
  a clear message when it is stale and when it is untracked. The
  `basic / third-party-notices` job passes here.
- Error paths: missing `multi-arch.mk` or `vendor/modules.txt`, a platform
  matrix out of sync, a filesystem `replace` directive, an unreadable module
  list. Each fails with a specific message and a non-zero exit.
- `make -C deployments/devel check-modules` passes.
- `shellcheck` clean on `hack/generate-third-party-notices.sh`. `actionlint` and
  `yamllint` report nothing on the new and changed workflows.


.github/workflows/third-party-notices-check.yaml
```diff
@@ -0,0 +1,49 @@
...
+# runtime inventory is the import closure of ./cmd/..., so it changes when
+# ordinary .go files change their imports, not only when go.mod or vendor/ move.
+# A filter keyed on dependency manifests would therefore pass green while the
+# committed file goes stale, and the failure would surface on main instead of on
+# the pull request that caused it.
+
+name: Third-Party Notices
+
+on:
+  workflow_call:
```
rahulait, line 19: Can we trigger this manually as well like we are doing for gpu-operator?

.github/workflows/basic-checks.yaml
```diff
@@ -35,3 +35,10 @@ jobs:
     uses: ./.github/workflows/code_scanning.yaml
     with:
       golang_version: ${{ needs.variables.outputs.golang_version }}
+
+  notices:
+    needs:
+    - variables
+    uses: ./.github/workflows/notices-check.yaml
```
tariq1890:
```suggestion
    uses: ./.github/workflows/third-party-notices-check.yaml
```

hack/create-release.sh
tariq1890, line 1: Let's not create a Release Asset for now. It's not clear why this needs to be its own Release Asset when GitHub already publishes the Source tar file by default. The `THIRD_PARTY_NOTICES.md` is already retrievable via Source tar.

Once we get more clarity on this and confirm that there is an actual need to create a separate Release Asset to host `THIRD_PARTY_NOTICES`, then we can create a follow-up PR

.gitignore
```diff
@@ -7,6 +7,8 @@ deployments/helm/gpu-feature-discovery
 cmd/gpu-feature-discovery/gfd-test-loop
 e2e_logs
 bin
+# Scratch tree written by hack/generate-notices.sh.
```
tariq1890:
```suggestion
```

Makefile
```diff
@@ -114,6 +114,30 @@ vendor: mod-vendor
...
+
+notices: bin/go-licenses
+	@bash hack/generate-notices.sh
+
+# Verify THIRD_PARTY_NOTICES.md is in sync with the dependency tree.
+notices-check:
+	@echo "- Checking if THIRD_PARTY_NOTICES.md is up to date..."
+	@# Cheap gate first, before spending minutes regenerating. git diff reports
+	@# nothing for an untracked path, so an uncommitted notices file would
+	@# otherwise pass this gate silently.
```
tariq1890:
```suggestion
```

Makefile
```diff
@@ -114,6 +114,30 @@ vendor: mod-vendor
 check-modules: | mod-tidy mod-verify mod-vendor
 	git diff --quiet HEAD -- $$(find . -name go.mod -o -name go.sum -o -name vendor)
 
+# THIRD_PARTY_NOTICES.md lists the license of every Go dependency, both the
+# modules linked into the released image and the build toolchain. go-licenses is
+# pinned in deployments/devel alongside the other build tools; install it into
+# ./bin rather than the shared GOBIN so the pinned version is the one that runs.
```
tariq1890:
```suggestion
```

Makefile
```diff
@@ -114,6 +114,30 @@ vendor: mod-vendor
...
+DEVEL_DIR := $(CURDIR)/deployments/devel
+
+bin/go-licenses: $(DEVEL_DIR)/go.mod $(DEVEL_DIR)/go.sum
+	GOBIN=$(CURDIR)/bin go -C $(DEVEL_DIR) install -mod=readonly github.com/google/go-licenses/v2
+
+notices: bin/go-licenses
+	@bash hack/generate-notices.sh
+
+# Verify THIRD_PARTY_NOTICES.md is in sync with the dependency tree.
```
tariq1890:
```suggestion
```

Makefile
```diff
@@ -114,6 +114,30 @@ vendor: mod-vendor
...
+notices-check:
+	@echo "- Checking if THIRD_PARTY_NOTICES.md is up to date..."
+	@# Cheap gate first, before spending minutes regenerating. git diff reports
+	@# nothing for an untracked path, so an uncommitted notices file would
+	@# otherwise pass this gate silently.
+	@git ls-files --error-unmatch THIRD_PARTY_NOTICES.md >/dev/null 2>&1 \
+		|| { echo "ERROR: THIRD_PARTY_NOTICES.md is not tracked. Run 'make notices' and commit the result."; exit 1; }
+	@$(MAKE) notices
```
tariq1890: shouldn't this be as a dependent make target instead?

Invoking make within a target is an anti-pattern

hack/create-release.sh
```diff
@@ -38,7 +38,17 @@ echo "Creating draft release"
 HELM_PACKAGE_VERSION=${VERSION#v}
 echo "Uploading release artifacts"
 
+# THIRD_PARTY_NOTICES.md is attached as a release asset rather than copied into
+# the image: the notices may be distributed alongside the artifact, and this
+# keeps the image size unchanged. It is committed and the notices job in CI
+# proves it matches the tree, so it is uploaded as-is from the tagged commit.
```
tariq1890:
```suggestion
```

tariq1890: As said in the earlier comment, let's not add it as a release asset for now

hack/generate-notices.sh
```diff
@@ -0,0 +1,534 @@
...
+
+    [[ -s "${INDEX_FILE}" ]] \
+        || die "go-licenses produced no entries for ${PACKAGES[*]} — refusing to write empty notices file."
+    [[ -s "${TOOLS_INDEX}" ]] \
+        || die "go-licenses produced no entries for the build toolchain — refusing to write incomplete notices file."
+
+    # A runtime row with no resolved module means modules.txt and the license
+    # tree disagree; that is a bug in this script, not a dependency change.
+    if cut -d, -f4 "${INDEX_FILE}" | grep -qx 'unknown'; then
+        die "some runtime packages could not be matched to a module in ${MODULES_TXT}." \
```
rahulait: Incomplete notices file can get shipped because of this. Should we fail loudly here?

.github/workflows/third-party-notices-check.yaml
abrarshivani: Good catch, we can. Unlike gpu-operator, this workflow currently gets the Go version from the calling workflow, so that value isn't available when someone runs it manually.
I'll add `workflow_dispatch` and have the workflow determine the Go version from `hack/golang-version.sh`, which is what `variables.yaml` already uses.

hack/generate-notices.sh
abrarshivani: Definitely, will update this.

.github/workflows/third-party-notices-check.yaml
tariq1890: Why do we need to manually trigger this?

.github/workflows/basic-checks.yaml
abrarshivani: Will update this

Makefile
abrarshivani: Yes this is anti-pattern. The only reason I had it inline was to run the cheap tracked-file check before the more expensive regeneration, but the dependency relationship is cleaner expressed directly in Make. I'll update it to `notices` a prerequisite of `notices-check` instead of invoking `$(MAKE)` from the recipe.

.github/workflows/notices-check.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Check out code
+        uses: actions/checkout@v7
+
+      - name: Install Go
+        uses: actions/setup-go@v7
+        with:
+          go-version: ${{ inputs.golang_version }}
+
+      - name: Verify THIRD_PARTY_NOTICES.md is up to date
+        run: make notices-check
```
tariq1890:
```suggestion
        run: make check-third-party-notices
```

hack/create-release.sh
abrarshivani: As discussed on slack, will update this.

hack/generate-notices.sh
```diff
@@ -0,0 +1,546 @@
...
+# verbatim text of each license, for ./cmd (from vendor/) and
+# deployments/devel/go.mod (from the module cache). Stdlib is excluded; the Go
+# distribution covers it.
+#
+# go-licenses resolves only the host platform, and build-tagged sources differ
+# per platform, so one run is both incomplete and host-dependent. There is no
+# union mode, so run it per target and merge, sorting with LC_ALL=C.
+#
+# Requires go (deployments/container/Dockerfile, via hack/golang-version.sh) and
+# go-licenses (deployments/devel/go.mod; 'make third-party-notices' installs it).
```
tariq1890:
```suggestion
```

hack/generate-notices.sh
tariq1890, line 1: rename to `generate-third-party-notices.sh`

hack/generate-notices.sh
```diff
@@ -0,0 +1,546 @@
...
+    while IFS= read -r pkg; do
+        [[ -n "${pkg}" ]] && TOOL_PKGS+=("${pkg}")
+    done < <(grep -E '^[[:space:]]*_ "' "${TOOLS_FILE}" | sed 's/.*"\(.*\)".*/\1/')
+
+    (( ${#TOOL_PKGS[@]} > 0 )) || die "no tool imports found in ${TOOLS_FILE}."
+}
+
+# Union into the shared tree: text is identical across platforms so overwrites
+# are no-ops. Module cache files are 0444 and cp preserves that, so restore
+# write permission or the next platform's copy fails.
```
tariq1890:
```suggestion
```

hack/generate-notices.sh
```diff
@@ -0,0 +1,546 @@
...
+    # tree disagree; that is a bug in this script, not a dependency change.
+    if cut -d, -f4 "${INDEX_FILE}" | LC_ALL=C grep -qx 'unknown'; then
+        die "some runtime packages could not be matched to a module in ${MODULES_TXT}." \
+            "Re-run 'make vendor' first; if it persists, fix annotate_modules in hack/generate-notices.sh."
+    fi
+
+    # Toolchain Source URLs come from go-licenses, which resolves non-github
+    # hosts over the network (GET <path>?go-get=1) and falls back to "Unknown"
+    # with a warning and a zero exit. Behind a proxy that blocks such a lookup
+    # the document would quietly lose URLs and look like a dependency change.
```
tariq1890:
```suggestion
```

.github/workflows/third-party-notices-check.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+# Regenerates THIRD_PARTY_NOTICES.md and fails if it differs from the committed
+# copy, so a dependency change cannot land without refreshed attribution.
+#
+# This runs on every build rather than behind a changed-paths filter. The
+# runtime inventory is the import closure of ./cmd/..., so it changes when
+# ordinary .go files change their imports, not only when go.mod or vendor/ move.
+# A filter keyed on dependency manifests would therefore pass green while the
+# committed file goes stale, and the failure would surface on main instead of on
+# the pull request that caused it.
```
tariq1890:
```suggestion
```

.github/workflows/third-party-notices-check.yaml
abrarshivani: I don't think we would need this. So, I will skip this for now.

tariq1890 (approved):

tariq1890:
The description still says this
> Covers two surfaces: the modules linked into the four cmd/ binaries in the
released image, and the build tools pinned in deployments/devel.

I am assuming this is just outdated?

abrarshivani:
@tariq1890 Yes this is outdated as following from deployments/devel are not shipped.
```
	github.com/google/go-licenses/v2 v2.0.1
	github.com/matryer/moq v0.7.1
```

PR #1955, opened 2026-08-13
[ci] don't specify version in third party licences

rahulait:
This change disables adding version to TPL file as each dependency update will require an update to TPL file with no change to license. We can avoid adding version to TPL file.


tariq1890:
Thanks a lot @rahulait !

tariq1890 (approved):

PR #1969, opened 2026-08-18
[ci] bump golang/x/text to pick CVE updates

rahulait:
Pick fix for CVE https://pkg.go.dev/vuln/GO-2026-5970


tariq1890:
Let's merge this PR as a last resort. Can we track down the direct dependency that needs updating and see if we can update `golang.org/x/text` transitively through that? This would be the preferred approach.

rahulait:
> Let's merge this PR as a last resort. Can we track down the direct dependency that needs updating and see if we can update `golang.org/x/text` transitively through that? This would be the preferred approach.

Its the node-feature-discovery which is bringing in v0.38.0 of x/text and there is no update available.

tariq1890 (approved):

PR #1982, opened 2026-08-25
fix: handle dangling symlink in config-manager updateSymlink

### LINKED ISSUE #364: Time-slicing config update: "Error: error creating symlink: file exists"

yoonhyunwoo:
Fixes #364
Follow-up to #365

dangling symlink is misjudged as "missing" and recreation fails.
os.Readlink checks the symlink itself, so it works correctly even when dangling.

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -398,29 +399,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
 		src = filepath.Join(f.ConfigFileSrcdir, config)
 	}
 
-	exists, err := fileExists(f.ConfigFileDst)
-	if err != nil {
-		return false, fmt.Errorf("error checking if file '%s' exists: %v", f.ConfigFileDst, err)
-	}
-	if exists {
-		srcRealpath, err := filepath.EvalSymlinks(src)
-		if err != nil {
-			return false, fmt.Errorf("error evaluating realpath of '%v': %v", src, err)
-		}
-
-		dstRealpath, err := filepath.EvalSymlinks(f.ConfigFileDst)
-		if err != nil {
-			return false, fmt.Errorf("error evaluating realpath of '%v': %v", f.ConfigFileDst, err)
-		}
-
-		if srcRealpath == dstRealpath {
+	current, err := os.Readlink(f.ConfigFileDst)
+	if err == nil {
+		if current == src {
 			return false, nil
 		}
+	} else if !os.IsNotExist(err) && !errors.Is(err, syscall.EINVAL) {
+		return false, fmt.Errorf("error reading symlink '%s': %v", f.ConfigFileDst, err)
+	}
 
-		err = os.Remove(f.ConfigFileDst)
-		if err != nil {
-			return false, fmt.Errorf("error removing existing config: %v", err)
-		}
+	err = os.Remove(f.ConfigFileDst)
+	if err != nil && !os.IsNotExist(err) {
+		return false, fmt.Errorf("error removing existing config: %v", err)
 	}
 
 	err = os.Symlink(src, f.ConfigFileDst)
@@ -461,17 +451,6 @@ func findPidToSignal(f *Flags) (int, error) {
 		}
 	}
 	return -1, fmt.Errorf("no process found")
-}
-
-func fileExists(filename string) (bool, error) {
-	info, err := os.Stat(filename)
-	if os.IsNotExist(err) {
-		return false, nil
-	}
-	if err != nil {
-		return false, err
-	}
-	return !info.IsDir(), nil
 }
 
 func getConfigFileNameMap(f *Flags) (map[string]bool, error) {
--- a/cmd/config-manager/main_test.go
+++ b/cmd/config-manager/main_test.go
@@ -1,0 +1,67 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package main
+
+import (
+	"os"
+	"path/filepath"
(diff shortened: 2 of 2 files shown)


yoonhyunwoo:
For context: we hit this in production. After a ConfigMap update, the `config-manager` sidecar of `gpu-feature-discovery` / `nvidia-device-plugin-mps-control-daemon` pods got stuck in CrashLoopBackOff, repeatedly logging:

```
Error: error creating symlink: symlink /available-configs/<config> /config/config.yaml: file exists
```

Retrying never recovers because the dangling symlink itself persists across restarts, so the pod needs manual intervention (delete the symlink or recreate the pod).

abrarshivani:
Thanks @yoonhyunwoo for fixing this. LGTM. Can we also add a regression test covering the dangling-symlink case?

abrarshivani:
@yoonhyunwoo Thanks for adding the tests. Can you please rebase the branch on the latest `main`, squash the commits, and make sure the resulting commit has a verified signature: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

yoonhyunwoo:
> @yoonhyunwoo Thanks for adding the tests. Can you please rebase the branch on the latest `main`, squash the commits, and make sure the resulting commit has a verified signature: [docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

Done!

abrarshivani (approved):

cmd/config-manager/main.go
```diff
@@ -398,29 +399,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
...
-		}
-
-		if srcRealpath == dstRealpath {
+	current, err := os.Readlink(f.ConfigFileDst)
+	if err == nil {
+		if current == src {
 			return false, nil
 		}
+	} else if !os.IsNotExist(err) && !errors.Is(err, syscall.EINVAL) {
+		return false, fmt.Errorf("error reading symlink '%s': %v", f.ConfigFileDst, err)
```
tariq1890, line 408: If you want to quote strings it's better to use `%q` instead of `%s`. In this case however, it's better to just unquote the string as the logs are emitted in JSON format.
```suggestion
		return false, fmt.Errorf("error reading symlink %s: %w", f.ConfigFileDst, err)
```

cmd/config-manager/main.go
```diff
@@ -398,29 +399,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
...
+	}
 
-		err = os.Remove(f.ConfigFileDst)
-		if err != nil {
-			return false, fmt.Errorf("error removing existing config: %v", err)
-		}
+	err = os.Remove(f.ConfigFileDst)
+	if err != nil && !os.IsNotExist(err) {
+		return false, fmt.Errorf("error removing existing config: %v", err)
```
tariq1890, line 413:
```suggestion
		return false, fmt.Errorf("error removing existing config: %w", err)
```

cmd/config-manager/main.go
```diff
@@ -398,29 +399,18 @@ func updateSymlink(config string, f *Flags) (bool, error) {
...
+	} else if !os.IsNotExist(err) && !errors.Is(err, syscall.EINVAL) {
+		return false, fmt.Errorf("error reading symlink '%s': %v", f.ConfigFileDst, err)
+	}
 
-		err = os.Remove(f.ConfigFileDst)
-		if err != nil {
-			return false, fmt.Errorf("error removing existing config: %v", err)
-		}
+	err = os.Remove(f.ConfigFileDst)
+	if err != nil && !os.IsNotExist(err) {
```
tariq1890, line 412: Isn't this redundant?
```
!os.IsNotExist(err)
```

We already check for an IsNotExist error earlier.

cmd/config-manager/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+		description string
+		config      string
+		wantChanged bool
+	}{
+		{
+			description: "create dangling symlink",
+			config:      "missing-config",
+			wantChanged: true,
+		},
+		{
```
tariq1890, line 49: This test is not clear to me. It seems like it depends on the previous test. If this is indeed the case, then it's an antipattern. The unit test should be independent of any other unit tests and it should pass when run in isolation.

PR #1994, opened 2026-08-31
validate vGPU capability record lengths

kvalliyurnatt:
### Summary
Fixes vGPU capability parsing so malformed PCI records cannot cause GPU Feature Discovery to hang or read outside the capability buffer.

### Details
A nonterminal vGPU capability record with a zero-length field caused Device.GetInfo() to repeatedly process the same record, resulting in an infinite loop and 100% CPU usage.

This change:

Uses an int cursor to prevent uint8 wraparound.
Requires nonterminal records to include at least their two-byte header.
Validates record boundaries before reading or advancing.
Validates the host-driver record and payload length.
Adds tests for zero-length, truncated, undersized, oversized, and valid records.

### DIFF
--- a/internal/vgpu/vgpu.go
+++ b/internal/vgpu/vgpu.go
@@ -40,7 +40,7 @@ type Info struct {
 
 const (
 	// VGPUCapabilityRecordStart indicates offset of beginning vGPU capability record
-	VGPUCapabilityRecordStart uint8 = 5
+	VGPUCapabilityRecordStart = 5
 	// HostDriverVersionLength indicates max length of driver version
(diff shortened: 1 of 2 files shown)


tariq1890:
LGTM. Thanks @kvalliyurnatt ! Please rebase this PR and I'll approve. This is good to merge.

kvalliyurnatt:
> LGTM. Thanks @kvalliyurnatt ! Please rebase this PR and I'll approve. This is good to merge.

@tariq1890 done

tariq1890 (approved):

PR #1995, opened 2026-09-01
fix: address review feedback on config-manager symlink handling

abrarshivani:
Follow-up to #1982, addressing review feedback from @tariq1890 that came in after the PR had already merged.

- Wrap errors with `%w` instead of `%v`, and drop the redundant quoting around the destination path since these logs are already emitted in JSON.
- Split `TestUpdateSymlinkDanglingDestination` into two independent subtests. Previously the second subtest relied on state left behind by the first, so it failed when run in isolation (e.g. via `-run`).

I left the `!os.IsNotExist(err)` check on the `os.Remove` call as is: it guards a different error value from a different syscall than the earlier `os.Readlink` check, so it isn't actually redundant, it's the standard idiom for tolerating a concurrent removal between the two calls.

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -405,17 +405,17 @@ func updateSymlink(config string, f *Flags) (bool, error) {
 			return false, nil
 		}
 	} else if !os.IsNotExist(err) && !errors.Is(err, syscall.EINVAL) {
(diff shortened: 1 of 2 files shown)


tariq1890 (approved):
Thanks @abrarshivani !

tariq1890:
Please squash the commit history as well

PR #1996, opened 2026-09-02
Go Modernize changes

kvalliyurnatt:
Ran go fix ./... which now includes the go modernize fixes, also added `modernize` to the lint rules

### DIFF
--- a/api/config/v1/config.go
+++ b/api/config/v1/config.go
@@ -33,10 +33,10 @@ const Version = "v1"
 // Config is a versioned struct used to hold configuration information.
 type Config struct {
 	Version   string    `json:"version"             yaml:"version"`
-	Flags     Flags     `json:"flags,omitempty"     yaml:"flags,omitempty"`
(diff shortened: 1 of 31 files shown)


api/config/v1/config.go
```diff
@@ -33,10 +33,10 @@ const Version = "v1"
 // Config is a versioned struct used to hold configuration information.
 type Config struct {
 	Version   string    `json:"version"             yaml:"version"`
-	Flags     Flags     `json:"flags,omitempty"     yaml:"flags,omitempty"`
```
kvalliyurnatt, line 36: `omitempty` does not really work on structs because structs are not really empty, if we want omit a struct we would need to add omitzero here, but that would be a change in behavior so left out adding that

henry118: seems like we just uncovered a divergence between json/yaml behavior. but it's not something this PR introduces. not a blocker imo.

henry118 (approved):

PR #2002, opened 2026-09-06
Reject an empty device list strategy

lenamonj:
## Description

An empty `deviceListStrategy` list in the config file is well-formed YAML, passes `NewDeviceListStrategies`, and `UpdateFromCLIFlags` leaves the non-nil pointer alone, so the `envvar` default never replaces it. `AllCDIEnabled()` is then vacuously true over the empty set, `Allocate` takes the all-CDI early return, and the container starts with no device list, no mounts, no CDI devices and no annotations: no GPU access, and no error anywhere.

`NewDeviceListStrategies` now rejects an empty set with an error naming the accepted strategies, which `validateFlags` surfaces as `invalid --device-list-strategy option`. `AllCDIEnabled()` returns false unless at least one CDI strategy is enabled, so a set with nothing enabled cannot take the all-CDI early return either; every set with at least one strategy enabled gets the same result as before.

## Checklist

## Testing

Three unit tests in `api/config/v1/strategy_test.go`: `NewDeviceListStrategies` over valid, unknown, empty and nil inputs, `AllCDIEnabled()` over a set with nothing enabled, and the config-file path, where `NewConfig` gives an omitted `deviceListStrategy` the `envvar` default and leaves an explicit empty list for `NewDeviceListStrategies` to reject. All three fail on main.


ssalagam:
Thanks for the contribution. We will review the PR soon.

api/config/v1/strategy_test.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+			require.Equal(t, tc.expectedAnyCDI, strategies.AnyCDIEnabled())
+			require.Equal(t, tc.expectedAllCDI, strategies.AllCDIEnabled())
+		})
+	}
+}
+
+// An empty list in a config file parses, and UpdateFromCLIFlags leaves the
+// non-nil pointer alone, so the "envvar" default never replaces it.
+func TestEmptyDeviceListStrategyFromConfigFileIsRejected(t *testing.T) {
```
abrarshivani: Can we cover the config default behavior here? In particular, it will be useful to verify that an omitted `deviceListStrategy` gets the `envvar` default, while an explicitly empty list does is rejected.

api/config/v1/strategy.go
```diff
@@ -25,13 +25,24 @@ import (
...
+
+// NewDeviceListStrategies constructs a new DeviceListStrategy from the
+// requested strategies. At least one strategy is required.
 func NewDeviceListStrategies(strategies []string) (DeviceListStrategies, error) {
-	ret := map[string]bool{
-		DeviceListStrategyEnvVar:         false,
-		DeviceListStrategyVolumeMounts:   false,
-		DeviceListStrategyCDIAnnotations: false,
-		DeviceListStrategyCDICRI:         false,
+	if len(strategies) == 0 {
```
abrarshivani, line 37: Can you also update `AllCDIEnabled()` to return `false` when no CDI strategy is enabled? This avoids the case where `AllCDIEnabled()` returns `true` when the strategy list is empty, which also contributed to this issue, and keeps the behavior consistent.

abrarshivani (commented):
@lenamonj Thanks for contributing this. The fix looks good overall. I left a few comments, PTAL.

lenamonj:
d72e5c4 covers both. AllCDIEnabled() now returns false unless at least one CDI strategy is enabled, so a set with nothing enabled no longer takes the all-CDI early return; every set with at least one strategy enabled keeps its old result. TestDeviceListStrategyFromConfigFile builds the config through NewConfig with the device-list-strategy flag at its default: an omitted deviceListStrategy comes back as envvar, an explicit empty list survives to NewDeviceListStrategies and is rejected there, and the earlier test that unmarshalled the config directly is folded into the table.

api/config/v1/strategy.go
```diff
@@ -58,8 +67,11 @@ func (s DeviceListStrategies) AnyCDIEnabled() bool {
 	return false
 }
 
-// AllCDIEnabled returns whether all strategies being used require CDI.
+// AllCDIEnabled returns whether at least one strategy is being used and all of them require CDI.
 func (s DeviceListStrategies) AllCDIEnabled() bool {
+	if !s.AnyCDIEnabled() {
+		return false
+	}
```
cdesiniotis: In my opinion, the below condition is more readable. 

```suggestion
	if len(s) == 0 {
		return false
	}
```

cdesiniotis (approved):

cdesiniotis:
@lenamonj thanks for the contribution. Can you squash your commits and make sure they are signed-off and cryptographically signed? Will trigger the GHAs after that. Thanks.

abrarshivani (approved):
LGTM

lenamonj:
Squashed into one commit, 864cddc, signed off and signed. Its tree is identical to d72e5c4, the head both approvals were given on.

tariq1890 (approved):

PR #2025, opened 2026-09-14
skip directories when resolving the nvml library path

### LINKED ISSUE #1945: vendored go-nvlib is affected by tryResolveLibrary directory-shadowing bug (NVIDIA/go-nvlib#107)
This repo vendors github.com/NVIDIA/go-nvlib v0.12.0 (vendor/github.com/NVIDIA/go-nvlib/pkg/nvlib/info/root.go), pulled in both directly and through github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi, and used from internal/rm, internal/plugin, internal/resource, and the plugin, gfd, and mps-control-daemon entrypoints.

That version has the bug from NVIDIA/go-nvlib#107: tryResolveLibrary resolves each candidate path with filepath.EvalSymlinks but never checks whether the result is a regular file. EvalSymlinks succeeds for directories too, so a stray directory named after a library in an earlier search path, for example /usr/lib64/libnvidia-ml.so.1, shadows the real library found later under /usr/lib/x86_64-linux-gnu. Driver library discovery here can end up pointing at a directory instead of the real .so.

The fix ("Skip non-regular files in tryResolveLibrary") is merged in NVIDIA/go-nvlib#108, but go-nvlib hasn't tagged a release past v0.12.0, so a normal go get -u won't pick it up yet.

Once go-nvlib cuts a release with #108 in it, or if a pseudo-version pin to the merge commit a9dc4f8 works as a stopgap, this repo's go.mod and vendor tree need the bump. nvidia-container-toolkit vendors the same broken copy directly (filed as NVIDIA/nvidia-container-toolkit#1966), and this repo pulls it in a second time through that module.

Closed issue: NVIDIA/go-nvlib#107
Merged fix: NVIDIA/go-nvlib#108
Same issue in nvidia-container-toolkit: NVIDIA/nvidia-container-toolkit#1966

I can open the dependency bump PR once a release or pseudo-version is agreed on.

rootkiller6788:
`root.tryResolveLibrary` in this package carries its own copy of the library-discovery code that go-nvlib has in `pkg/nvlib/info/root.go`, so it has the same bug that go-nvlib#107 reported there: it returns the first candidate that `filepath.EvalSymlinks` resolves, and EvalSymlinks resolves directories just as happily as it resolves files.

So a directory named `libnvidia-ml.so.1` under `/usr/lib64` (searched first) wins over the real library in `/usr/lib/x86_64-linux-gnu`, and the `dlopen` that follows fails with a not-found error. Not hypothetical — that is what was reported in kubernetes-sigs/dra-driver-nvidia-gpu#1272, which ships the same copy of this code.

The path matters here: the default `--container-driver-root` is `/driver-root`, not `/`, so this search loop is live, and `main.go:362` hands its result straight to `nvml.WithLibraryPath`. Get that wrong and the plugin cannot load NVML at all. The same function also feeds `nvinfo`'s `assertHasLibrary` when the platform is being resolved.

The change stats the resolved path and skips candidates that are not regular files, keeping the search going, mirroring NVIDIA/go-nvlib#108 upstream.

The test covers the shadowing case plus the ordinary lookup, a symlinked library (driver roots are mostly symlinks) and the fallback to the bare library name. I checked it fails without the change: the shadowing case returns the directory.

One thing worth flagging: this does not close #1945. That one is about the vendored go-nvlib copy, which still needs a bump once it is settled how (a release or a pseudo-version pin). I deliberately left `go.mod` and `vendor/` alone.


tariq1890:
Thanks for your PR, please ensure the following preconditions are met:

- DCO checks pass. The commit needs to be signed off
- The commit needs to be signed with a GPG key or similar. See [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Commits are squashed

cmd/nvidia-device-plugin/root_test.go
```diff
@@ -0,0 +1,96 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
```
tariq1890, line 2:
```suggestion
# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
```

rootkiller6788:
Thanks for the review.

The copyright header on the new test file is fixed as suggested — it now reads `# Copyright (c) NVIDIA CORPORATION.  All rights reserved.`. The test still passes: `go test -run TestTryResolveLibrary ./cmd/nvidia-device-plugin/` is 5/5.

On the three preconditions: the newest commit carries a `Signed-off-by`, but the two earlier commits on this branch still don't, and the branch is still three commits rather than one. Squashing them into a single signed-off and GPG-signed commit rewrites the branch, so I'll force-push that once the code side is settled rather than leave it half done.

Could you please take another look?

tariq1890:
DCO check is still failing. Please squash the commit history as well

tariq1890:
@rootkiller6788 Your commit signature is unverified.

PR #2036, opened 2026-09-15
Group MIG instances by parent GPU when applying allocation policies

jonathan-meiri:
### What

Allocation candidates are bucketed by the device UUID (`AnnotatedID.GetID()`). A MIG instance has its own UUID, so every MIG instance looks like a distinct physical GPU. As a result a multi-slot request under `distributed`/`packed` (and `spread`, in #1936) can land entirely on MIG instances of a **single** physical card, even when the intent is to span distinct GPUs.

This is pre-existing behavior surfaced during review of #1936 (thanks @tariq1890).

### Fix

Bucket candidates by the **physical GPU** instead of the advertised device:

- MIG instances group by their parent GPU, taken from `Device.Index` (`"<gpu>:<mig>"`).
- All other devices continue to key on their UUID — no change for the non-MIG case.

`Devices.PhysicalGPUKey` centralizes this, and `greedyAlloc`/`prepareCandidates` use it. All three policies benefit; `distributed`/`packed` are unchanged for non-MIG setups.

### Test

`TestDistributedAlloc_MIG_SpansDistinctPhysicalGPUs` sets up two idle MIG instances on GPU 0 and one busier time-sliced instance on GPU 1; a 2-slot request now spans both physical GPUs. It fails on the pre-fix bucketing and passes with it.

Follow-up to #1936.


myeolenv:

PR #2039, opened 2026-09-17
fix(rm): fall back to an unaligned allocation when device links are unavailable

### LINKED ISSUE #2038: [Bug]: A GPU that has fallen off the bus (Xid 79) makes GetPreferredAllocation fail for every pod on the node
* OS/Version(e.g. RHEL8.6, Ubuntu22.04): Ubuntu 22.04.5
* Kernel Version: 5.15.0-179-generic
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker): containerd 2.1.5
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): RKE2 v1.34.3
* Device plugin: v0.18.1 (deployed by GPU Operator), default config — full GPUs, no MIG, no time-slicing
* Driver: 580.159.03 (open kernel module), 8x GeForce RTX 5090, no NVLink

**Expected behavior.** When one GPU becomes unusable, pods that fit in the remaining healthy GPUs should still be admitted. The unhealthy device should not be able to break allocation for the rest of the node.

**Current behavior.** When a GPU falls off the PCIe bus (Xid 79), the health checker correctly marks that device `Unhealthy` and the node's `nvidia.com/gpu` allocatable drops from 8 to 7 — but from that moment **every** GPU pod scheduled to the node is rejected at kubelet admission, including pods that request a single GPU:

```
Warning  UnexpectedAdmissionError  pod/...  Allocate failed due to device plugin GetPreferredAllocation rpc failed
with err: rpc error: code = Unknown desc = error getting list of preferred allocation devices:
unable to get device link information: ...
```

The scheduler only looks at `allocatable = 7`, so it keeps binding pods to the node while kubelet keeps rejecting them. Any controller-owned pod (a Deployment in our case) therefore enters a tight create → `Failed` loop: we measured **~1,000 Failed pods per 5 minutes** per Deployment, sustained for 3 days until the workloads were stopped by hand, with the pod objects churning against `--terminated-pod-gc-threshold`. The device plugin itself kept running the whole time (it had initialised before the fault), so nothing restarted it, and `gpu-feature-discovery` on the same node crash-looped with `error getting device handle for index '4': Unknown Error`.

**Where it fails** (v0.18.1; the same code is on `main` today). On a node that exposes whole GPUs — i.e. not MIG devices and not replicated/shared ones, which is what `Devices.AlignedAllocationSupported()` checks — `getPreferredAllocation` always takes the aligned path, and that path builds the topology graph for *all* devices on the node **before** the `available` list is applied — `internal/rm/nvml_manager.go`:

```go
func (r *nvmlResourceManager) alignedAlloc(available, required []string, size int) ([]string, error) {
	linkedDevices, err := gpuallocator.NewDevices(
		gpuallocator.WithNvmlLib(r.nvml),
	)
	if err != nil {
		return nil, fmt.Errorf("unable to get device link information: %w", err)
	}
	availableDevices, err := linkedDevices.Filter(available)   // filtering happens only after the full walk
```

In `go-gpuallocator`, `NewDevices` → `build()` enumerates every device by index (`devicelib.GetDevices()` → `DeviceGetHandleByIndex(i)`) and then queries P2P/NVLink topology for **every pair** of devices. A lost GPU fails either at the handle lookup or at one of the pairwise topology calls, so the whole call returns an error. The plugin passes that error to kubelet, and kubelet's `callGetPreferredAllocationIfAvailable` has no fallback — the error becomes an admission failure.

Two consequences worth calling out:

* The plugin already knows which device is unhealthy (it reported it through `ListAndWatch`, and kubelet excludes it from `AvailableDeviceIDs`), but the allocation path never uses that information.
* There is no way to opt out: `GetPreferredAllocationAvailable` is unconditionally `true` and no flag disables it.

**Steps to reproduce**

1. A node with >= 2 full GPUs, device plugin v0.18.1, default configuration.
2. Make one GPU unreachable over PCIe. We hit a genuine Xid 79; the host then shows `nvidia-smi -L` → `Unable to determine the device handle for gpu <bdf>: Unknown Error`, and `lspci -s <bdf> -xxx` reads all `0xff` (the device no longer answers config reads).
3. Confirm `allocatable` = N-1 and the plugin log line `XidCriticalError: Xid=79 ... marking device as unhealthy`.
4. Create a pod requesting `nvidia.com/gpu: 1` with a nodeSelector for that node → `UnexpectedAdmissionError` with the message above.
5. Wrap the same pod in a Deployment → unbounded Failed-pod churn.

**Proposed fix** (happy to send the PRs)

* **A. Degrade instead of reject** (plugin only, small): in `alignedAlloc`, if `gpuallocator.NewDevices` fails, log a warning and fall back to a topology-unaware allocation (`required` first, then `available` in order, up to `size`) instead of returning an error. This removes both the admission failure and the pod churn; the only thing lost is topology optimisation while the node is degraded. It does not change health reporting, so no device is marked unhealthy that isn't known to be.
* **B. Build the graph only from the candidate devices** (`go-gpuallocator` + plugin): add an option such as `WithDeviceUUIDs(uuids []string)` so that `NewDevices` enumerates by UUID and computes links only within that subset, and have the plugin pass `available ∪ required`. A lost GPU is then never touched. This is the more complete fix but spans two repositories and a release ordering.

I would suggest landing A first (a few lines plus unit tests that inject a lost device and assert the healthy ones are still allocated), then B — but the direction is of course yours to decide, and I am happy to go straight to B, to a variant of either, or to drop A entirely if you would rather fix this only in `go-gpuallocator`.

Worth noting that the two do not overlap: once B is in place this class of failure never reaches the fallback, because the lost device is no longer enumerated. A would remain as a safety net for any other reason the link discovery can fail (NVML initialisation problems, transient driver errors), where returning a topology-unaware allocation is still preferable to failing admission for the whole node.

One design question I expect: should the fallback trigger on any `NewDevices` failure, or only on a lost device? I went with any failure, because the plugin cannot reliably classify the underlying NVML error, and the fallback result is still a correct allocation — it is only not topology-optimal, which is the same trade-off the plugin already accepts on nodes where links are reported as `P2PLinkUnknown`. Narrowing it to specific NVML return codes is easy to add if you prefer.

I already have the patch and unit tests for A ready locally (one file changed, no new dependencies, `make test`/`make lint`/`make check-modules` clean) and will open a PR shortly. Happy to be assigned this if that helps.

**Relation to existing issues.** #231 (2021) reports exactly this failure — "a single GPU crashing causes the other GPUs to become unallocatable", with the same error string and the same Failed-pod accumulation — and was closed by the stale bot in 2025 with a request to reopen in 2026. #1640 describes the same cascading pattern from a different trigger (NVML init failure) and was closed asking for a retest. #1641 proposed marking all devices unhealthy when the health check fails to start and was closed with the guidance that an NVML *init* failure should make the plugin exit instead. This report is deliberately narrower and different from that case: NVML is initialised and healthy for the other seven devices, and only per-device queries for the lost GPU fail, so neither "exit the plugin" nor "mark everything unhealthy" addresses it — the allocation path simply needs to tolerate (or avoid) the lost device.

### 3. Information to attach (optional if deemed irrelevant)

Common error checking:
 - [x] The output of `nvidia-smi -a` on your host — the lost GPU is not enumerated at all:
   ```
   Unable to determine the device handle for GPU 0000:99:00.0: Unknown Error
   # the other 7 GPUs report normally (idle, ~30 C, ~13 W)
   ```
 - [ ] Your docker configuration file — N/A (containerd)
 - [x] The k8s-device-plugin container logs — health check reacted correctly:
   ```
   health.go:143] Processing event {Device:{...} EventType:8 EventData:79 ...}
   health.go:169] XidCriticalError: Xid=79 on Device=GPU-3d04...; marking device as unhealthy.
   server.go:278] 'nvidia.com/gpu' device marked unhealthy: GPU-3d04...
   ```
   No further plugin errors and no restart; the plugin kept serving with 7 healthy devices.
 - [x] The kubelet logs / pod events on the node — one line per rejected pod:
   ```
   Allocate failed due to device plugin GetPreferredAllocation rpc failed with err: rpc error: code = Unknown
   desc = error getting list of preferred allocation devices: unable to get device link information: ...
   ```

Additional information that might help better understand your environment and reproduce the bug:
 - [ ] Docker version — N/A (containerd 2.1.5)
 - [x] Image and tag used: `nvcr.io/nvidia/k8s-device-plugin:v0.18.1` (GPU Operator deployment)
 - [x] Kernel version: `Linux 5.15.0-179-generic #189-Ubuntu SMP x86_64`
 - [x] Relevant kernel output from `dmesg` (host name removed):
   ```
   NVRM: GPU at PCI:0000:99:00: GPU-3d04...
   NVRM: Xid (PCI:0000:99:00): 79, GPU has fallen off the bus.
   NVRM: GPU 0000:99:00.0: GPU has fallen off the bus.
   NVRM: kgspRcAndNotifyAllChannels_IMPL: RC all channels for critical error 79.
   NVRM: Xid (PCI:0000:17:00): 154, GPU recovery action changed from 0x0 (None) to 0x2 (Node Reboot Required)
   ... (Xid 154 repeated for all 8 GPUs)
   ```
   For completeness on how long the bad window can last in practice: a soft `reboot` hung in the shutdown path waiting for the lost device to quiesce, so the node needed a full power cycle. After the power cycle all 8 GPUs enumerated again and passed a load test, but the plugin had been rejecting every GPU pod on that node for three days.
 - [x] NVIDIA driver version: 580.159.03 (open kernel module, `nvidia-driver-580-server-open`)
 - [x] NVIDIA Container Toolkit: 1.18.x (GPU Operator managed)

sirzzang:
## Description

Closes #2038.

On a node that exposes whole GPUs, `GetPreferredAllocation` takes the aligned path, and that path builds the link information for *all* physical devices before the `available` list is applied. A device that has fallen off the bus (Xid 79) therefore makes the call fail even though the health check has already marked it unhealthy and kubelet no longer offers it in `available`. The error becomes an admission failure for every GPU pod on the node, including single-device allocations that the remaining healthy devices could satisfy, and pods owned by a controller are recreated indefinitely: on the cluster where we hit this, one Deployment produced roughly a thousand `UnexpectedAdmissionError` pods every five minutes for three days.

This turns that failure into a degraded allocation instead. When link discovery fails, log a warning and select devices without considering the topology. `unalignedAlloc` selects only from `available`, always includes `required`, de-duplicates both, and still returns an error when the request cannot be satisfied — a negative size, more distinct required devices than the allocation size, a required device that is not available, or not enough distinct available devices.

The change is deliberately narrow and orthogonal to health reporting: no device is marked healthy or unhealthy, NVML initialisation policy is unchanged, and the aligned path is untouched whenever link discovery succeeds.

#2038 also proposes the more complete fix — enumerating only the candidate devices in `go-gpuallocator` so that a lost device is never touched. The two compose rather than compete: once that lands, this class of failure no longer reaches the fallback, which then only covers other reasons link discovery can fail. I am happy to follow whichever direction you prefer, including going straight to that one instead.

## Checklist

## Testing

Unit tests only — the hardware fault is not reproducible on demand.

`internal/rm/nvml_manager_fallback_test.go` injects a minimal `nvml.Interface` and `nvml.Device` test double, in the same style as `fakeNvmlLib` in `internal/rm/health_test.go` and without adding a dependency. It reports four devices but returns `ERROR_GPU_IS_LOST` for the handle of one of them, and only the three healthy IDs are passed as `available`, which is the state kubelet is in once the health check has removed the lost device.

This is also the first test to execute the aligned path: `TestFullGPUNodeIgnoresAllocationPolicy` currently notes that it can only check the branch condition "since alignedAlloc requires NVML".

I checked that the tests actually discriminate the change:

1. Tests added, fallback not yet implemented — the behavioural cases fail with the production error:

```
--- FAIL: TestAlignedAllocFallsBackWhenDeviceLinkInfoUnavailable/a_single_device_is_allocated_from_the_healthy_devices
    unable to get device link information: failed to get devices: error getting device handle for index '2': ERROR_GPU_IS_LOST
```

2. With the fallback — all cases pass: five subtests for the allocation path (single device, required devices included, the lost device is never allocated, the first device having fallen off the bus, allocating more devices than are available still fails) and ten table cases for `unalignedAlloc`.

3. Reverting only the call site while keeping the helper brings the failures back, and the boundary case ("allocating more devices than are available still fails") passes in both directions.

Local results on this branch:

```
make test                        # all packages pass, no failures
make lint                        # 0 issues
make check-modules               # clean
make check-third-party-notices   # clean
gofmt -l internal/rm/            # no output
```


myeolenv:

PR #2040, opened 2026-09-17
fix: omit Pod ownerRef on GFD NodeFeature objects

dundysm:
## Description

GFD was setting both the DaemonSet (controller) and the Pod as `ownerReferences` on `nvidia-features-for-<node>` NodeFeature objects (added in #1597).

After a GFD pod delete/recreate or node reboot, Kubernetes GC removes the deleted Pod ownerRef, leaving only the DaemonSet. The new pod then tries to re-add itself as an ownerRef. `OwnerReferencesPermissionEnforcement` rejects that update unless the caller can delete the NodeFeature:

```
cannot set an ownerRef on a resource you can't delete
```

GFD RBAC (helm role and gpu-operator `assets/gpu-feature-discovery`) grants get/list/watch/create/update on `nodefeatures`, but not delete. Create still works because that admission check applies to updates.

This change keeps only the DaemonSet controller ownerRef. That is enough for garbage collection when the GFD DaemonSet is removed, and ownerRefs stay stable across pod restarts so updates no longer attempt to add a new Pod ownerRef.

Related: NVIDIA/gpu-operator#2914

## Checklist

- [ ] Lint checks passing (`make lint`)
- [x] Unit tests passing (`make test` scoped: `go test ./internal/lm/`)
- [ ] Go mod artifacts in-sync (`make check-modules`)
- [ ] Third-party notices in-sync (`make check-third-party-notices`)

## Testing

- `go test ./internal/lm/ -count=1` (includes updated `TestGetOwnerRefs` and new `TestGetOwnerRefsStableAcrossPodRestart`)
- No GPU cluster or kind e2e run for this change

cc @tariq1890 @cdesiniotis @rahulait @elezar


dundysm:
@tariq1890 @cdesiniotis @rahulait @elezar gentle ping for review: gfd nodefeature ownerref fix for the pod-restart crashloop (gpu-operator#2914). ds-only ownerref; unit tests updated.

myeolenv:

tariq1890:
The addition of Pod UIDs as the owner ref is intentional, we must be careful before just deleting them.

Closing this PR for now. Please wait till the issue is fully understood and there is alignment with the maintainers on the approach.

PR #2044, opened 2026-09-18
Fix Helm effective namespace validation

ajavanma:
## Description

Validate the Helm chart's effective resource namespace so `--namespace default --set namespaceOverride=gpu-system` succeeds and the reverse override requires `allowDefaultNamespace=true`.

Use the existing namespace helper in one validation guard and remove the duplicate release-namespace guard. Preserve the error text and explicit opt-in. Add Go/Helm regression coverage for release namespaces with and without overrides, both with opt-in disabled and enabled, asserting errors and rendered DaemonSet namespaces.

## Checklist

- [ ] Go mod artifacts in-sync (`make check-modules`)
- [ ] Third-party notices in-sync (`make check-third-party-notices`)

## Testing

- Added regression test on the unchanged template: fails in the two override-direction cases, with 10 control cases passing.
- `make test-helm`: passes with the fix, including all 12 new regression cases.
- `make fmt` plus `gofmt -w tests/helm/helm_template_test.go`: passes, no unrelated changes.
- `make check`: passes, 0 lint issues.
- `make test`: passes, including package compilation and all four command builds. Vendored NVML code emits deprecation warnings.
- `git diff --check`: passes.
- Additional separate-module lint (`cd tests && golangci-lint run --allow-serial-runners --timeout=5m ./helm/...`) reports three `modernize/newexpr` suggestions in unchanged `ptr` code. Running the same lint against the base test file reproduces all three; no unrelated modernization is included.

Validated on Linux ARM64 with Go 1.27.1, Helm 4.3.0, and golangci-lint 2.13.2. GPU/cluster e2e tests were not run; this change is exercised through real Helm rendering without a cluster. Module and third-party-notice regeneration checks were not run because no dependencies or license inputs changed; their checklist boxes remain unchecked.


myeolenv:

PR #2045, opened 2026-09-18
Include missing default configuration name in error

ajavanma:
## Description

When config-manager selects an explicit default that does not exist, its error currently omits the profile name (`specified config  does not exist`). Report the requested default name and add regression coverage for missing/existing defaults and explicit node selections, preserving selection precedence and failure behavior.

## Checklist

## Testing

- New `TestUpdateConfigName` fails on unchanged upstream only for the missing-default diagnostic; its four control cases pass.
- `go test ./cmd/config-manager -count=1 -v` passes after the fix.
- `make fmt`, `make check` (0 lint issues), and `make test` pass with Go 1.27.1 on Linux ARM64. `make test` also builds all packages and four commands.
- `make check-modules` and `make check-third-party-notices` pass; notices were regenerated for Linux AMD64 and ARM64.

No Kubernetes/GPU end-to-end tests were run. Module and third-party notice regeneration checks also pass on this commit, with no file differences. The module check passed after retrying transient dependency-download EOF failures. The build emits deprecation warnings from vendored NVML C bindings.


myeolenv:

abrarshivani (approved):

PR #2054, opened 2026-09-22
Prevent duplicate GPU registration during health checks

### LINKED ISSUE #2052: [Bug]: Duplicate GPU health checking registration
Currently, `internal/rm/health.go`'s `checkHealth` function loops over all discovered `Device`s. On every iteration, it finds the current `Device`'s parent UUID and calls `gpu.RegisterEvents` on that parent. The problem is multiple `Device` objects can have the same parent: an example is with two MIG devices sharing the same parent GPU; or two time-sliced instances of the same GPU. This can result in multiple duplicate `gpu.RegisterEvents` calls being made to the same parent GPU. 

This is related to [#1929](https://github.com/NVIDIA/k8s-device-plugin/issues/1929). Implementation should wait until [#1930](https://github.com/NVIDIA/k8s-device-plugin/pull/1930) is merged because it modifies the same loop and parent-to-device mapping. The follow-up can then register events once per unique parent UUID.

aryangorwade:
## Description

This PR removes redundant GPU health registration calls that can happen when multiple `Device` objects share the same parent GPU. It builds on changes made in #1930 and should be merged in afterwards.

Closes #2052 . 

## Checklist

- [] Test cases are added for new code paths (test cases not possible without nvml mocks) 

## Testing

Verified via manual testing on a GPU node. Scenarios tested include unshared GPUs, time-sliced & MPS shared GPUs, multiple MIG instances and time-sliced MIG & mixed MIG. Plugin restart/config reload behavior was also tested.


kvalliyurnatt (commented):
LGTM. Is it worth adding a unit test for this ?

aryangorwade:
The only practical way I think we can test this is by mocking checkHealth's nvml event loop. This repo generally avoids doing that (the test file for this mocks nvml but it was recently written by an external contributor). It has its disadvantages like causing bugs like this to be present. If we do choose to write a test for `checkHealth` it should probably be an end-to-end test that checks every part of the loop. I think for now there's no point in adding a unit test only for this.

